PR #4, opened 2019-03-13
switch #file/#function to String

weissi:
Motivation:

With the new, awesome Swift 5 `String`s, there doesn't seem to be any
point in using `StaticString`s anymore. `String` seems to use immortal
strings for literals now and is in fact 1 byte less wide than
StaticString.

Modification:

Switch all StaticString for String

Result:

better asm (slightly shorter), fewer registers burnt, more features


weissi:
CC @ilseman

PR #8, opened 2019-03-14
first cut at API docs

weissi:
same thing here: it's far from perfect but it's a start for API docs.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -17,27 +17,58 @@
     import Glibc
 #endif
 
-/// This is the protocol a custom logger implements.
+/// A `LogHandler` is an implementation of a logging backend.
+///
+/// This type is an implementation detail and should not normally be used, unless implementing your own logging backend.
+/// For user API of please refer to the documentation of `Logger`.
 public protocol LogHandler {
-    // This is the custom logger implementation's log function. A user would not invoke this but rather go through
-    // `Logger`'s `info`, `error`, or `warning` functions.
-    //
-    // An implementation does not need to check the log level because that has been done before by `Logger` itself.
-    func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, file: String, function: String, line: UInt)
+    /// This functions is called when a `LogHandler` must emit a log message. There is no need for the `LogHandler` to
+    /// check if the `level` is above or below the configured `logLevel` as `Logger` already performed this check and
+    /// determined that a message should be logged.
+    ///
+    /// - parameters:
+    ///     - level: The log level the message was logged at.
+    ///     - message: The message to log. To objtain a `String` represenation call `message.description`.
+    ///     - metadata: The metadata associated to this log message.
+    ///     - file: The file the log message was emitted from.
+    ///     - function: The function the log line was emitted from.
+    ///      -line: The line the log message was emitted from.
+    func log(level: Logger.Level,
+             message: Logger.Message,
+             metadata: Logger.Metadata?,
+             file: String, function: String, line: UInt)
 
-    // This adds metadata to a place the concrete logger considers appropriate. Some loggers
-    // might not support this feature at all.
+    /// Add, remove, or change the logging metadata.
+    ///
+    /// - note: `LogHandler`s must treat logging metadata as a value type. This means that the change in metadata must
+    ///         only affect this very `LogHandler`.
+    ///
+    /// - parameters:
+    ///    - metadataKey: The key for the metadata item
     subscript(metadataKey _: String) -> Logger.Metadata.Value? { get set }
 
-    // All available metatdata
+    /// Get or set the entire metadata storage as a dictionary.
+    ///
+    /// - note: `LogHandler`s must treat logging metadata as a value type. This means that the change in metadata must
+    ///         only affect this very `LogHandler`.
     var metadata: Logger.Metadata { get set }
 
-    // The log level
+    /// Get or set the configured log level.
+    ///
+    /// - note: `LogHandler`s must treat the log level as a value type. This means that the change in metadata must
+    ///         only affect this very `LogHandler`. It is acceptable to provide some form of global log level override
+    ///         that means a change in log level on a particular `LogHandler` might not be reflected in any
+    ///        `LogHandler`.
     var logLevel: Logger.Level { get set }
 }
 
-// This is the logger itself. It can either have value or reference semantics, depending on the `LogHandler`
-// implementation.
+/// A `Logger` is the central type in `swift-log`. Its central function is to emit log messages using one of the methods
+/// corresponding to a log level.
+///
+/// The most basic usage of a `Logger` is
+///
+///     logger.info("Hello World!")
+///
 public struct Logger {
     @usableFromInline
     var handler: LogHandler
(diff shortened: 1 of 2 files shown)


Sources/Logging/Logging.swift
```diff
@@ -156,32 +335,78 @@ public enum LoggingSystem {
 }
 
 extension Logger {
+    /// `Metadata` is a typealias for `[String: Logger.Metadata.Value]` the type of the metadata storage.
```
tomerd: Logger.Metadata.Value -> MetadataValue?

Sources/Logging/Logging.swift
```diff
@@ -156,32 +335,78 @@ public enum LoggingSystem {
...
 
+    /// A logging metadata value. `Logger.Metadata.Value` is string, array, and dictionary literal convertible.
     public enum MetadataValue {
+        /// A metadata value which is a `String`.
         case string(String)
+
+        /// A metadata value which is some `CustomStringConvertible`.
         case stringConvertible(CustomStringConvertible)
+
+        /// A metadata value which is a dictionary from `String` to `Logger.Metadata.Value`.
```
tomerd: Logger.Metadata.Value -> MetadataValue ?

Sources/Logging/Logging.swift
```diff
@@ -156,32 +335,78 @@ public enum LoggingSystem {
...
         case dictionary(Metadata)
+
+        /// A metadata value which is an array of `Logger.Metadata.Value`s.
```
tomerd: Logger.Metadata.Value -> MetadataValue?

Sources/Logging/Logging.swift
```diff
@@ -156,32 +335,78 @@ public enum LoggingSystem {
...
+    ///
+    /// The `label` should identify the creator of the `Logger`. This can be an application, a sub-system, or even
+    /// a datatype.
+    ///
+    /// This initialiser provides an escape hatch in case the global default logging backend implementation (set up
+    /// using `LoggingSystem.bootstrap` is not appropriate for this particular logger.
+    ///
+    /// - parameters:
+    ///     - label: An identifier for the creator of a `Logger`.
+    ///     - factory: A closure creating non-standard `LogHandler`s.
```
tomerd: i kinda like the original tone of "we do not expect this API to be used in normal circumstances, so if you find yourself using it make sure its for a good reason" and suggest we make a note of that nature

Sources/Logging/Logging.swift
```diff
@@ -17,27 +17,57 @@
...
+    ///        `LogHandler`.
     var logLevel: Logger.Level { get set }
 }
 
-// This is the logger itself. It can either have value or reference semantics, depending on the `LogHandler`
-// implementation.
+/// A `Logger` is the central type in `swift-log`. Its central function is to emit log messages using one of the methods
+/// corresponding to a log level.
+///
+/// The most basic usage of a `Logger` is
```
ktoso: `is: `

Sources/Logging/Logging.swift
```diff
@@ -156,32 +335,78 @@ public enum LoggingSystem {
...
     }
     
-    // this is to provide an escape hatch for situations one must use a custom factory instead of the gloabl one
-    // we do not expect this API to be used in normal circumstances, so if you find yourself using it make sure its for a good reason
+    /// Construct a `Logger` given a `label` identifying the creator of the `Logger` or a non-standard `LogHandler`.
+    ///
+    /// The `label` should identify the creator of the `Logger`. This can be an application, a sub-system, or even
+    /// a datatype.
+    ///
+    /// This initialiser provides an escape hatch in case the global default logging backend implementation (set up
```
ktoso: Let's stick to american spelling perhaps? (yeah I dislike it as well, though it is used through all [official docs](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html) and we say `initialized` in the sources as well; _consistency is king_).

initialiser -> `initializer`

Sources/Logging/Logging.swift
```diff
@@ -156,32 +335,78 @@ public enum LoggingSystem {
...
+        /// A metadata value which is a dictionary from `String` to `Logger.Metadata.Value`.
         case dictionary(Metadata)
+
+        /// A metadata value which is an array of `Logger.Metadata.Value`s.
         case array([Metadata.Value])
     }
 
+    /// The log level.
     public enum Level: Int {
+        /// The `.debug` log level, appropriate for messages that contain information normally of use only when
```
ktoso: Not sure the `The `.debug` log level,` are helpful.

Maybe just the "Appropriate for..."

?

Sources/Logging/Logging.swift
```diff
@@ -86,59 +148,176 @@ extension Logger {
             self.handler.logLevel = newValue
         }
     }
-
 }
 
 extension Logger {
+    /// Log a message passing with the `Logger.info` log level.
+    ///
+    /// If `.info` is more severe than the `Logger`'s `logLevel`, it will be logged, otherwise nothing will happen.
```
ktoso:
```
 If `.info` is more seve
```

->

```
 If `.debug` is more seve
```

Sources/Logging/Logging.swift
```diff
@@ -156,32 +335,78 @@ public enum LoggingSystem {
...
         case stringConvertible(CustomStringConvertible)
+
+        /// A metadata value which is a dictionary from `String` to `Logger.Metadata.Value`.
         case dictionary(Metadata)
+
+        /// A metadata value which is an array of `Logger.Metadata.Value`s.
         case array([Metadata.Value])
     }
 
+    /// The log level.
```
ktoso: Let's also clarify which order severity goes in the numeric values?

```
/// Raw values of log levels correspond to their severity, and are ordered by lowest numeric value (0) being the most severe.
```

or something like that?

Sources/Logging/Logging.swift
```diff
@@ -214,6 +439,17 @@ extension Logger.MetadataValue: Equatable {
...
 
 extension Logger {
+    /// `Logger.Message` represents a log message's text. It is usually created using string literals.
+    ///
+    /// Example creating a `Logger.Message`:
+    ///
+    ///     let world: String = "world"
+    ///     let myLogMessage: Logger.Message = "Hello \(world)"
+    ///
+    /// Most commonly, `Logger.Message`s appear simply as the parameter to a logging method such as
```
ktoso: `as:`

Sources/Logging/Logging.swift
```diff
@@ -229,7 +465,11 @@ extension Logger {
     }
 }
 
-/// Ships with the logging module, used to multiplex to multiple logging handlers
+/// A pseudo-`LogHandler` that can be used to send messages to multiple other `LogHandler`s.
```
ktoso: Maybe avoid calling it "pseudo"?  

Maybe just proxying or "A `LogHandler` forwarding all messages to multiple wrapped handlers." (also avoiding the repetition of LogHandler word)

Sources/Logging/Logging.swift
```diff
@@ -86,59 +148,176 @@ extension Logger {
             self.handler.logLevel = newValue
         }
     }
-
 }
 
 extension Logger {
+    /// Log a message passing with the `Logger.info` log level.
```
ktoso: This reads a bit weird... "passing with the ... log level"?

Maybe like this instead:

```
/// Log a message on the `Logger.Level.info` level
```

?

Sources/Logging/Logging.swift
```diff
@@ -50,13 +80,35 @@ public struct Logger {
 }
 
 extension Logger {
+    /// Log a message passing the log level as a parameter.
+    ///
+    /// If the `logLevel` passed to this method is more severe than the `Logger`'s `logLevel`, it will be logged,
```
ktoso: Wording of "if ... is more severe ..." has to be 

> is more severe _or equal_

Since: `   if self.logLevel >= level {`

Sources/Logging/Logging.swift
ktoso: The missing `Level` type is only here, not on the other levels.

Though the wording, whichever we use should be consistent then.

Sources/Logging/Logging.swift
```diff
@@ -17,27 +17,57 @@
...
 public protocol LogHandler {
-    // This is the custom logger implementation's log function. A user would not invoke this but rather go through
-    // `Logger`'s `info`, `error`, or `warning` functions.
-    //
-    // An implementation does not need to check the log level because that has been done before by `Logger` itself.
-    func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, file: String, function: String, line: UInt)
-
-    // This adds metadata to a place the concrete logger considers appropriate. Some loggers
-    // might not support this feature at all.
+    /// `log` is called when a `LogHandler` must emit a log message. There is no need for the `LogHandler` to check
```
ktoso: `This function` rather than start sentence of `log`?

Sources/Logging/Logging.swift
```diff
@@ -17,27 +17,57 @@
...
+    /// if the `level` is above or below the configured `logLevel`, `Logger` already performed this check and
```
ktoso: add "as" before `Logger` to read as:

> there is no need to check ..., as Logger already performed this check

so the "as" leads into the reason for why

Sources/Logging/Logging.swift
```diff
@@ -17,27 +17,57 @@
     import Glibc
 #endif
 
-/// This is the protocol a custom logger implements.
+/// A `LogHandler` is an implementation of a logging backend. You only need to worry about `LogHandler`s if you wish
```
ktoso: Cut line after the first sentence?

Perhaps avoid "worry" phrasing? 

e.g.

```
/// A `LogHandler` is an implementation of a logging backend. 
/// 
/// This type is an implementation detail and should not normally be accessed, unless implementing your own logging backend.
/// For user API of please refer to the documentation of `Logger`.
```

or something like that?

Sources/Logging/Logging.swift
```diff
@@ -86,59 +148,176 @@ extension Logger {
...
     public func emergency(_ message: @autoclosure () -> Logger.Message, metadata: @autoclosure () -> Logger.Metadata? = nil, file: String = #file, function: String = #function, line: UInt = #line) {
         self.log(level: .emergency, message(), metadata: metadata(), file: file, function: function, line: line)
     }
 }
 
-// This is the logging system itself, it's mostly used to obtain loggers and to set the type of the `LogHandler`
-// implementation.
+/// The `LoggingSystem` is a global facility where the default logging backend implementation (`LogHandler`) can be
+/// configured. `LoggingSystem` is usually just once in a given program to set up the desired logging backend
+/// implementation.
```
ktoso: I think this is missing a verb:

> `LoggingSystem` is usually just once in a given program to set up the desired logging backend implementation.

"is just once" what? Maybe "Only one instance of LoggingSystem exists..." or "is set only once"?

Sources/Logging/Logging.swift
ktoso: +1 on keeping around some discouraging wording about this initializer.

PR #15, opened 2019-03-19
Formatting, readability

kirilltitov:
Mostly it's about 120 characters per line.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -102,12 +102,10 @@ extension Logger {
                     metadata: @autoclosure () -> Logger.Metadata? = nil,
                     file: String = #file, function: String = #function, line: UInt = #line) {
         if self.logLevel >= level {
-            self.handler.log(
-                level: level,
-                message: message(),
-                metadata: metadata(),
-                file: file, function: function, line: line
-            )
+            self.handler.log(level: level,
+                             message: message(),
+                             metadata: metadata(),
+                             file: file, function: function, line: line)
         }
     }
 
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -97,12 +97,10 @@ extension Logger {
     ///    - line: The line this log message originates from (there's usually no need to pass it explicitly as it
     ///            defaults to `#line`.
     @inlinable
-    public func log(
-        level: Logger.Level,
-        _ message: @autoclosure () -> Logger.Message,
-        metadata: @autoclosure () -> Logger.Metadata? = nil,
-        file: String = #file, function: String = #function, line: UInt = #line
-    ) {
+    public func log(level: Logger.Level,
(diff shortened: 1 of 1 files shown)


Sources/Logging/Logging.swift
```diff
@@ -454,7 +494,10 @@ extension Logger {
     ///
     ///     logger.info("Hello \(world)")
     ///
-    public struct Message: ExpressibleByStringLiteral, Equatable, CustomStringConvertible, ExpressibleByStringInterpolation {
+    public struct Message: ExpressibleByStringLiteral,
+                           Equatable,
+                           CustomStringConvertible,
+                           ExpressibleByStringInterpolation {
```
kirilltitov: These lines are questionable. I'm not sure about this formatting convention.

Sources/Logging/Logging.swift
```diff
@@ -97,12 +97,19 @@ extension Logger {
...
     @inlinable
-    public func log(level: Logger.Level,
-                    _ message: @autoclosure () -> Logger.Message,
-                    metadata: @autoclosure () -> Logger.Metadata? = nil,
-                    file: String = #file, function: String = #function, line: UInt = #line) {
+    public func log(
+        level: Logger.Level,
+        _ message: @autoclosure () -> Logger.Message,
+        metadata: @autoclosure () -> Logger.Metadata? = nil,
+        file: String = #file, function: String = #function, line: UInt = #line
```
kirilltitov: Maybe `file` and `function` should be `StaticString`?

weissi: @kirilltitov 

- see #4 , `String` in Swift 5 leads to slightly better code and one byte less storage that needs to be copied per call.
- the indentation doesn't match Xcode's default, I think we should use the existing indentation

```swift
func foo(bar: Bar,
         buz: Buz)
```

Sources/Logging/Logging.swift
weissi: I'd be fine with this

Sources/Logging/Logging.swift
kirilltitov: Oh well, I don't like that C-like arguments formatting at all, but whatever :)

kirilltitov:
@weissi addressed your comments 👍

Sources/Logging/Logging.swift
```diff
@@ -102,7 +102,12 @@ extension Logger {
                     metadata: @autoclosure () -> Logger.Metadata? = nil,
                     file: String = #file, function: String = #function, line: UInt = #line) {
         if self.logLevel >= level {
-            self.handler.log(level: level, message: message(), metadata: metadata(), file: file, function: function, line: line)
+            self.handler.log(
+                level: level,
```
weissi: @kirilltitov mind fixing the indentation here too?

kirilltitov: Oof, right. Done.

PR #22, opened 2019-03-21
[DO NOT MERGE] Swift 4.2 support for a 0.x release

weissi:
---

**NOTES**: This PR should never land on the `master` branch as it _removes_ functionality. If we think it's useful we could however put this on a `swift-log-0` branch and tag say version `0.1.0` with it. The real version will remain Swift 5-only and we could then just tag it `1.0.0` when Swift 5 becomes generally available.

---
Motivation:

For an easier migration path, it is probably useful to ship a `0.x`
version of `swift-log` that is compatible with Swift 4.2. It can't use
any of the features that Swift 5 brings and therefore the Swift 4.2
version will have a slightly reduced feature set:

- log messages will be `String`
- logging metadata values will be `String` only (no dictionaries,
  arrays, ...)

That means, if you use the Swift 4.2 version of swift-log carefully
(only use string literals), your migration to the Swift 5 version should
only be a matter of changing your `Package.swift`.

Modifications:

- remove all string interpolation and use `String` directly

Result:

- we can release a 0.x version that supports Swift 4.2

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:4.0
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift Logging API open source project
--- a/Sources/Logging/Locks.swift
+++ b/Sources/Logging/Locks.swift
@@ -49,7 +49,11 @@ internal final class Lock {
     deinit {
         let err = pthread_mutex_destroy(self.mutex)
         precondition(err == 0)
+        #if swift(>=4.0) && !swift(>=4.0.50)
+        self.mutex.deallocate(capacity: 1)
+        #else
         self.mutex.deallocate()
+        #endif
     }
 
     /// Acquire the lock.
@@ -80,7 +84,6 @@ extension Lock {
     ///
     /// - Parameter body: The block to execute while holding the lock.
     /// - Returns: The value returned by the block.
-    @inlinable
     internal func withLock<T>(_ body: () throws -> T) rethrows -> T {
         self.lock()
         defer {
@@ -90,7 +93,6 @@ extension Lock {
     }
 
     // specialise Void return (for performance)
-    @inlinable
     internal func withLockVoid(_ body: () throws -> Void) rethrows {
         try self.withLock(body)
     }
@@ -113,7 +115,11 @@ internal final class ReadWriteLock {
     deinit {
         let err = pthread_rwlock_destroy(self.rwlock)
         precondition(err == 0)
+        #if swift(>=4.0) && !swift(>=4.0.50)
+        self.rwlock.deallocate(capacity: 1)
+        #else
         self.rwlock.deallocate()
+        #endif
     }
 
     /// Acquire a reader lock.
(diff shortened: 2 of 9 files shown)


ianpartridge:
Alternatively, we could merge this to `master` and ship it as 1.0.0, then add the interpolation feature later once Swift 5 is widely required. It would also let us see how painful only having `MetadataValue` as a simple String is in practice.

weissi:
> Alternatively, we could merge this to `master` and ship it as 1.0.0, then add the interpolation feature later once Swift 5 is widely required. It would also let us see how painful only having `MetadataValue` as a simple String is in practice.

@ianpartridge I don't really want to do this because the `LogHandler`s need to properly support all `MetadataValue`s. But if they all start assuming it's just a `String`, then it'll probably be hard to make them all change their code. I want it to be very clear that the 4.x version is something pretty transitional.

weissi:
@ianpartridge actually, I have an idea! Let's keep the `Metadata` enum for Swift 4 and we can just convert (in `Logger`) the `String` to a `MetadataValue.string(theStringPassedIn)`. Then `LogHandlers` have to support all `MetadataValue`s but they just happen to only ever get `String` in Swift 4 :). That actually would also work for the `Logger.Message` thing.

Ok, this is much better, scrap this PR, I'll improve this later.

I still think this should become version 0.x for Swift 4 and 1.x for Swift 5.

ktoso:
> Let's keep the Metadata enum for Swift 4 and we can just convert (in Logger) the String to a MetadataValue.string(theStringPassedIn). Then LogHandlers have to support all MetadataValues but they just happen to only ever get String in Swift 4 :). That actually would also work for the Logger.Message thing.

Big +1, I was wondering if it was really necessary to drop the metadata value...

tomerd:
nice, even better! 

1. need to adjust linux tests
2. consider `metadata: @autoclosure () -> [String: String] = [:]` instead of `metadata: @autoclosure () -> [String: String]? = nil`, but not a strong opinion
3. branching wise, i suggest we follow what we do with swift-nio, i.e have master be the swift 5 version, and create a separate 4.x branch (swift-log-0.1 or whatnot) which we can tag

weissi:
> nice, even better!
> 
> 1. need to adjust linux tests

I did run `scripts/generate_linux_tests`, is that what you mean?



> 2\. consider `metadata: @autoclosure () -> [String: String] = [:]` instead of `metadata: @autoclosure () -> [String: String]? = nil`, but not a strong opinion

I think this is something that should be changed on `master`. `master` has `metadata: Logger.Metadata?`, therefore I went for `metadata: [String: String]?` here which is the equivalent.


> 3\. branching wise, i suggest we follow what we do with swift-nio, i.e have master be the swift 5 version, and create a separate 4.x branch (swift-log-0.1 or whatnot) which we can tag

agreed

tomerd:
@weissi CI is failing with `=> Checking linux tests... missing changes!` so i think something went a miss there

weissi:
> @weissi CI is failing with `=> Checking linux tests... missing changes!` so i think something went a miss there

@tomerd ah thanks! I think that was my `git` config that's trying not to record whitespace errors but that's not a good idea for generated files ;)

weissi:
updated

tomerd:
lgtm!

weissi:
@tomerd updated for 4.2, also `origin/jw-4.2-branch`

tomerd:
thanks @weissi

weissi:
now lives on the `swift-log-0.0` branch

PR #30, opened 2019-03-26
Add small tweaks to the README.

Yasumoto:
Small typos and added another example. This package looks so great, thank you all for the hard work!

Happy to make other tweaks or improvements y'all notice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ What `swift-log` provides today can be found in the [API docs][api-docs]. At thi
 
 ---
 
-**NOTE**: You will need Xcode 10.2 beta 4 or a recent [Swift 5.0 development snapshot](http://swift.org/download/#swift-50-development) to try out `swift-log`.
+**NOTE**: You will need Xcode 10.2 beta 4, [Xcode 10.2](https://itunes.apple.com/us/app/xcode/id497799835) or a recent [Swift 5.0 development snapshot](http://swift.org/download/#swift-50-development) to try out `swift-log`.
 
 ---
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,7 +25,11 @@ To depend on the logging API package, you need to declare your dependency in you
 .package(url: "https://github.com/apple/swift-log.git", .branch("master")),
 ```
 
-and to your application/library target, add `"Logging"` to your `dependencies`.
+and to your application/library target, add `"Logging"` to your `dependencies`:
+
+```swift
+.target(name: "BestExampleApp", dependencies: ["Logging']),
+```
```
ktoso: good idea to add this, perhaps the above should lead into this though:

> and to your application/library target, add `"Logging"` to your `dependencies`, e.g. like this:
> ...

README.md
```diff
@@ -8,7 +8,7 @@ What `swift-log` provides today can be found in the [API docs][api-docs]. At thi
 
 ---
 
-**NOTE**: You will need Xcode 10.2 beta 4 or a recent [Swift 5.0 development snapshot](http://swift.org/download/#swift-50-development) to try out `swift-log`.
+**NOTE**: You will need Xcode 10.2 beta 4, [Xcode 10.2](https://itunes.apple.com/us/app/xcode/id497799835) or a recent [Swift 5.0 development snapshot](http://swift.org/download/#swift-50-development) to try out `swift-log`.
```
ktoso: his this a good link, won't change?

weissi: oh thanks. If you don't mind you could actually just say that you need Xcode 10.2 or Swift 5.0 . Don't think the dev snapshots or betas are still relevant and we won't unit test them anyway so I think they can be removed

Yasumoto: Looks like it to me:

> These URLs are great for linking to your app or company page in the App Store or Mac App Store from your website or inside your app, or anywhere you don't need to type or speak the URL. As a rule, always use these standard links when driving users to your app from your digital marketing communications.


https://developer.apple.com/library/archive/qa/qa1633/_index.html

Yasumoto: Ah, yah it's also linked from https://swift.org/download/#releases for the Xcode link.

Yasumoto: Great call, fixed.

PR #33, opened 2019-03-28
[HOLD] Remove `Logger.Level` numeric values while still supporting `Comparable` conformance

Mordil:


Sources/Logging/Logging.swift
```diff
@@ -392,8 +400,21 @@ extension Logger {
 }
 
 extension Logger.Level: Comparable {
+    private var severityOrdering: Int {
+        switch self {
+        case .debug: return 0
```
weissi: So I was very worried about the code-size that this will generate but the Swift compiler is really smart and as long as the order of the cases is aligned with the numbers, this works really well as the compiler can see through this.

Sources/Logging/Logging.swift
```diff
@@ -392,8 +400,21 @@ extension Logger {
 }
 
 extension Logger.Level: Comparable {
+    private var severityOrdering: Int {
```
weissi: this will need to become `@inlinable`

weissi:
thanks! I'm happy with this but we should sit on this until the discussion on the forums concludes if this is a good idea or not.

Mordil:
Definitely agree! Just wanted to help contribute where I had free time and to have something concrete to discuss

Sources/Logging/Logging.swift
Mordil: https://twitter.com/migueldeicaza/status/1109173266483036160

Sources/Logging/Logging.swift
Mordil: Does it? Or just `@usableFromInline var ...`?

I am probably misunderstanding what the compiler can and can't do - but I assume that the protocol conformance methods are not inlinable, but code within it would be?

Mordil:
Closed in favor of #35

PR #35, opened 2019-04-05
change log levels to be more natural

weissi:
Motivation:

The syslog log level was unnatural in order (`.error` < `.info`) but also
didn't have `.trace` and instead the odd `.emergency` and `.alert`.

Modification:

As per https://forums.swift.org/t/logging-levels-for-swifts-server-side-logging-apis-and-new-os-log-apis/20365/56 ,
we decided to go for

    trace < debug < info < notice < warning < error < critical

and also remove the integrals completely. The levels remain comparable
though.

Result:

log levels more natural


ktoso:
Late to the party but LGTM :-) Looks and feels _really_  good now!

PR #36, opened 2019-04-08
remove Logger.metadata property

### LINKED ISSUE #31: take out `metadata` to start with and make opaque later
On the forums , a few have rightly pointed out that currently, there's two places to get/set the metadata:

    logger[metdataKey: "foo"] = "bar"

and

    logger.metadata["foo"] = "bar"

The intention is to always use the former (`logger[metdataKey: "foo"] = "bar"`) unless you want to store/retrieve the whole metadata storage, for example for MDC.

But as it was correctly pointed out, if that's the case we could/should make it opaque.

My proposal for right now is that the `LogHandler`s need to support both but the `Logger` should only expose the subscript. In a new minor version that we can release anytime, we could then add a back a well-designed story for MDC and friends.

weissi:
Motivation:

Logger had two ways to access the metadata:

    logger[metadataKey: "foo"] = "bar"

and

    logger.metadata["foo"] = "bar"

. The first one is the preferred API (as it doesn't force the
`LogHandler`s to store it in a dictionary) but the second one was still
available for cases where you need to import/export the whole metadata
storage.

Instead of exposing this as a dictionary, we should make the
'whole-metadata import/export' opaque. We're close to tagging 1.0.0
however let's design this API properly and add a sensible implementation
for 1.1.0 or so.

Moditications:

remove `Logger.metadata` property

Result:

- We can design the import/export APIs and add them when they're ready.
- the `LogHandler` API is unchanged
- fixes #31

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -79,20 +79,6 @@ extension Logger {
         }
     }
 
-    /// Get or set the entire metadata storage.
-    ///
-    /// - note: Logging metadata behaves as a value that means a change to the logging metadata will only affect the
-    ///         very `Logger` it was changed on.
-    @inlinable
-    public var metadata: Logger.Metadata {
-        get {
-            return self.handler.metadata
-        }
-        set {
-            self.handler.metadata = newValue
-        }
-    }
-
     /// Get or set the log level configured for this `Logger`.
     ///
     /// - note: `Logger`s treat `logLevel` as a value. This means that a change in `logLevel` will only affect this
--- a/Tests/LoggingTests/TestLogger.swift
+++ b/Tests/LoggingTests/TestLogger.swift
@@ -45,9 +45,7 @@ internal struct TestLogHandler: LogHandler {
 
     func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, file: String, function: String, line: UInt) {
         let metadata = (self._metadataSet ? self.metadata : MDC.global.metadata).merging(metadata ?? [:], uniquingKeysWith: { _, new in new })
-        var l = logger // local copy since we gonna override its metadata
-        l.metadata = metadata
-        l.log(level: level, message, metadata: metadata, file: file, function: function, line: line)
+        logger.log(level: level, message, metadata: metadata, file: file, function: function, line: line)
         self.recorder.record(level: level, metadata: metadata, message: message)
     }
 


ktoso:
> Instead of exposing this as a dictionary, we should make the 'whole-metadata import/export' opaque. 

To sanity check my understanding of the phrase: We'll want to design a type that is "the metadata container" that could be then implemented in various ways, but not forcing it to be a dict?

> We're close to tagging 1.0.0 however let's design this API properly and add a sensible implementation for 1.1.0 or so.

Agreed with cutting out and designing well later for 1.1.0, I think I'd definitely have use cases for it (can open the ticket with an example use case).

ktoso:
Btw, how are we treating APIs in sandbox mode with regards to source compat, e.g. if we suddenly find out that we have to change the way LogHandler has this defined? As with this PR it still has the `var metadata: Logger.Metadata { get set }`, would we end up calling a 2.0 if we had to remove it?

weissi:
> Btw, how are we treating APIs in sandbox mode with regards to source compat, e.g. if we suddenly find out that we have to change the way LogHandler has this defined? As with this PR it still has the `var metadata: Logger.Metadata { get set }`, would we end up calling a 2.0 if we had to remove it?

Yes, we follow SemVer, so changing an existing API would mean `2.0.0` but that's okay I think. We could go along way with deprecations and back-filling old APIs with default implementations I think.

weissi:
> > Instead of exposing this as a dictionary, we should make the 'whole-metadata import/export' opaque.
> 
> To sanity check my understanding of the phrase: We'll want to design a type that is "the metadata container" that could be then implemented in various ways, but not forcing it to be a dict?
> 
> > We're close to tagging 1.0.0 however let's design this API properly and add a sensible implementation for 1.1.0 or so.
> 
> Agreed with cutting out and designing well later for 1.1.0, I think I'd definitely have use cases for it (can open the ticket with an example use case).

Yes, totally agreed that there are use-cases, I hope to bring it back in 1.1.0 or so

ktoso:
Cool, thanks for confirming. Happy to bump major even during sandbox time, no reason not to -- otherwise people and projects end up with "2.0 fear" 😜 👍

weissi:
> Happy to bump major even during sandbox time, no reason not to -- otherwise people and projects end up with "2.0 fear" 😜 👍

Agreed. We talked about it in the SSWG meeting and the idea is that the maturity levels and the version numbers are totally orthogonal. The only meaning attached to 1.0.0 is "we follow SemVer"

ktoso:
> looks good, let’s add an issue to track enabling mdc use cases

Already here: https://github.com/apple/swift-log/issues/37 👍

PR #40, opened 2019-04-10
Support Swift 4.1

ianpartridge:
For some users, having Swift 4.1 support will be useful. Update the `jw-4.2-support` branch to support Swift 4.1.

Changes:

- Remove `@inlinable` and `@usableFromInline` annotations which were added in Swift 4.2. If the package is too slow, people should use Swift 5 and `from: "1.0.0"`.
- Remove `Int.random(in:)` which was added in Swift 4.2. Although this makes the tests slightly worse, it's NBD. (It was either that or start implementing `Int.random(in:)` via `arc4_random()` etc. and life is too short.)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.2
+// swift-tools-version:4.1
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift Logging API open source project
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.1
+// swift-tools-version:4.0
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift Logging API open source project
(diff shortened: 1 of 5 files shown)


ianpartridge:
@weissi sorry I forgot to tag you in this yesterday.

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:4.2
+// swift-tools-version:4.1
```
weissi: @ianpartridge  couldn't this just be `4.0` now?

ianpartridge: So on Swift 4.1 we'd build in Swift 4.0 mode?

weissi: @ianpartridge Swift 4.0 mode isn't a thing, there's only `4` (4.0 + 4.1), `4.2` (4.2), and `5` (5.0). Generally new language modes are only for major language revs (apart from 4.2 which is an exception )

ianpartridge:
Thanks @weissi - updated.

weissi:
thanks @ianpartridge

PR #44, opened 2019-04-11
retain SwiftNIO copyright header

tomerd:
motivation: retain copyrights

changes:
* add notice file with references to derived work
* add swift-nio license header to lock


normanmaurer:
late LGTM

PR #45, opened 2019-04-11
fix linux tests formatting

tomerd:
motivation: use consistent formatting

changes: update generate_linux_tests to match the formatting rules we use: `swiftformat --self insert --patternlet inline --stripunusedargs unnamed-only --comments ignore`

### DIFF
--- a/Tests/LoggingTests/TestLogger.swift
+++ b/Tests/LoggingTests/TestLogger.swift
@@ -252,11 +252,11 @@ public class MDC {
     }
 
     private var threadId: Int {
-      #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
-          return Int(pthread_mach_thread_np(pthread_self()))
-      #else
-          return Int(pthread_self())
-      #endif
+        #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+            return Int(pthread_mach_thread_np(pthread_self()))
+        #else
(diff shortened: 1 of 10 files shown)


Tests/LoggingTests/TestLogger.swift
```diff
@@ -252,11 +252,11 @@ public class MDC {
     }
 
     private var threadId: Int {
-      #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
-          return Int(pthread_mach_thread_np(pthread_self()))
-      #else
-          return Int(pthread_self())
-      #endif
+        #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
```
weissi: shouldn't we match the formatting that Xcode does by default? For this one, Xcode does

```
    private var threadId: Int {
        #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
        return Int(pthread_mach_thread_np(pthread_self()))
        #else
        return Int(pthread_self())
        #endif
    }
```

tomerd: i have no opinion on the what format we should use. just want to make sure its consistent. currently we mostly use `swiftformat --self insert --patternlet inline --stripunusedargs unnamed-only --comments ignore` but happy apply other settings if the team prefers something else. just let me know or put a follow up PR to "fix"

PR #46, opened 2019-04-12
Added CaseIterable to Logger.Level enum

dokun1:


Sources/Logging/Logging.swift
```diff
@@ -304,35 +304,35 @@ extension Logger {
     ///
     /// Raw values of log levels correspond to their severity, and are ordered by lowest numeric value (0) being
     /// the most severe. The raw values match the syslog values.
-    public enum Level {
+    public enum Level: String, CaseIterable {
```
weissi: We can't take make this backed by `String`, that would make it slower unfortunately. I'm fine with `CaseIterable`.

If you want a `String` reprpresentation, you can add

```
extension Logger.Level {
    var uppercasedStringRepresentation: String {
        return String(describing: self).uppercased()
    }
}
```

or

```
extension Logger.Level {
    var uppercasedStringRepresentation: String {
        switch self {
            case .trace: return "TRACE"
            case .debug: return "DEBUG"
            case .info: return "INFO"
            case .notice: return "NOTICE"
            case .warning: return "WARNING"
            case .error: return "ERROR"
            case .critical: return "CRITICAL"
    }
}
```

to your package.

dokun1: Good call on the extension - figured I could ask, but easy to remove.

weissi:

weissi:

weissi:

PR #48, opened 2019-04-12
fix docker setup to use release version of swift 5

tomerd:
motivation: swift 5 was already released, use the release version for CI

changes: remove the dev verion from docker compose setup


ianpartridge:
Great, thanks.

Issue #52, opened 2019-04-15
Align .bootstrap() style between SwiftLog and SwiftMetrics


ktoso:
The two recently accepted by the SSWG API packages both use the "bootstrap at system start style" to select a (logger | metrics) backend.

The Logging API requires:

```swift
LoggingSystem.bootstrap(MyLogHandler.init)
```

so exposing the LogHandler directly.

At the same time we say that "log handler is internal detail, don't worry about it."

The style is similar yet notably different in the metrics API:

```swift
MetricsSystem.bootstrap(SelectedMetricsImplementation())
```

as the `SelectedMetricsImplementation` often may / will contain more state than only being a function that creates a stateless metric.

Note as well that one can obtain the factory like `MetricsSystem.factory` which is important in order to implement more advanced features [e.g. removing metrics](https://github.com/apple/swift-metrics/pull/3).

---

I would like to propose re-aligning the style of how one performs bootstrap in both those APIs, so the SSWG APIs have a shared "feel" and meaning to words used -- e.g. that `*Handler` should never really appear in "user land" unless there is good reasons to.

**Proposal:**

- adopt the style that SwiftMetrics have, so we'd `bootstrap(SimpleLoggingFactory())` etc

Another point for this would be that I can definitely imagine using such factory to keep state around such as connections etc, if the logger wanted to send metrics to an external service for gathering etc.

WDYT?

tomerd:
+1

ktoso:
Ok, I'm on it then 👍

ktoso:
Also ping @weissi

ktoso:
I ended up closing the PRs as had no immediate need for them and it would cause some bigger discussions about the "Factory type" which we currently do not need. 

We can revisit this if/when someone encounters an use case they can not workaround using `Logger.init(label:factory:)`.

PR #53, opened 2019-04-15
Expose log handler "factory"

ktoso:
Another "alignment with swift-metrics" issue I suppose, where we did decide to expose the factory.

Relates to: 
- This also relates to the question of making the factory a real type as well: _Align .bootstrap() style between SwiftLog and SwiftMetrics #52_

Reasoning: 

- When implementing a `LogHandler` which wants to *delegate*/*proxy* to "the real log handler," because maybe it is doing something with metadata or maybe it's a testing specialized one which captures all logged messages etc -- but I only set it some executions, not globally;
  - for the testing one one could consider if maybe importing `@testable` and then putting the proxying one there could be okey... but I'm not sure it's the only case
  - the "proxying one" nowadays has to create the `Logger()` and has to emit into it, rather than a to the handler directly (to get the "globally configured by user" logging), but could directly create the handler the user has configured globally and can delegate to it then.
- alignment between swift metrics and swift log: https://github.com/apple/swift-metrics/blob/master/Sources/CoreMetrics/Metrics.swift#L366-L369

Others:

- Should I remove the alias? It relates to #52, where I'd hope we could introduce a real type -- though that's up for debate as well I suppose.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -255,8 +255,10 @@ extension Logger {
 /// configured. `LoggingSystem` is set up just once in a given program to set up the desired logging backend
 /// implementation.
 public enum LoggingSystem {
+    public typealias LogHandlerFactory = (String) -> LogHandler
+
     fileprivate static let lock = ReadWriteLock()
-    fileprivate static var factory: (String) -> LogHandler = StdoutLogHandler.init
+    fileprivate static var _factory: LogHandlerFactory = StdoutLogHandler.init
     fileprivate static var initialized = false
 
     /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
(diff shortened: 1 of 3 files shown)


weissi:
probably fine by me

ktoso:
The alias thing as well or should I remove that part from this PR?

weissi:
> The alias thing as well or should I remove that part from this PR?

I'd probably kill the typealias but don't mind it much ;)

ktoso:
The alias builds up to the question posed in https://github.com/apple/swift-log/issues/52 I guess, so perhaps depends what we think about that issue. :)

tomerd:
@ktoso @weissi is this ready to be merged?

ktoso:
Ok will add; We want to get it in independently from the type change then I suppose https://github.com/apple/swift-log/pull/56 ?

ktoso:
Addressed, small improvement should be good regardless of opinions on #56 I think (which we may decide to not do)

weissi:
ping @moiseev just to see if `factory` is the usual vocabulary here?

moiseev:
I don't have anything against "factory" as a term, but in this particular example, consider the use-site: `LoggingSystem.factory("My Logger")`. Not at all clear without knowing the types and all. Internal field is an implementation detail, but I think it would result in a better API if the public property is called `createLogHandler` or something along these lines, i.e. a name that's suitable for a function, which it is. `LoggingSystem.createLogHandler("My Logger")`.

ktoso:
Yeah, though worth noting that this entire method is more of a "expose a backdoor in case you'd need to reach into factory itself and can not do so otherwise" -- similar to https://github.com/apple/swift-metrics/blob/master/Sources/CoreMetrics/Metrics.swift#L443 where the type is more interesting than a plain function. And to be honest, I also do think the types would become more involved than just a plain function here over time...

At the same time... I looked over my use cases and am currently able to achieve what I needed with the Logger(label: "...", factory: { label in specialHandler })`... so perhaps we dan drop this PR until we hit a serious need which is not work-aroundable with the `Logger.init(label:factory:)`?

ktoso:
The more I think about it the more I think we should stash this and the https://github.com/apple/swift-log/pull/56 PRs for now -- until an use case pops up that would need them. I'll close the PRs for now, but happy to revive if someone needs them.

PR #54, opened 2019-04-15
add a CocoaPods podspec generator

### LINKED ISSUE #19: CocoaPods / Carthage support

weissi:
Motivation:

Users ask for a CocoaPod. Unfortunately, the CocodPod `SwiftLog` is
already taken, so I went for `SwiftLogAPI`.

Modification:

Add a podspec generator.

Result:

- Happier users.
- fixes #19 

---

Example output:

```
Pod::Spec.new do |s|
  s.name = 'SwiftLogAPI'
  s.version = '1.0.0'
  s.license = { :type => 'Apache 2.0', :file => 'LICENSE.txt' }
  s.summary = 'A Logging API for Swift.'
  s.homepage = 'https://github.com/apple/swift-log'
  s.author = 'Apple Inc.'
  s.source = { :git => 'https://github.com/apple/swift-log.git', :tag => s.version.to_s }
  s.documentation_url = 'https://github.com/apple/swift-log'
  s.module_name = 'Logging'

  s.swift_version = '5.0'
  s.cocoapods_version = '>=1.6.0'
  s.ios.deployment_target = '12.0'
  s.osx.deployment_target = '10.14'
  s.tvos.deployment_target = '12.0'

  s.source_files = 'Sources/Logging/**/*.swift'
end
```

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -35,7 +35,8 @@ This product contains a derivation of the Tony Stone's 'process_test_files.rb'.
 
 ---
 
-This product contains a derivation of SwiftNIO locks.
+This product contains a derivation of the lock implementation and various
+scripts from SwiftNIO.
 
   * LICENSE (Apache License 2.0):
     * https://www.apache.org/licenses/LICENSE-2.0
--- a/scripts/build_podspec.sh
+++ b/scripts/build_podspec.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift Logging API open source project
+##
+## Copyright (c) 2019 Apple Inc. and the Swift Logging API project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of Swift Logging API project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+set -eu
+
+function usage() {
+  echo "$0 [-u] version"
+  echo
+  echo "OPTIONS:"
+  echo "  -u: Additionally upload the podspec"
+}
+
+upload=false
+while getopts ":u" opt; do
+  case $opt in
+    u)
+      upload=true
+      ;;
+    \?)
+      usage
+      exit 1
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
+
+if [[ $# -eq 0 ]]; then
+  echo "Must provide target version"
+  exit 1
+fi
+
+version=$1
+podspec_name="Logging"
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+tmpdir=$(mktemp -d /tmp/.build_podspecsXXXXXX)
+echo "Building podspec in $tmpdir"
+
+cat > "${tmpdir}/${podspec_name}.podspec" <<- EOF
+Pod::Spec.new do |s|
+  s.name = '$podspec_name'
+  s.version = '$version'
+  s.license = { :type => 'Apache 2.0', :file => 'LICENSE.txt' }
+  s.summary = 'A Logging API for Swift.'
+  s.homepage = 'https://github.com/apple/swift-log'
+  s.author = 'Apple Inc.'
+  s.source = { :git => 'https://github.com/apple/swift-log.git', :tag => s.version.to_s }
+  s.documentation_url = 'https://apple.github.io/swift-log'
+  s.module_name = 'Logging'
+
+  s.swift_version = '5.0'
+  s.cocoapods_version = '>=1.6.0'
+  s.ios.deployment_target = '8.0'
+  s.osx.deployment_target = '10.9'
+  s.tvos.deployment_target = '9.0'
+  s.watchos.deployment_target = '2.0'
+
+  s.source_files = 'Sources/Logging/**/*.swift'
+end
+EOF
+
+if $upload; then
+  echo "Uploading ${tmpdir}/${podspec_name}.podspec"
+  pod trunk push "${tmpdir}/${podspec_name}.podspec"
+else
+  echo "Linting ${tmpdir}/${podspec_name}.podspec"
+  pod spec lint "${tmpdir}/${podspec_name}.podspec"
+fi
--- a/scripts/sanity.sh
+++ b/scripts/sanity.sh
@@ -13,8 +13,27 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2017-2019 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
 set -eu
 here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+function replace_acceptable_years() {
+    # this needs to replace all acceptable forms with 'YEARS'
+    sed -e 's/2018-2019/YEARS/' -e 's/2019/YEARS/'
+}
 
 printf "=> Checking linux tests... "
 FIRST_OUT="$(git status --porcelain)"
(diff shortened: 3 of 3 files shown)


larryonoff:
Please see below my feedback:

1. I don't see watchOS as a deployment target. Is it supported? I don't think that there're limitations to support watchOS.
2. What about name 'AppleSwiftLog' or 'swift-log' instead of 'SwiftLogAPI'?

weissi:
> Please see below my feedback:
> 
> 1. I don't see watchOS as a deployment target. Is it supported? I don't think that there're limitations to support watchOS.

Oh thanks! Would you mind checking if it works? I have no idea how to build a Watch app :)


> 2. What about name 'AppleSwiftLog' or 'swift-log' instead of 'SwiftLogAPI'?

Could do that, or maybe `SSWGSwiftLog`. @tomerd / @ktoso / @ianpartridge / @tanner0101

ianpartridge:
`SwiftLogAPI` seems fine to me - it makes it clear that this is an API package, not a backend.

ktoso:
A bit unfortunate with the module name already occupied hm... `SwiftLogAPI` sounds fine though.
Would we care to have all API packages in both systems adopt same naming style for this?

scripts/sanity.sh
```diff
@@ -108,9 +127,9 @@ EOF
...
       \( "${matching_files[@]}" \) -a \
       \( \! \( "${exceptions[@]}" \) \) \) | while read line; do
-      if [[ "$(cat "$line" | head -n $expected_lines | shasum)" != "$expected_sha" ]]; then
+      if [[ "$(cat "$line" | replace_acceptable_years | head -n $expected_lines | shasum)" != "$expected_sha" ]]; then
         printf "\033[0;31mmissing headers in file '$line'!\033[0m\n"
-        diff -u <(cat "$line" | head -n $expected_lines) "$tmp"
+        diff -u <(cat "$line" | replace_acceptable_years | head -n $expected_lines) "$tmp"
         exit 1
       fi
     done
```
tomerd: are these changes related?

weissi: Yeah, only with these can I do copyright 2019

tomerd: i was wondering if this is better done in a separate PR, but not a big deal

scripts/build_podspec.sh
```diff
@@ -0,0 +1,91 @@
...
+  s.cocoapods_version = '>=1.6.0'
+  s.ios.deployment_target = '12.0'
+  s.osx.deployment_target = '10.14'
+  s.tvos.deployment_target = '12.0'
+
+  s.source_files = 'Sources/Logging/**/*.swift'
+end
+EOF
+
+if $upload; then
```
tomerd: worth calling `pod spec lint "${tmpdir}/${NAME}.podspec"` here?

scripts/sanity.sh
weissi: @tomerd the PR will stay red unless I do this :)

scripts/build_podspec.sh
weissi: @tomerd it does the linting itself automatically

weissi:
```
[!] No podspec found in directory `.`
Usage:

    $ pod trunk push [PATH]

      Publish the podspec at `PATH` to make it available to all users of the ‘master’
      spec-repo. If `PATH` is not provided, defaults to the current directory.

      Before pushing the podspec to cocoapods.org, this will perform a local lint of
      the podspec, including a build of the library. However, it remains *your*
      responsibility to ensure that the published podspec will actually work for your
      users. Thus it is recommended that you *first* try to use the podspec to
      integrate the library into your demo and/or real application.

      If this is the first time you publish a spec for this pod, you will
      automatically be registered as the ‘owner’ of this pod. (Note that ‘owner’ in
      this case implies a person that is allowed to publish new versions and add other
      ‘owners’, not necessarily the library author.)
```

tomerd: right, but when you do local dev, you dont set `upload` (push) so this help validating imo. no big deal tho if you think it does not add value

weissi: @tomerd I'll actually add it if you don't set `upload`, right?

weissi: that's a good point because then you can validate before actually uploading ;)

tomerd: yes something like https://github.com/tomerd/swift-metrics/blob/feature/podspec/scripts/build_podspec.sh#L87

weissi: done

weissi:
@tanner0101 you also fine with pod name `SwiftLogAPI`?

Sources/Logging/Logging.swift
```diff
@@ -498,8 +498,16 @@ public struct MultiplexLogHandler: LogHandler {
...
+    func write(_ string: String) {
+        string.withCString { ptr in
+            _ = fputs(ptr, stderr)
+        }
+    }
+}
+
 /// Ships with the logging module, really boring just prints something using the `print` function
-internal struct StdoutLogHandler: LogHandler {
+internal struct StderrLogHandler: LogHandler {
```
tomerd: this change intentional as part of this PR?

weissi: omg, definitely not, sorry about that!

weissi: @tomerd done

NOTICE.txt
```diff
@@ -35,7 +35,8 @@ This product contains a derivation of the Tony Stone's 'process_test_files.rb'.
 
 ---
 
-This product contains a derivation of SwiftNIO locks.
+This product contains a derivation of SwiftNIO's locks and its build_podspec.sh
+and sanity.sh scripts.
```
ktoso: And generate_linux_tests.rb I guess, do we have to list them all explicitly?

weissi: good point, I'll just say locks & scripts

weissi: done

allenhumphreys:
Why is the iOS/tvOS deployment target 12 and macOS 10.14? Are there any unavailable APIs in use? (I'm currently using all this source deploying to iOS 11)

weissi:
> Why is the iOS/tvOS deployment target 12 and macOS 10.14? Are there any unavailable APIs in use? (I'm currently using all this source deploying to iOS 11)

Very happy to relax the version here if you tell me what the lowest version is that you could confirm actually working.

allenhumphreys:
@weissi I'll report back shortly

allenhumphreys:
@weissi Here is a [git repo](https://github.com/allenhumphreys/swift-log-deployment-tests) that has a rather large set of Apple targets. It uses a local version of the podspec with the following deployment targets:

```
s.ios.deployment_target = '8.0'
s.osx.deployment_target = '10.9'
s.tvos.deployment_target = '9.0'
s.watchos.deployment_target = '2.0'
```

iOS, tvOS, and watchOS are informed by Xcode 10.2's minimums that the GUI allows. macOS is 10.9 because Swift isn't supported with a lower deployment target.

I did not actually test on the *old* platforms running these versions as that could be a rather complicated endeavor. The key thing is that it compiles correctly with these deployment targets. If users actually have issues with running on those older platform versions, that'll be the most informative to any other changes I'd think.

I did verify functionality on all the latest simulators as well as my mac @ 10.14.

allenhumphreys:
Also wanted to mention there was a lot of conversation about the *Pod* name being `SwiftLogAPI`, but the actual module name in code is still `Logging` which means `import Logging` in actual source code which aligns with SPM integration method. 👍

weissi:
> @weissi Here is a [git repo](https://github.com/allenhumphreys/swift-log-deployment-tests) that has a rather large set of Apple targets. It uses a local version of the podspec with the following deployment targets:

Thank you so much! I updated this PR, please check :).

> Also wanted to mention there was a lot of conversation about the _Pod_ name being `SwiftLogAPI`, but the actual module name in code is still `Logging` which means `import Logging` in actual source code which aligns with SPM integration method. 👍

I mean we can also name the pod `Logging` if that's what you're suggesting, that name is still available. @tomerd / @ianpartridge / @ktoso / @tanner0101 ?

ktoso:
A bit hard for me to comment on package naming -- don't know that ecosystem very well... whichever you think works best 👍

allenhumphreys:
In the pods ecosystem, having module names match the pod names is pretty common if it's possible. It actually took me a few minutes to realize I shouldn't be doing `import SwiftLogAPI`.

shuoli84:
@weissi I believe it is better to use Logging as package name. It will lock the name to prevent other lib use this general name. Sorry for the late comment.

weissi:
Thanks @shuoli84 . @tomerd / @ianpartridge okay with making the CocoaPod just `Logging` then?

ianpartridge:
Fine by me. Really hope we don't have to do CocoaPods for very long!

weissi:
@ianpartridge thanks, I updated the PR to use `Logging`. If @tomerd is happy with this I think I'll pull the trigger and upload it.

tomerd:
no strong feeling here, but is it worth reaching out to the owner of SwiftLog pod and see if he is will wot work with us?

tomerd:
@daltoniam ^^

scripts/build_podspec.sh
```diff
@@ -0,0 +1,94 @@
...
+cat > "${tmpdir}/${podspec_name}.podspec" <<- EOF
+Pod::Spec.new do |s|
+  s.name = '$podspec_name'
+  s.version = '$version'
+  s.license = { :type => 'Apache 2.0', :file => 'LICENSE.txt' }
+  s.summary = 'A Logging API for Swift.'
+  s.homepage = 'https://github.com/apple/swift-log'
+  s.author = 'Apple Inc.'
+  s.source = { :git => 'https://github.com/apple/swift-log.git', :tag => s.version.to_s }
+  s.documentation_url = 'https://github.com/apple/swift-log'
```
tomerd: https://apple.github.io/swift-log

weissi: thanks, done

weissi:
@tomerd the `SwiftLog` pod has already a published version 1.0 so I'm not sure if we should use that (even if we could). What do you propose to go forward here? pod name `Logging`?

tomerd:
> What do you propose to go forward here? pod name Logging?

yes, lets do that

PR #56, opened 2019-04-16
Introduce LogHandlerFactory and align .bootstrap() style with SwiftMetrics #52

ktoso:
Align the style of `SwiftLog` with the other official SSWG API that was recently accepted: https://github.com/apple/swift-metrics

### Depends on:

- `Expose log handler "factory"` #53 

### Motivation:

- align style of library and creating instances to https://github.com/apple/swift-metrics
  - both APIs are very similar, having the same "look and feel" will be beneficial for understandability and adoption
  - also sets some more precedence for other libraries in SSWG to follow this style
- Having an explicit Factory type makes a lot of sense and implementations more complex than print / log to file will likely need some shared "container" for their config or other things; Examples:
  - an async logger would want to _accept_ an event loop or other execution context when configuring the logger, 
  - a networking logging client would take configuration of which endpoint to send data to, this is configured on "the logging library level", and not per handler really, even if per handle may have slight different configs
- Avoid having the APIs be great for the purpose of "our tests" but a bit weird for outside users. 
  - The `LoggingSystem.bootstrap(Handler.make)` and `LoggingSystem.bootstrap(Handler.init)` feels like it is great for "us" since it makes writing weird one off handlers super simple, but in practice we anticipate less of this use case, and more of logging libraries offering their "boostrap with my Factory type please," at least IMHO.

### Modifications:

- introduces an explicit `LogHandlerFactory` type
  - bootstrap methods accept it explicitly

### Result:

- More similar API design to the Metrics API
- Provides default "go to" place for keeping logging library configuration / resources
- Makes exposing the Factory (PR: #53) a bit more natural, even through this is a small use case.
- Resolves https://github.com/apple/swift-log/issues/52

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,30 @@ logger.info("Hello World!")
 
 For further information, please check the [API documentation][api-docs].
 
+### Selecting a logging backend implementation (applications only)
+
+Note: If you are building a library, you don't need to concern yourself with this section. 
+It is the end users of your library (the applications) who will decide which logging backend to use. 
+Libraries should never change the logging implementation as that is something owned by the application.
+
+SwiftLog only provides the logging system API. 
+As an application owner, you may want to select a logging backend different than the built-in minimal standard out 
+implementation in order to make the logging information more useful.
+
+Selecting a backend is done by adding a dependency on the desired backend client implementation and invoking the `LoggingSystem.bootstrap` function at the beginning of the program: 
+
+```swift 
+LoggingSystem.bootstrap(SelectedLoggingImplementation())
+```
+
+This instructs the `LoggingSystem` to install `SelectedLoggingImplementation` (actual name will differ) as the logging backend to use.
+
+As the API has just launched, not many implementations exist yet. If you are interested in implementing one see the "Implementing a logging backend" section below explaining how to do so.
+ 
+List of existing SwiftLog API compatible libraries:
+
+- Your library? [Get in touch!](https://forums.swift.org/c/server)
+
 ## What is an API package?
 
 Glad you asked. We believe that for the Swift on Server ecosystem, it's crucial to have a logging API that can be adopted by anybody so a multitude of libraries from different parties can all log to a shared destination. More concretely this means that we believe all the log messages from all libraries end up in the same file, database, Elastic Stack/Splunk instance, or whatever you may choose.
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -251,14 +251,26 @@ extension Logger {
     }
 }
 
-/// The `LoggingSystem` is a global facility where the default logging backend implementation (`LogHandler`) can be
-/// configured. `LoggingSystem` is set up just once in a given program to set up the desired logging backend
-/// implementation.
-public enum LoggingSystem {
-    public typealias LogHandlerFactory = (String) -> LogHandler
+/// The `LogHandlerFactory` is the bridge between the `LoggingSystem` and the metrics backend implementation.
+/// `LogHandlerFactory` role is to initialize concrete `LogHandler` instances,
+/// it may also contain any additional state it wishes to keep for operating those handlers if it needs to.
+///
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -45,6 +45,30 @@ logger.info("Hello World!")
 
 For further information, please check the [API documentation][api-docs].
 
+### Selecting a logging backend implementation (applications only)
```
ktoso: Imported this section and adjusted from https://github.com/apple/swift-metrics#selecting-a-metrics-backend-implementation-applications-only

Putting more focus on that it's possible to do this and hopefully getting more interest in implementing some backends?

Sources/Logging/Logging.swift
```diff
@@ -251,34 +251,57 @@ extension Logger {
...
+public protocol LogHandlerFactory {
+    func make(label: String) -> LogHandler
+}
+
+//// Note: Other implementations can also offer their factories like this as an extension for easier discoverability.
+//public extension LogHandlerFactory {
+//    /// Convenience for creating simple ad-hoc log handler factories which are easily expressed as simple closures.
+//    static func makeFromClosure(_ closure: @escaping (String) -> LogHandler) -> LogHandlerFactory {
+//        return ClosureLogHandlerFactory(closure)
+//    }
```
ktoso: TODO: Remove this, was playing around if we could provide some convenience but sadly can't put it like that on a protocol... Did I miss any good ideas™ here?

ktoso:
Ehh, sorry for amending this one like 1000 times, kept spotting minor cleanup opportunities.

Sources/Logging/Logging.swift
```diff
@@ -251,34 +251,53 @@ extension Logger {
...
+
 /// The `LoggingSystem` is a global facility where the default logging backend implementation (`LogHandler`) can be
-/// configured. `LoggingSystem` is set up just once in a given program to set up the desired logging backend
-/// implementation.
+/// configured. `LoggingSystem` is set up just once by calling `.bootstrap(LogHandlerFactory)` in a given program to set
+/// up the desired logging backend implementation.
 public enum LoggingSystem {
     fileprivate static let lock = ReadWriteLock()
-    fileprivate static var factory: (String) -> LogHandler = StdoutLogHandler.init
+    fileprivate static var _factory: LogHandlerFactory = StdoutLogHandlerFactory()
```
weissi: this changes public API

Sources/Logging/Logging.swift
```diff
@@ -251,34 +251,53 @@ extension Logger {
     }
 }
 
+/// The `LogHandlerFactory` is the bridge between the `LoggingSystem` and the metrics backend implementation.
+/// `LogHandlerFactory` role is to initialize concrete `LogHandler` instances,
+/// it may also contain any additional state it wishes to keep for operating those handlers if it needs to.
+///
+/// This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
+/// To use the Logging API, please refer to the documentation of `LoggingSystem`.
+public protocol LogHandlerFactory {
```
weissi: why do we need a protocol if it's just one func?

ktoso: Reasons are a bit other than "we really need one to (store / do) X" I guess

- as PR title says, init style alignment between the metrics and logging APIs -- was the main motivation
- code structure "hint" to people implementing logging backends (that's a weaker argument I guess, though hopefully would result in uniform looking "this is how you init library X and this is how library Y")

weissi: @ktoso couldn't we change metrics instead? This would break major version here

weissi: also unconvinced a protocol is the right choice here

ktoso: > @ktoso couldn't we change metrics instead? This would break major version here

Metrics definitely need a "factory" type though, what's your idea?

> also unconvinced a protocol is the right choice here

Open to ideas :)

ktoso:
After some more time figured that might be good to leave sleeping dragons sleep -- closing this until there's a real need for it other than parity of the APIs.

PR #60, opened 2019-04-16
RFC: log to stderr not stdout [might not take this one]

weissi:
Motivation:

Logging to stderr makes more sense as it enables command line utilities
to log.

Modifications:

Log to stderr, not stdout

Result:

more useful

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -256,7 +256,7 @@ extension Logger {
 /// implementation.
 public enum LoggingSystem {
     fileprivate static let lock = ReadWriteLock()
-    fileprivate static var factory: (String) -> LogHandler = StdoutLogHandler.init
+    fileprivate static var factory: (String) -> LogHandler = StderrLogHandler.init
     fileprivate static var initialized = false
 
     /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
@@ -498,8 +498,16 @@ public struct MultiplexLogHandler: LogHandler {
     }
 }
 
+private struct StandardErrorOutputStream: TextOutputStream {
+    func write(_ string: String) {
+        string.withCString { ptr in
+            _ = fputs(ptr, stderr)
+        }
+    }
+}
+
 /// Ships with the logging module, really boring just prints something using the `print` function
-internal struct StdoutLogHandler: LogHandler {
+internal struct StderrLogHandler: LogHandler {
     private let lock = Lock()
 
     public init(label: String) {}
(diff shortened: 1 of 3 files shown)


weissi:
> I like the introduction of the two but I'm not sure if stderr should be the default one to print to 🤔

Do you think stdout should be default? What about command line utilities like

    cat /tmp/image.gif | my-transcoder > /tmp/image.jpg

?

ktoso:
Yeah that may be a good point... It's one of those things where there's no "best" default I guess though... If you prefer stderr and people agree then should be fine 👍 

I also skimmed for some prior art. Results are "it's a mix" 😉 (skimmed java defaults (default stderr, though most popular lib defaults to stdout, configurable; rust has a million mini libs which all do differently, no clear "by default ...").

weissi:
> Yeah that may be a good point... It's one of those things where there's no "best" default I guess though... If you prefer stderr and people agree then should be fine 👍
> 
> I also skimmed for some prior art. Results are "it's a mix" 😉 (skimmed java defaults (default stderr, though most popular lib defaults to stdout, configurable; rust has a million mini libs which all do differently, no clear "by default ...").

Hehe, yes, nobody knows. I guess the best argument for `stdout` is that that's what 1.0.0 does and you could see this as a breaking change. Maybe we'll just leave it as is.

Yasumoto:
I’m a fan of `stderr` by default as I can pipe it away or can just directly pipe my output to another tool without having to add some kind of `grep -v`

Sources/Logging/Logging.swift
```diff
@@ -498,8 +498,16 @@ public struct MultiplexLogHandler: LogHandler {
...
+    func write(_ string: String) {
+        string.withCString { ptr in
+            _ = fputs(ptr, stderr)
+        }
+    }
+}
+
 /// Ships with the logging module, really boring just prints something using the `print` function
-internal struct StdoutLogHandler: LogHandler {
+internal struct StderrLogHandler: LogHandler {
```
tomerd: => i suggest we take an enum here to allow the user the choose between stdout and stderr

=> as for default, its good question!

weissi: @tomerd sure, could do that. `StdioLogHandler(stream: StdioLogHandler.Stream)` with `enum Stream { case stdout; case stderr }`

weissi: I think I like it. Default: dunno :)

tomerd: lets keep the default stdout, CLIs can fairly easily switch to stderr with this design

tomerd:
@weissi should we close this in favor of #61?

weissi:
yes

Issue #51, opened 2019-04-15
StdoutLogHandler should be public


allenhumphreys:
I'm looking at being an early adopter of SwiftLog because it looks good and I'm tired of solving the logging issue every time I start a new project. It occurs to me though, that `StdoutLogHandler` should be `public` but it is marked as `internal`.

The `StdoutLogHandler` is great because it provides out-of-the-box console logging to get up and moving fast and prevents the need to re-implement that, but being marked `internal` means that as soon as the use-case gets a little more interesting, I'd have to write my own console log handlers. The use case I'm referring to really just looks like this:

```
LoggingSystem.bootstrap {
  MultiplexLogHandler([StdoutLogHandler(label: $0), RemoteLogHandler(label: $0)])
}
```

I suppose `MultiplexLogHandler` could have some kind of configurability to implicitly create a `StdoutLogHandler` in addition to whichever log handlers are passed in. This could be an alternative to exposing `StdoutLogHandler` if that is truly the desire.

It's also worth pointing out that `StdoutLogHandler`'s initializer is explicitly marked public. So I'm wondering if this has been an oversight.

I can open a PR to make a change if needed.

ktoso:
Hi Allen, really cool that you're looking to adopt the API, and thanks for chiming in!

As you notice, it is early days here, and not many implementations of this API exist yet. Our hope is that soon someone would implement a much better standard out logger that would support formatting etc. It is true though that the built in one provides a nice out of the box default... 

I think it sounds good on making it public, it seems to have been intended to be so; and it is accessible in any case for using it via "not configuring anything"; We could also address the docs comment on it while we're at it (perhaps something more clear that it's "boring" :-)), mentioning that it is intended to be a bare-minimum implementation and proper log handlers with formatting are expected to be external modules? 

We'll see what @weissi and @tomerd think, but sounds to me like a plan to have it publicly constructable, since we "ship it" and it is "to be used by users" anyway.

weissi:
I'd be okay with that, but we should name it `StderrHandler` and make it log to `stderr` :). It's no good logging to stdout but I forgot to fix it...

ianpartridge:
Why stderr?  https://12factor.net/logs

weissi:
> Why stderr? https://12factor.net/logs

Command line apps are pretty unusable if you log to `stdout`. Imagine you do

    cat myImage.HEIC | my-swift-heic-to-jpg-transcoder > myImage.jpg

[...]

PR #61, opened 2019-04-17
Expose an API for configuring basic stdio log handling

allenhumphreys:
Expose `StreamLogHandler` to the public which replaces the previously internal `StdoutLogHandlers`. Users can choose now the output stream, while `stdout` is the default.

### Motivation:

As discussed on #51 and #60

### Modifications:

Rename `StdoutLogHandler` to `StreamLogHandler` and provide 2 factory methods `standardOutput(label:)` and `standardError(label:)` for using during bootstrapping.

### Result:

Users who adopt other logging backends during `LoggingSystem` bootstrapping will still be able to use the default log handler to get very basic console output to their choice of `stdout` or `stderr`.

### Example:
```
// Log to both stderr and stdout for good measure
LoggingSystem.bootstrap { 
    MultiplexLogHandler([
        FancyLoggingBackend(label: $0),
        StreamLogHandler.standardError(label: $0), 
        StreamLogHandler.standardOutput(label: $0)
    ])
}
```

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -258,7 +258,7 @@ extension Logger {
 /// implementation.
 public enum LoggingSystem {
     fileprivate static let lock = ReadWriteLock()
-    fileprivate static var factory: (String) -> LogHandler = StdioLogHandler.init
+    fileprivate static var factory: (String) -> LogHandler = { StdioLogHandler.init(label: $0) }
     fileprivate static var initialized = false
 
     /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
@@ -507,7 +507,7 @@ public struct MultiplexLogHandler: LogHandler {
 
 /// A wrapper to facilitate `print`-ing to stderr and stdio
 private struct FileOutputStream: TextOutputStream {
-    var file: UnsafeMutablePointer<FILE>
+    let file: UnsafeMutablePointer<FILE>
 
     func write(_ string: String) {
         string.withCString { ptr in
@@ -525,13 +525,11 @@ public struct StdioLogHandler: LogHandler {
 
     private let lock = Lock()
     public let stream: Stream
+    private let output: FileOutputStream
 
-    public init(label: String) {
-        self.init(label: label, stream: .stdout)
-    }
-
-    public init(label: String, stream: Stream) {
+    public init(label: String, stream: Stream = .stdout) {
         self.stream = stream
+        output = FileOutputStream(file: stream == .stdout ? stdout : stderr)
     }
 
     private var _logLevel: Logger.Level = .info
@@ -561,7 +559,8 @@ public struct StdioLogHandler: LogHandler {
         let prettyMetadata = metadata?.isEmpty ?? true
             ? self.prettyMetadata
             : self.prettify(self.metadata.merging(metadata!, uniquingKeysWith: { _, new in new }))
-        var output = FileOutputStream(file: stream == .stdout ? stdout : stderr)
+
+        var output = self.output
         print("\(self.timestamp()) \(level):\(prettyMetadata.map { " \($0)" } ?? "") \(message)", to: &output)
     }
 
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -530,7 +530,7 @@ public struct StdioLogHandler: LogHandler {
         self.init(label: label, stream: .stdout)
     }
 
-    public init(label: String, stream: Stream = .stdout) {
+    public init(label: String, stream: Stream) {
         self.stream = stream
     }
 
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -508,10 +508,10 @@ public struct MultiplexLogHandler: LogHandler {
 /// A wrapper to facilitate `print`-ing to stderr and stdio that
 /// ensures access to the underlying `FILE` is locked to prevent
 /// cross-thread interleaving of output.
-private struct FileOutputStream: TextOutputStream {
-    let file: UnsafeMutablePointer<FILE>
+public struct StdioOutputStream: TextOutputStream {
+    internal let file: UnsafeMutablePointer<FILE>
 
-    func write(_ string: String) {
+    public func write(_ string: String) {
         string.withCString { ptr in
             flockfile(file)
             defer {
@@ -520,6 +520,9 @@ private struct FileOutputStream: TextOutputStream {
             _ = fputs(ptr, file)
         }
     }
+
+    public static let stderr = StdioOutputStream(file: systemStderr)
+    public static let stdout = StdioOutputStream(file: systemStdout)
 }
 
 // Prevent name clashes
@@ -537,32 +540,8 @@ let systemStdout = Glibc.stdout!
 /// This powers `Logger`'s out-of-the-box behavior of outputing logs to stdout.
 public struct StreamLogHandler: LogHandler {
 
-    /// `Stream` wraps a `TextOutputStream` so that `StreamLogHandler` can
-    /// `print` directly to its stream.
-    ///
-    /// TODO: <How to provide your own, safely with a simple example>
-    ///
-    /// The static properties `stderr` and `stdout` abstract the system-dependent
-    /// implementations for outputting to the standard process-level outputs
-    public struct Stream: TextOutputStream {
-
-        private let underlying: TextOutputStream
-
-        public init(underlying: TextOutputStream) {
-            self.underlying = underlying
-        }
-
-        public mutating func write(_ string: String) {
-            var underlying = self.underlying
-            underlying.write(string)
-        }
-
-        public static let stderr: Stream = .init(underlying: FileOutputStream(file: systemStderr))
-        public static let stdout: Stream = .init(underlying: FileOutputStream(file: systemStdout))
-    }
-
-    /// The instance of `Stream` to which this handler will output
-    public let stream: Stream
+    /// The instance of `TextOutputStream` to which this handler will write to
+    public let stream: TextOutputStream
 
     public var logLevel: Logger.Level = .info
 
@@ -582,7 +561,7 @@ public struct StreamLogHandler: LogHandler {
         }
     }
 
-    public init(label: String, stream: Stream = .stdout) {
+    public init(label: String, stream: TextOutputStream = StdioOutputStream.stdout) {
         self.stream = stream
     }
 
@@ -595,7 +574,7 @@ public struct StreamLogHandler: LogHandler {
             : self.prettify(self.metadata.merging(metadata!, uniquingKeysWith: { _, new in new }))
 
         var stream = self.stream
-        print("\(self.timestamp()) \(level):\(prettyMetadata.map { " \($0)" } ?? "") \(message)", to: &stream)
+        stream.write("\(timestamp()) \(level):\(prettyMetadata.map { " \($0)" } ?? "") \(message)\n")
     }
 
     private func prettify(_ metadata: Logger.Metadata) -> String? {
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -258,7 +258,7 @@ extension Logger {
 /// implementation.
 public enum LoggingSystem {
     fileprivate static let lock = ReadWriteLock()
-    fileprivate static var factory: (String) -> LogHandler = { StdioLogHandler.init(label: $0) }
+    fileprivate static var factory: (String) -> LogHandler = { StreamLogHandler.init(label: $0) }
     fileprivate static var initialized = false
 
     /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
@@ -505,7 +505,9 @@ public struct MultiplexLogHandler: LogHandler {
     }
 }
 
-/// A wrapper to facilitate `print`-ing to stderr and stdio
+/// A wrapper to facilitate `print`-ing to stderr and stdio that
+/// ensures access to the underlying `FILE` is locked to prevent
+/// cross-thread interleaving of output.
 private struct FileOutputStream: TextOutputStream {
     let file: UnsafeMutablePointer<FILE>
 
@@ -520,14 +522,49 @@ private struct FileOutputStream: TextOutputStream {
     }
 }
 
-/// Ships with the logging module, really boring just prints something using the `print` function
-public struct StdioLogHandler: LogHandler {
+// Prevent name clashes
+#if os(macOS) || os(tvOS) || os(iOS) || os(watchOS)
+let systemStderr = Darwin.stderr
+let systemStdout = Darwin.stdout
+#else
+let systemStderr = Glibc.stderr!
+let systemStdout = Glibc.stdout!
+#endif
 
-    public enum Stream {
-        case stdout, stderr
+/// `StreamLogHandler` is a simple implementation of `LogHandler` for directing
+/// `Logger` output to a `TextOutputStream`.
+///
+/// This powers `Logger`'s out-of-the-box behavior of outputing logs to stdout.
+public struct StreamLogHandler: LogHandler {
+
+    /// `Stream` wraps a `TextOutputStream` so that `StreamLogHandler` can
+    /// `print` directly to its stream.
+    ///
+    /// TODO: <How to provide your own, safely with a simple example>
+    ///
+    /// The static properties `stderr` and `stdout` abstract the system-dependent
+    /// implementations for outputting to the standard process-level outputs
+    public struct Stream: TextOutputStream {
+
+        private let underlying: TextOutputStream
+
+        public init(underlying: TextOutputStream) {
+            self.underlying = underlying
+        }
+
+        public mutating func write(_ string: String) {
+            var underlying = self.underlying
+            underlying.write(string)
+        }
+
+        public static let stderr: Stream = .init(underlying: FileOutputStream(file: systemStderr))
+        public static let stdout: Stream = .init(underlying: FileOutputStream(file: systemStdout))
     }
 
+    /// Lock for accessing underlying properties, does not provide locking behavior for output
     private let lock = Lock()
+
+    /// The instance of `Stream` to which this handler will output
     public let stream: Stream
 
     public init(label: String, stream: Stream = .stdout) {
@@ -562,8 +599,8 @@ public struct StdioLogHandler: LogHandler {
             ? self.prettyMetadata
             : self.prettify(self.metadata.merging(metadata!, uniquingKeysWith: { _, new in new }))
 
-        var output = FileOutputStream(file: stream == .stdout ? stdout : stderr)
-        print("\(self.timestamp()) \(level):\(prettyMetadata.map { " \($0)" } ?? "") \(message)", to: &output)
+        var stream = self.stream
+        print("\(self.timestamp()) \(level):\(prettyMetadata.map { " \($0)" } ?? "") \(message)", to: &stream)
     }
 
     public var metadata: Logger.Metadata {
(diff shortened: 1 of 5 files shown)


tomerd:

tomerd:
we may want to finalize https://github.com/apple/swift-log/pull/60 before making it public?

tomerd:
@allenhumphreys would you mind changing this PR following the suggestion in https://github.com/apple/swift-log/pull/60#discussion_r276393436, ie create a public generic `StdioLogHandler` that can be configured to log to either stdout or stderr

allenhumphreys:
@tomerd I can give it a shot. I briefly looked into the correct way to do this in Swift the other day, `stdout` and `stderr` are `UnsafeMutablePointer<FILE>`. I think I can effectively just replace `print(...)` with `fputs(..., stderr|stdout)`. That about the gist of it?

allenhumphreys:
Digging into the [stdlib](https://github.com/apple/swift/blob/master/stdlib/public/core/Print.swift#L217) a little bit.. using something like:
```
extension UnsafeMutablePointer: TextOutputStream where Pointee == FILE {
    public mutating func write(_ string: String) {
        fputs(string, self)
    }
}
```

with 

```
print(..., to: &stderr)
```

might be better but I don't know why you can't `print` to `stdout` as part of the stdlib in the first place.

tomerd:
@abdullahselek way about what @weissi suggested in https://github.com/apple/swift-log/pull/60/files#diff-259e1bfbf733f7f456500e470bf358cdR501

allenhumphreys:
Oh good point, I hadn't seen that portion of it! Thank you!

Sources/Logging/Logging.swift
```diff
@@ -498,11 +498,34 @@ public struct MultiplexLogHandler: LogHandler {
...
+public struct StdioLogHandler: LogHandler {
+
+    public enum Stream {
+        case stdout, stderr
+    }
+
     private let lock = Lock()
+    public let stream: Stream
+
+    public init(label: String) {
```
tomerd: this actor seems redundant given the other one has a default?

allenhumphreys: To better express that and prevent the knee-jerk reaction you had, I think I'll remove the default from the other one, calling conventions will remain the same.

Sources/Logging/Logging.swift
```diff
@@ -531,7 +554,8 @@ public struct StdoutLogHandler: LogHandler {
         let prettyMetadata = metadata?.isEmpty ?? true
             ? self.prettyMetadata
             : self.prettify(self.metadata.merging(metadata!, uniquingKeysWith: { _, new in new }))
-        print("\(self.timestamp()) \(level):\(prettyMetadata.map { " \($0)" } ?? "") \(message)")
+        var output = FileOutputStream(file: stream == .stdout ? stdout : stderr)
```
tomerd: creating a new instance of FileOutputStream on every log call seems expensive? since stream is mutable, maybe move to ctor?

Sources/Logging/Logging.swift
```diff
@@ -498,11 +498,34 @@ public struct MultiplexLogHandler: LogHandler {
     }
 }
 
+/// A wrapper to facilitate `print`-ing to stderr and stdio
+private struct FileOutputStream: TextOutputStream {
+    var file: UnsafeMutablePointer<FILE>
```
tomerd: should this be let instead of var?

Sources/Logging/Logging.swift
allenhumphreys: I had the same though on this, but was following the setup from the other PR. I'll revisit

tomerd: yup. its used in as an inout param in L558 so needs to be a local var, but i am not sure if it'd be cheaper to create on the ctor and take a local copy or this is fine as is. cc @weissi maybe he has some ideas

Sources/Logging/Logging.swift
allenhumphreys: Yes!

Sources/Logging/Logging.swift
allenhumphreys: I changed it! Take a look

allenhumphreys:
@tomerd Tests will have to come on a different day!

Sources/Logging/Logging.swift
allenhumphreys: My screen didn't refresh, just saw your latest comment. I have the same unsure-ness about the way I had it vs the way I changed it to regarding performance impact.

weissi: @tomerd / @allenhumphreys `FileOutputStream` is a `struct` that holds only a pointer. Creating this there is just fine and free like putting an integer there :).

In other words, the code up there is exactly as expensive as

    var output = stream == .stdout ? 1 : 2

Sources/Logging/Logging.swift
```diff
@@ -498,11 +498,30 @@ public struct MultiplexLogHandler: LogHandler {
     }
 }
 
+/// A wrapper to facilitate `print`-ing to stderr and stdio
+private struct FileOutputStream: TextOutputStream {
+    let file: UnsafeMutablePointer<FILE>
+
+    func write(_ string: String) {
+        string.withCString { ptr in
+            _ = fputs(ptr, file)
```
weissi:
```suggestion
            flockfile(file)
            defer {
                funlockfile(file)
            }
            _ = fputs(ptr, file)
```

one thing I only notice now (sorry) is that Swift's `print` function uses `flockfile`/`funlockfile` to make sure that the prints from different threads don't interleave. So I think we should make this function as the above suggestion.

allenhumphreys: Thanks for going deeper on that, I saw that `Swift.print` attempts to [lock something](https://github.com/apple/swift/blob/master/stdlib/public/core/Print.swift#L225) on the output stream, but now that I went further I see that the `_lock`. I had originally assumed the the lock was standard library magic and did not pursue, bad assumption in this case.

I'll make this change 👍

tomerd:

tomerd:
@allenhumphreys looks like tests are failing on linux. you can run them locally with `docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.1804.50.yaml run test`

Sources/Logging/Logging.swift
```diff
@@ -498,11 +498,34 @@ public struct MultiplexLogHandler: LogHandler {
...
+        }
+    }
+}
+
 /// Ships with the logging module, really boring just prints something using the `print` function
-internal struct StdoutLogHandler: LogHandler {
+public struct StdioLogHandler: LogHandler {
+
+    public enum Stream {
```
weissi: @allenhumphreys just as a question if you think this is useful. We could also do the following here: Instead of an `enum Stream` we could do

```swift
// this is just to prevent name clashes further down
#if os(macOS) || os(tvOS) || os(iOS) || os(watchOS)
let sysStderr = Darwin.stderr
let sysStdout = Darwin.stdout
#else
let sysStderr = Glibc.stderr
let sysStdout = Glibc.stdout
#endif

public struct Stream {
    private let underlying: UnsafePointer<FILE>

    internal init(underlying: UnsafePointer<FILE>) {
        self.underlying = underlying
    }

    public static let stderr: Stream = .init(underlying: sysStderr)
    public static let stdout: Stream = .init(underlying: sysStdout)
}
```

that way, the API is the same for the user: `StdioLogHandler(label: "foo", stream: .stdout)` will still work but we have two extra benefits:

1. in tests, we can use `StdioLogHandler(label: "foo", stream: .init(underlying: fopen(".....", "w+")))`
2. if we chose to allow other streams in the future, then we can add them _without breaking public API_. The problem with `enum`s is: Adding a case breaks the API because someone might have done an exhaustive `switch` over it. The `struct` + `static let` solution from above alleviates that issue.

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -417,4 +417,75 @@ class LoggingTest: XCTestCase {
...
+        let messageSucceeded = interceptedString?.hasSuffix(testString) ?? false
+
+        XCTAssertTrue(messageSucceeded)
+    }
+}
+
+extension UnsafeMutablePointer where Pointee == FILE {
+
+    /// Intercepts data written to a file descriptor during the `intercepted` closure
+    func intercept(_ intercepted: () -> Void) -> Data {
```
weissi: @allenhumphreys thanks so much for actually testing this (despite it being a pain). I'd be happy with this but I propose an alternative solution up there, let me know what you think about that.

Sources/Logging/Logging.swift
allenhumphreys: Thanks @weissi , this is good, but I'd like to propose going a little further. 
1. Make `Stream` wrap `TextOutputStream` instead
2. Rename `StdioLogHandler` to `StreamLogHandler`
3. Go ahead and expose the initializer for `Stream`

It'd look something like this:
```
public struct StreamLogHandler: LogHandler {

    public struct Stream: TextOutputStream {

        private let underlying: TextOutputStream

        public init(underlying: TextOutputStream) {
            self.underlying = underlying
        }

        public mutating func write(_ string: String) {
            var underlying = self.underlying
            underlying.write(string)
        }

        public static let stderr: Stream = .init(underlying: FileOutputStream(file: sysStderr))
        public static let stdout: Stream = .init(underlying: FileOutputStream(file: sysStdout))
    }

    private let lock = Lock()
    public let stream: Stream

    public init(label: String, stream: Stream = .stdout) {
        self.stream = stream
    }

...
...
}
```

`FileOutputStream` would remain internal, but clearly available as a reference for consumers. This would allow people to immediately swap in their own streams. It eliminates the need to work with Pointers, or even the c standard library at all. `Stream` ends up's up just being a box and a convenient way to get access to `stderr` and `stdout` in a way that hides the `UnsafeMutablePointer<FILE>` stuff.

Thoughts? Too far?

weissi: @allenhumphreys Thanks, that sounds like a great idea to me. I think you can remove the `Lock` because if the user wants locking, they could implement that themselves, no?

ktoso: Seems like a good idea 👍

ktoso: > @allenhumphreys Thanks, that sounds like a great idea to me. I think you can remove the Lock because if the user wants locking, they could implement that themselves, no?

Is that a safe default? in case a concurrent program would use the default loggers shipped?

weissi: @ktoso I'd leave it up to the `TextOutputStream` to lock. That's at least what default `print` does: It uses a `TextOutputStream` that uses `f(un)lockfile`.

allenhumphreys: If we instead keep `init(underlying: TextOutputStream)` internal for this PR, could we punt on the locking question?

allenhumphreys: I suppose if we did that, the `FileOutputStream` is still handling the locking, which means removing it at the log handler level should happen either way. We could have a `LockingTextOutputStream` to make sure consumers of the API know that they should provide locking to avoid interleaving. Or having something like `init(underlying: TextOutputStream & Lockable)`

tomerd: +1 on removing the lock and leaving it as a concern of the `TextOutputStream`. the one we ship, `FileOutputStream`, should obv do the right thing, but if users choose to write their own they should know that they need to deal with concurrency, and we should document this at the API + readme level

allenhumphreys: @tomerd @weissi @ktoso Finally looked at making the change. The lock in question was never actually used to lock the output, it is used to lock access to the properties on the `StreamLogHandler` itself, e.g. `metadata` and `logLevel`. `StreamLogHandler.Stream` will not do any locking of its output, the underlying `TextOutputStream` can do that. I'll make callouts for API adopters that they should consider locking if they implement their own.

allenhumphreys:
@tomerd @ktoso @weissi I have updated the implementation, the documentation I've proposed isn't complete yet, but I've laid out the general direction. I've left the locking behavior the same per my last comment on the thread.

Thanks everyone for continuing to help evolve the direction of this change.

Sources/Logging/Logging.swift
```diff
@@ -505,11 +505,71 @@ public struct MultiplexLogHandler: LogHandler {
...
+    }
+}
+
+// Prevent name clashes
+#if os(macOS) || os(tvOS) || os(iOS) || os(watchOS)
+let systemStderr = Darwin.stderr
+let systemStdout = Darwin.stdout
+#else
+let systemStderr = Glibc.stderr!
+let systemStdout = Glibc.stdout!
```
allenhumphreys: @ktoso You recommended the local definitions of the system `stderr` but I found that `Glibc.stderr` is an IOU. I'm not sure why the semantics are different between Glibc and Darwin, but I've added bangs here to resolve the incongruity, not sure if you have another recommendation.

allenhumphreys: Sorry, I mean @weissi above

tomerd:

Sources/Logging/Logging.swift
weissi: @allenhumphreys ah, you can remove that lock. `LogHandler`s are supposed to have value semantics for the configuration (metadata & logLevel) and indeed your handler has, so there's no need to lock.

allenhumphreys: @weissi Thanks for pointing that out, it looks like the lock was a holdover from when `StdoutLogHandler` was a class, changed at https://github.com/apple/swift-log/commit/1d983153b45a6b6c90a8ebfab58d4d3e23350949

weissi: @allenhumphreys thank you! Apologies, for some reason I thought you added the lock. I'd propose to just remove it because it's baggage from the past :).

Sources/Logging/Logging.swift
```diff
@@ -538,27 +572,9 @@ internal struct StdoutLogHandler: LogHandler {
...
-        get {
-            return self.lock.withLock { self._metadata[metadataKey] }
-        }
-        set {
-            self.lock.withLock {
-                self._metadata[metadataKey] = newValue
-            }
-        }
+        var stream = self.stream
+        stream.write("\(timestamp()) \(level):\(prettyMetadata.map { " \($0)" } ?? "") \(message)\n")
```
allenhumphreys: @weissi I found that `Swift.print` actually calls `TextOutpuStream.write` multiple times. Looking at the stdlib implementation, they're able to ensure the underlying operations are performed with a single locking action (see https://github.com/apple/swift/blob/master/stdlib/public/core/Print.swift#L224-L231)

To ensure our lock works, I've switched to calling `write` directly. I've appended a newline to the message. Any fundamental issues you might see with that?

Sources/Logging/Logging.swift
allenhumphreys: In light of discovering how `Swift.print` functions, I tweaked the API again. I removed the wrapper type `StreamLogHandler.Stream` and exposed `StdioOutputStream` to allow easily switching between `stderr` and `stdio`. `StdioOutputStream` is not publicly instantiable, though, so the usage is still restricted. I believe this API is now much more straightforward as it just uses `TextOutputStream`!

Sources/Logging/Logging.swift
weissi: @allenhumphreys that looks good to me! Thank you

Sources/Logging/Logging.swift
```diff
@@ -505,30 +505,64 @@ public struct MultiplexLogHandler: LogHandler {
     }
 }
 
-/// Ships with the logging module, really boring just prints something using the `print` function
-internal struct StdoutLogHandler: LogHandler {
-    private let lock = Lock()
+/// A wrapper to facilitate `print`-ing to stderr and stdio that
+/// ensures access to the underlying `FILE` is locked to prevent
+/// cross-thread interleaving of output.
+public struct StdioOutputStream: TextOutputStream {
```
weissi: The `StdioOutputStream` we should leave `internal` or `private` as that's a thing the stdlib should really provide.

Sources/Logging/Logging.swift
weissi: @allenhumphreys perfect. I think it you make the output streams non-public, then we can merge this.

Sources/Logging/Logging.swift
allenhumphreys: Yeah, I have a similar feeling but keeping it private requires some kind of additional wrapper type or switch to power `StreamLogHandler`'s public initializer (fundamentally providing the consumer the ability to switch between stderr and stdout). 

As I had it before `StreamLogHandler.Stream` also conformed to `TextOutputStream`.

Previously it was
```
public static let stderr: Stream = .init(underlying: FileOutputStream(file: systemStderr))
```

now this has the same level of access:
```
public static let stdout = StdioOutputStream(file: systemStdout)
```

but with less indirection.

Ultimately _something_ has to be public in that initializer because one of the goals of this change is to allow easily redirected to `stderr`. Alternatively, to keep the type private, I could have multiple specialized initializers or possibly static factory methods to provide this functionality.

Sources/Logging/Logging.swift
weissi: bangs are fine

Sources/Logging/Logging.swift
weissi: @allenhumphreys if you overload the `init` of `StreamLogHandler` instead of using default arguments, it'll work. All the streams stuff can stay `internal`/`private`. So

```swift
public struct StreamLogHandler { 
    ...
    public init(label: String) {
        self.init(stream: StdioOutputStream(stream: sysStdout))
    }

    public init(label: String, stream: TextOutputStream) {
        ...
    }
    ...
}
```

allenhumphreys: That doesn't leave a good name for an init that goes to `stderr`, any suggestion good or not as good?
Just thinking out loud:
```
init(toStdoutWithLabel label: String)
init(toStderrWithLabel label: String)
```

weissi: @allenhumphreys Ahh, I see! I'd then go with

```swift
public struct StreamLogHandler {
    public static func makeStdoutLogHandler(label: String) -> StreamLogHandler { ... }
    public static func makeStderrLogHandler(label: String) -> StreamLogHandler { ... }

   ...
}
```

and then we can even leave the `init` private

weissi:

weissi:
@ktoso / @tomerd / @ianpartridge I'm happy exactly as is, would you mind having a quick look? Just because there were a lot of changes.

Sources/Logging/Logging.swift
```diff
@@ -505,30 +505,72 @@ public struct MultiplexLogHandler: LogHandler {
...
+#endif
 
-    private var _logLevel: Logger.Level = .info
+/// `StreamLogHandler` is a simple implementation of `LogHandler` for directing
+/// `Logger` output to either `stderr` or `stdout` via the factory methods.
+public struct StreamLogHandler: LogHandler {
 
-    public var logLevel: Logger.Level {
+    /// Factory that makes a `StreamLogHandler` to directs its output to `stdout`
+    public static func makeStdoutLogHandler(label: String) -> StreamLogHandler {
```
tomerd: naming nit: i like the new factories but dont love the names. maybe `makeStdout` and `makeStderr` is enough. in any case not a big deal, so can go with this too

allenhumphreys: You're right, it is a bit repetitive in the full context `StreamLogHandler.makeStdoutLogHandler`, happy to change. While you're considering names, I made up the `StreamLogHandler` name to convey it's new expanded capabilities, but it could just as well be `TextOutputStreamLogHandler` :-)

tomerd: naming is hard. @weissi any preferences?

weissi: @tomerd not really but I agree with both of you that `makeStdoutLogHandler` is a bit long and redundant. We could even do something like `standardOutput` only. Then it would read

    Logging.bootstrap(StreamLogHandler.standardOutput)

`makeStdout` sounds like it's making a 'stdout' to me...

weissi:

tomerd:
<3

PR #78, opened 2019-06-25
Fix documented ordering of Level

kevints:
Fix documented ordering of Level.

### Motivation:

Levels are actually ordered from least severe to most severe, and their raw values are not exposed.

### Modifications:

Fix the documented ordering of `Level`.

### Result:

The documentation matches the implementation.


weissi:

weissi:

PR #84, opened 2019-07-30
test formatting issues as part of sanity check

tomerd:
motivaiton: unified format

changes:
* use official docker image 🎉
* fix outstanding formatting issues
* add a call to swiftformat as part of sanity script
* fix sanity script language check debugging statements

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -13,9 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
-    import Darwin
+import Darwin
 #else
-    import Glibc
+import Glibc
 #endif
 
 /// A `Logger` is the central type in `SwiftLog`. Its central function is to emit log messages using one of the methods
@@ -527,11 +527,11 @@ internal struct StdioOutputStream: TextOutputStream {
 
 // Prevent name clashes
 #if os(macOS) || os(tvOS) || os(iOS) || os(watchOS)
-    let systemStderr = Darwin.stderr
-    let systemStdout = Darwin.stdout
+let systemStderr = Darwin.stderr
+let systemStdout = Darwin.stdout
 #else
-    let systemStderr = Glibc.stderr!
-    let systemStdout = Glibc.stdout!
+let systemStderr = Glibc.stderr!
+let systemStdout = Glibc.stdout!
 #endif
 
 /// `StreamLogHandler` is a simple implementation of `LogHandler` for directing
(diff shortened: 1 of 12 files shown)


Sources/Logging/Logging.swift
```diff
@@ -527,17 +527,16 @@ internal struct StdioOutputStream: TextOutputStream {
...
-let systemStderr = Darwin.stderr
-let systemStdout = Darwin.stdout
+    let systemStderr = Darwin.stderr
+    let systemStdout = Darwin.stdout
 #else
-let systemStderr = Glibc.stderr!
-let systemStdout = Glibc.stdout!
+    let systemStderr = Glibc.stderr!
+    let systemStdout = Glibc.stdout!
 #endif
```
ktoso: nit / question: I was under the impression "official formatting" for #if is to not indent? 

Either way, happy with having an automatted formatted so whichever we want is fine, just confused about this one.

Sources/Logging/Logging.swift
```diff
@@ -555,16 +554,16 @@ public struct StreamLogHandler: LogHandler {
...
     }
 
     public subscript(metadataKey metadataKey: String) -> Logger.Metadata.Value? {
         get {
-            return metadata[metadataKey]
+            return self.metadata[metadataKey]
         }
         set {
-            metadata[metadataKey] = newValue
+            self.metadata[metadataKey] = newValue
```
ktoso: 👍

Sources/Logging/Logging.swift
tomerd: personally don't have a strong opinion on concrete format rules. what we agreed on in the past is capture in the repo's `.swiftformat` file, so we can change it if we want to agree on a different set of rules

ktoso: Yeah same, though this one is surprising... curious if choice was conscious on the `#ifs` @weissi ? As mentioned, not an issue and LGTM as I'm more than happy to have automated formatter

weissi: I'd go with whatever Xcode does just because it's a pain otherwise because it'll re-indent in the way it thinks it's right.

docker/Dockerfile
```diff
@@ -1,56 +1,35 @@
-ARG ubuntu_version=18.04
-FROM ubuntu:$ubuntu_version
+ARG swift_version=5.0
```
weissi: 🙌

Sources/Logging/Logging.swift
tomerd: https://github.com/apple/swift-log/pull/84/commits/967999c856cc497c0d6831e1b99db01fd17a03ce

Issue #86, opened 2019-08-06
default logger flushing behavior 


tomerd:
### Expected behavior

log statements are immediately written to stream, so if a process crashes we dont lose logs

### Actual behavior

default stream based logger has default flush behavior which is not immediate 

somewhat related: https://bugs.swift.org/browse/SR-1127

tomerd:
resolved via #87

PR #87, opened 2019-08-09
add flush api to default stdout/stderr logger

tomerd:
motivation: default logger may not flush the streams on crush, so need to make flushing moe explicit

changes:
* add a constructor argument to control flushing behaviour, default to "everytime"
* add a "flush" method to allow user defined flush timing. this api is now internal but we could expose it if there is a need
* add tests

### DIFF
--- a/Tests/LoggingTests/LoggingTest.swift
+++ b/Tests/LoggingTests/LoggingTest.swift
@@ -453,7 +453,7 @@ class LoggingTest: XCTestCase {
 
     func testStdioOutputStreamFlush() {
         // flush on every statement
-        self.testStdioOutputStreamFlush { writeFD, readFD, readBuffer in
+        self.withWriteReadFDsAndReadBuffer { writeFD, readFD, readBuffer in
             let logStream = StdioOutputStream(file: writeFD, flushMode: .always)
             LoggingSystem.bootstrapInternal { StreamLogHandler(label: $0, stream: logStream) }
             Logger(label: "test").critical("test")
@@ -466,7 +466,7 @@ class LoggingTest: XCTestCase {
             XCTAssertEqual(size2, -1, "expected no flush")
         }
         // default flushing
-        self.testStdioOutputStreamFlush { writeFD, readFD, readBuffer in
+        self.withWriteReadFDsAndReadBuffer { writeFD, readFD, readBuffer in
             let logStream = StdioOutputStream(file: writeFD, flushMode: .undefined)
             LoggingSystem.bootstrapInternal { StreamLogHandler(label: $0, stream: logStream) }
             Logger(label: "test").critical("test")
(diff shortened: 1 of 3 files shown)


tomerd:
solved #86

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -444,4 +450,66 @@ class LoggingTest: XCTestCase {
...
+            let size2 = read(readFD, readBuffer, 256)
+            XCTAssertGreaterThan(size2, -1, "expected flush")
+        }
+    }
+
+    func testStdioOutputStreamFlush(_ body: (UnsafeMutablePointer<FILE>, CInt, UnsafeMutablePointer<Int8>) -> Void) {
+        var fds: [Int32] = [-1, -1]
+        fds.withUnsafeMutableBufferPointer { ptr in
+            let err = pipe(ptr.baseAddress!)
+            assert(err == 0, "pipe: \(err)")
```
weissi: because it's in a test case, you can use `XCTAssertEqual(0, err)`

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -444,4 +450,66 @@ class LoggingTest: XCTestCase {
...
+
+        let writeFD = fdopen(fds[1], "w")
+        let writeBuffer = UnsafeMutablePointer<Int8>.allocate(capacity: 256)
+        defer {
+            writeBuffer.deinitialize(count: 256)
+            writeBuffer.deallocate()
+        }
+
+        var err = setvbuf(writeFD, writeBuffer, _IOFBF, 256)
+        assert(err == 0)
```
weissi: same here, `XCTAssertEqual`

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -444,4 +450,66 @@ class LoggingTest: XCTestCase {
...
+
+        let readFD = fds[0]
+        err = fcntl(readFD, F_SETFL, fcntl(readFD, F_GETFL) | O_NONBLOCK)
+        assert(err == 0)
```
weissi: and here

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -444,4 +450,66 @@ class LoggingTest: XCTestCase {
...
+            let size = read(readFD, readBuffer, 256)
+            XCTAssertEqual(size, -1, "expected no flush")
+
+            logStream.flush()
+            let size2 = read(readFD, readBuffer, 256)
+            XCTAssertGreaterThan(size2, -1, "expected flush")
+        }
+    }
+
+    func testStdioOutputStreamFlush(_ body: (UnsafeMutablePointer<FILE>, CInt, UnsafeMutablePointer<Int8>) -> Void) {
```
weissi: this method should start with the word `with` because it follows the usual `with*` rules: it provides you something that's valid only within the supplied closure. A better name would be

`func withWriteReadFDsAndReadBuffer(...)`

tomerd:
@weissi addressed

PR #103, opened 2019-11-07
Add amzn/smoke-http to the list of compatible libraries

yim-lee:
Motivation:

[amzn/smoke-http](https://github.com/amzn/smoke-http) has adopted swift-log.

Modifications:
- Add amzn/smoke-http to the list of compatible libraries.
- Rearrange list in alphabetic order.

Result:
An expanded list of compatible libraries.


tomerd:

tomerd:
hmm actually looking more into this, i think smoke *uses* it, but do not offer a backend implementation?

PR #108, opened 2019-12-15
[feat] StreamLogHandler now adds label to log output

### LINKED ISSUE #57: Should StdoutLogHandler include label in log output?
### Expected behavior

When logging using different loggers, e.g.:

```
let one = Logger(label: "one")
let two = Logger(label: "two")
```

I would expect the labels `one` and `two` to appear in log output.
This is not the case with the default `StdoutLogHandler`:

```swift
internal struct StdoutLogHandler: LogHandler {
    private let lock = Lock()

    public init(label: String) {} // does not store label
```

I know the StdOutLogHandler is meant to be _super minimal_ but this seems more like an omission?

### Actual behavior

Label is not included in log statements logged through to default std out handler.

prafsoni:
`StreamLogHandler` now includes `label` in the log output

### Motivation:

In my opinion, It's the most basic feature missing even for a test `LogHandler`. More details here https://github.com/apple/swift-log/issues/57

### Modifications:

Store the `label` in `StreamLogHandler` which was previously ignored don't know the rationale behind it. Since it's mandatory to pass a label at the time of initialization it makes sense to store it. Finally, include it in the log output right after the log level.

### Result:

`StreamLogHandler` will include `label` in the log output and resolves #57 

Sample output:- 

Before: `2019-12-15T17:09:32-0600 info: info`
After: `2019-12-15T17:09:32-0600 info test : info`

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -578,6 +578,7 @@ public struct StreamLogHandler: LogHandler {
     }
 
     private let stream: TextOutputStream
+    private let label: String
 
     public var logLevel: Logger.Level = .info
 
@@ -599,6 +600,7 @@ public struct StreamLogHandler: LogHandler {
 
     // internal for testing only
     internal init(label: String, stream: TextOutputStream) {
+        self.label = label
         self.stream = stream
     }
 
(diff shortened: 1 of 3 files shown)


weissi:
I'm cool with this :).

tomerd:

tomerd:
@prafsoni thanks for the contribution +1

do we have a test that covers the expected format?

weissi:
I realise I disagree with my former self who argued the label should not be in the message :)

weissi:
@tomerd I think this should be minor as it changes observable behaviour.

prafsoni:
@tomerd  After making changes I ran tests and all passed and looked to find a test where we are asserting the output pattern but didn't find any. Hence, we don't have a test case to verifies the `label` in the message.

If we add `var label: String { get }` to `LogHandler` protocol and make label public. Will this be useful? 
Let me know what you all think. :)

tomerd:
@prafsoni maybe a good opportunity to add a test that asserts the output. or is that an overkill? @weissi @ktoso wdyt?

prafsoni:

prafsoni:
@tomerd added a couple of tests

tomerd:

tomerd:
thanks @prafsoni looks like some formatting issues. please run `swiftformat .`

prafsoni:
done

tomerd:

PR #112, opened 2020-01-29
drop all caps

tomerd:
motivation: make docker setup more secured

changes: drop all caps


tomerd:

PR #113, opened 2020-02-03
Fix small typo in tests

FranzBusch:
### Motivation:

I just checked the tests and noticed that it should test `trace` and not `debug`.

### DIFF
--- a/Tests/LoggingTests/LoggingTest.swift
+++ b/Tests/LoggingTests/LoggingTest.swift
@@ -32,7 +32,7 @@ class LoggingTest: XCTestCase {
             XCTFail("trace should not be called")
             return "trace"
         }())
-        logger.debug({
+        logger.trace({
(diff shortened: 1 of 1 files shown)


Tests/LoggingTests/LoggingTest.swift
```diff
@@ -32,7 +32,7 @@ class LoggingTest: XCTestCase {
             XCTFail("trace should not be called")
```
FranzBusch:
```suggestion
            XCTFail("debug should not be called")
```

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -32,7 +32,7 @@ class LoggingTest: XCTestCase {
...
             return "trace"
```
FranzBusch:
```suggestion
            return "debug"
```

weissi:

PR #117, opened 2020-02-21
Update README to clarify the drawbacks of using os_log through swift-log using the backend implementation linked in the README

ravikandhadai:
The README file links to an implementation of a backend that uses os_log. However, the backend always uses `%{public}@` as the format string and eagerly converts all string interpolations to Strings.  This has two drawbacks: 1. the static components of the string interpolation would be eagerly copied by the unified logging system, which will result in loss of performance. 2. It also changes the default privacy scheme used by os_log and makes all messages public and doesn't allow controlling privacy. This PR updates the README to warn about these problems. It also links to the on-going effort on improving os_log Swift APIs and aligning it with swift-log.


ravikandhadai:
@weissi Thanks. Can you merge this PR?

PR #120, opened 2020-02-26
build: add a CMake based build for Windows

compnerd:
Given that swift-package-manager is not currently ready for Windows,
this addition makes it possible to build for Windows.

### Motivation:

### Modifications:

### Result:


weissi (changes requested):
@compnerd couldn't we generate the CMake files from SwiftPM? In SwiftNIO for example we generate the cocoapods using a script: https://github.com/apple/swift-nio/blob/master/scripts/build_podspecs.sh

If we just add the hard-coded CMake files, then we have two sources of truth. Then folks probably want to add Cocoapods files too and we'd have three. I think we should have one source of truth which is `Package.swift` and generate everything from there using a script. SwiftPM can output a stable `.json` description of the build.

compnerd:
@weissi - I don't think that package.swift has the information to really directly generate the CMake (its missing certain bits including list of sources, etc).  s-p-m doesn't run on Windows really, so, this would need to be committed ahead of time.

I think it may be possible to enhance CMake to generate Package.swift ....

Either way, in the mean time, I think that this would block the use of swift-log on Windows :-(

ktoso (approved):
LGTM in general, but somewhat nervous to merge a changes which have no (windows) CI validation 🤔

tomerd:
thanks for this @compnerd. I really like @weissi idea of generating these from the output of `swift package dump-package`, if can develop a script that does that, we could then have it added to the other swift-server projects until swift-pm gets windows support. wdyt?

compnerd:
I don’t think that the script would be trivial to write.  Since s-p-m doesn’t work on windows though, not sure how much I can help.   But, if you can write such a script, sure, having the output of it checked in would be fine (remember that s-p-m doesn’t run on Windows, so you cannot run the script on Windows, so the make would still need to be committed)

weissi:
@compnerd it just needs to work on Linux so the CI can verify that the checked in version is equal to the generated version. Happy to check it in (like the generated XCTest thingies) if we have checking that it's accurate.

ktoso:
Hey there, we're planning to tag 1.3.0 soon -- not sure about status of this PR, what should we do here?

tomerd:
I suggest we continue with release, and let this bake some more in the context of automating the generation of the CMake script or fixing SwiftPM to work on Windows

PR #121, opened 2020-02-27
update docs generation script to work better with selinux

tomerd:
motivation: docs scripts is broken on selinux due to restriction on ssh-agent

changes: change the jazzy workspace to a directory accessible by the CI node so we can push from the agent instead of from the docker container


tomerd:
@weissi once I bake the new solution on swift-log will need to carry over to other projects too

PR #122, opened 2020-02-28
additional doc generation fixes

tomerd:
motivation: better doc generation

changes:
* dont remove older docs when pushing new ones
* add version information and source links


tomerd:
follow up from #121

PR #123, opened 2020-03-13
test that demonstrates overloading

weissi:
Motivation:

SwiftLog is meant as a base logging library so users will often overload
certain functions with their own specialised and opinionated overloads.

Modification:

Add a test as demonstration.

Result:

More tests.


tomerd:
@weissi lgtm but generate linux tests

weissi:
@tomerd done

PR #133, opened 2020-06-01
silence #file to #filePath warnings

weissi:
Motivation:
    
    Swift 5.3 will keep `#file` == `#filePath` but starts warning if you
    pass `#file` to an argument that's `#filePath` defaulted (XCTest).
    
    Modification:
    
    - Replace `#file` with `(#file)` to silence the warning. In Swift 5.3
      `#file` == `#filePath`.
    
    Result:
    
    No warnings

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -11,3 +11,5 @@
 --ifdef no-indent
 
 # rules
+
+--disable redundantParens # https://github.com/nicklockwood/SwiftFormat/issues/638
--- a/Tests/LoggingTests/TestLogger.swift
+++ b/Tests/LoggingTests/TestLogger.swift
@@ -173,12 +173,20 @@ internal struct LogEntry {
 }
 
 extension History {
-    func assertExist(level: Logger.Level, message: String, metadata: Logger.Metadata? = nil, file: StaticString = #file, line: UInt = #line) {
+    func assertExist(level: Logger.Level,
+                     message: String,
+                     metadata: Logger.Metadata? = nil,
+                     file: StaticString = (#file),
+                     line: UInt = #line) {
         let entry = self.find(level: level, message: message, metadata: metadata)
         XCTAssertNotNil(entry, "entry not found: \(level), \(String(describing: metadata)), \(message) ", file: file, line: line)
     }
 
-    func assertNotExist(level: Logger.Level, message: String, metadata: Logger.Metadata? = nil, file: StaticString = #file, line: UInt = #line) {
+    func assertNotExist(level: Logger.Level,
+                        message: String,
+                        metadata: Logger.Metadata? = nil,
+                        file: StaticString = (#file),
+                        line: UInt = #line) {
         let entry = self.find(level: level, message: message, metadata: metadata)
         XCTAssertNil(entry, "entry was found: \(level), \(String(describing: metadata)), \(message)", file: file, line: line)
     }
--- a/Tests/LoggingTests/LoggingTest.swift
+++ b/Tests/LoggingTests/LoggingTest.swift
@@ -152,7 +152,7 @@ class LoggingTest: XCTestCase {
                                                    "lazy": .stringConvertible(LazyMetadataBox { "rendered-at-first-use" })])
     }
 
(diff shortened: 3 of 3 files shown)


Tests/LoggingTests/TestLogger.swift
```diff
@@ -292,3 +300,13 @@ internal struct TestLibrary {
         }
     }
 }
+
+#if compiler(>=5.3)
+internal func fullFilePath(_ filePath: StaticString = #filePath) -> StaticString {
```
weissi, line 305: this is here because `XCTest` wants a `StaticString` and not a `String`.

ktoso:
Formatting issues: https://ci.swiftserver.group/job/swift-log-swift50-prb/162/console (whitespace)

PR LGTM, ok to keep compatibility and no surprising changes I guess.
In the long run honestly path is what we'll want but we need to discuss and _get_ #module etc... Will make a separate ticket about it though.

weissi:
I'm deliberately not fixing the formatting right now because I think this is blocked on https://bugs.swift.org/browse/SR-12934

weissi:
@ktoso PR totally different now. Swift 5.3 actually keeps the behaviour, we just need to silence the  XCTest warnings for now.

.swiftformat
```diff
@@ -11,3 +11,5 @@
 --ifdef no-indent
 
 # rules
+
+--disable redundantParens # https://github.com/nicklockwood/SwiftFormat/issues/638
```
ktoso, line 15: 👍

Tests/LoggingTests/TestLogger.swift
```diff
@@ -292,3 +300,14 @@ internal struct TestLibrary {
...
```
ktoso, line 305: Why do we need this one in tests?
If #file == #filePath but just the warning around `= #file` sounds like there should be no difference on how we treat this in tests and the lib itself?

Sources/Logging/Logging.swift
```diff
@@ -235,7 +235,7 @@ extension Logger {
     ///    - message: The message to be logged. `message` can be used with any string interpolation literal.
     ///    - metadata: One-off metadata to attach to this log message
     ///    - file: The file this log message originates from (there's usually no need to pass it explicitly as it
-    ///            defaults to `#file`).
+    ///            defaults to `#file` (until Swift 5.2) / `#filePath` (from Swift 5.3)).
```
ktoso, line 238: Comment is semi true I guess... Would it be better to remove it? It's not like _we_ make that `if ...` but it's just Swift's behavior, so why should _we_ document that here if observed behavior is the same?

weissi: good catch, thank you!

ktoso:
Status update — I think this likely will be fine but I’ve ran out of time in the day to get it to completion in my project to really see it work out end to end.

It makes the label pretty useless but I guess we’ll agree that’s what it is. I’ll post a review tomorrow

weissi:
> Status update — I think this likely will be fine but I’ve ran out of time in the day to get it to completion in my project to really see it work out end to end.
> 
> It makes the label pretty useless but I guess we’ll agree that’s what it is. I’ll post a review tomorrow

I thought about `label` more and I actually now think it's useful. It's basically orthogonal to `source`.

Let's consider a whole system and let's assume the system offers 2 services, let's call them service A and service B. Services A & B both use a multitude of sub-subsystems. Then everything that belongs to Service A gets `label=MyApp.ServiceA` and B gets `label=MyApp.ServiceB`. If now both A & B use say AsyncHTTPClient, then the label suddenly becomes useful.

- `source=AHC` gives us AHC calls from both A & B which might be annoying if I'm a developer working only on A
- `source=AHC & label=MyApp.ServiceA` however gives me exactly what I want.

weissi:
@ktoso I think we're discussing this on the wrong issue actually :P

ktoso:
> @ktoso I think we're discussing this on the wrong issue actually :P

Whoops, you're right 😱 

Moving to https://github.com/apple/swift-log/pull/135

PR #137, opened 2020-06-18
forward #file to #filePath differently

weissi:
Motivation:

only works if done when _calling_ a function, not when defining one :|.

- https://github.com/apple/swift/pull/32445
- https://bugs.swift.org/browse/SR-12936
- https://bugs.swift.org/browse/SR-12934
- https://bugs.swift.org/browse/SR-13041

Modifications:

Silence #file to #filePath differently.

Result:

Hopefully at some point we get this working.

### DIFF
--- a/Tests/LoggingTests/LoggingTest.swift
+++ b/Tests/LoggingTests/LoggingTest.swift
@@ -152,8 +152,8 @@ class LoggingTest: XCTestCase {
                                                    "lazy": .stringConvertible(LazyMetadataBox { "rendered-at-first-use" })])
     }
 
-    private func dontEvaluateThisString(file: StaticString = (#file), line: UInt = #line) -> Logger.Message {
-        XCTFail("should not have been evaluated", file: file, line: line)
+    private func dontEvaluateThisString(file: StaticString = #file, line: UInt = #line) -> Logger.Message {
+        XCTFail("should not have been evaluated", file: (file), line: line)
         return "should not have been evaluated"
     }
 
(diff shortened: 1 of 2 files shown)


Tests/LoggingTests/LoggingTest.swift
```diff
@@ -152,8 +152,8 @@ class LoggingTest: XCTestCase {
                                                    "lazy": .stringConvertible(LazyMetadataBox { "rendered-at-first-use" })])
     }
 
-    private func dontEvaluateThisString(file: StaticString = (#file), line: UInt = #line) -> Logger.Message {
-        XCTFail("should not have been evaluated", file: file, line: line)
+    private func dontEvaluateThisString(file: StaticString = #file, line: UInt = #line) -> Logger.Message {
+        XCTFail("should not have been evaluated", file: (file), line: line)
```
ktoso, line 156: so the issue is that the previous impl would still cause warnings, and we have to silence the warning when using the functions, rather when defining them?

weissi: @ktoso no, worse, they would default to `LoggingTest.swift` instead of the caller's `#file`.

If you do `file: StaticString = #file`, then `#file` gets evaluated in the _caller_'s file. If you do `file: StaticString = (#file)`, then `(#file)` gets evaluated to the file where you type `(#file)`. It's totally broken...

ktoso: omg 😱

weissi: but there's hope: https://forums.swift.org/t/revisiting-the-source-compatibility-impact-of-se-0274-concise-magic-file-names/37720

ktoso: Yeah, keeping an eye on there... I didn't fully get that it's as broken as it is at first.

Issue #139, opened 2020-06-25
Issue setting different levels on log handlers in a MultiplexLogHandler


ohayon:
### Expected behavior
I expected that by setting different levels on the different handlers each one would be respected.

### Actual behavior
Looks like logs are being handled by both handlers only when the log function used is above the levels of *both* handlers.

### Steps to reproduce

In the following code snippet, `MyLogger` sets its own log level to `.info` in its `init` call.

```
    let stdoutLogger: (String) -> StreamLogHandler = { label in
      var logger = StreamLogHandler.standardOutput(label: label)
      logger.logLevel = .trace
      return logger
    }
    LoggingSystem.bootstrap { label in
      return MultiplexLogHandler([MyLogger(), stdoutLogger(label)])
    }

    log.trace("Test Trace")
    log.info("Test Info")
```

The output in the debugger is only:

```
2020-06-25T13:48:21-0400 info: Test Info
```

### SwiftLog version/commit hash
1.2.0

### Swift & OS version (output of `swift --version && uname -a`)
Swift 5.2 on iOS 13.5.1

ktoso:
Interesting catch, thanks for reporting!

So this is a weird interaction between how the `Logger` decides if it should dispatch to the underlying handler or not, it does so by:

```swift
        if self.logLevel <= level {
            self.handler.log(level: level, ...
```


and the multiplex handler naively implements logLevel as:

```swift
    public var logLevel: Logger.Level {
        get {
            return self.handlers[0].logLevel
```

so that's going to yield surprising behaviors as you just noticed.

I think we should be setting the loglevel to the "lowest of any passed in loggers" I suppose.

I'll give this a look since I have a moment now, thanks :-)

ohayon:
Thanks @ktoso ! I just read over the PR comments and that discussion is really great. It seems like a valid workaround for now would be to just reorder the handlers that are being passed into the multiplex handler?

ktoso:
Correct yes, until we land that PR and the behavior changes that's what is documented and the behavior -- so yeah, you can put the "lowest" log level one first and that should work :)

ohayon:
Thanks for the quick reply! 

I think maybe I have misunderstood some of that conversation on the PR, but now, by switching the order of the two loggers, `trace` messages are still getting sent to `MyLogger` which has its log level set to `info`. Is that still expected?

ohayon:
Ah, disregard the last message, sorry! I am seeing now in the code change that that will respect the multiple log levels 😃.

PR #140, opened 2020-06-27
=multiplex #139 more properly handle logLevel in multiplex handler

ktoso:
### Motivation:

Log level handling is a bit too naive in the multiplex handler when underlying loggers have varying levels.

### Modifications:

- effective log level has to be the lowest of all passed in handlers / same on level modification
- when dispatching, the multiplex handler has to filter again based on the _specific_ handler.

### Result:

- resolves https://github.com/apple/swift-log/issues/139

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -484,11 +484,11 @@ public struct MultiplexLogHandler: LogHandler {
             self.mutatingForEachHandler {
                 $0.logLevel = newValue
             }
-            self.effectiveLogLevel = MultiplexLogHandler.effectiveLogLevel(self.handlers.map { $0.logLevel })
+            self.effectiveLogLevel = newValue
         }
     }
 
-    static func effectiveLogLevel(_ levels: [Logger.Level]) -> Logger.Level {
+    private static func effectiveLogLevel(_ levels: [Logger.Level]) -> Logger.Level {
         return levels.min() ?? .trace
     }
 
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -471,14 +471,14 @@ extension Logger {
 /// and thus it would not actually log that log message).
 ///
 /// If the log level is _set_ on a `Logger` backed by an `MultiplexLogHandler` the log level will apply to *all*
-/// underlying log handlers. Allowing a logger to still select at what level it wants to log, regardless if the underlying
+/// underlying log handlers, allowing a logger to still select at what level it wants to log regardless of if the underlying
 /// handler is a multiplex or a normal one. If for some reason one might want to not allow changing a log level of a specific
-/// handled passed into the multiplex log handler, this is possible by wrapping it in a handler which ignores any log level changes.
+/// handler passed into the multiplex log handler, this is possible by wrapping it in a handler which ignores any log level changes.
 ///
 /// ### Effective Logger.Metadata
 ///
-/// Since a `MultiplexLogHandler` is a combination of multiple log handlers, the handling of metadata can be non-obvious
-/// at first when the underlying log handlers have some metadata of their own set, before they are used to initialize the handler.
+/// Since a `MultiplexLogHandler` is a combination of multiple log handlers, the handling of metadata can be non-obvious.
+/// For example, the underlying log handlers may have metadata of their own set before they are used to initialize the multiplex log handler.
 ///
 /// The multiplex log handler acts purely as proxy and does not make any changes to underlying handler metadata other than
 /// proxying writes that users made on a `Logger` instance backed by this handler.
@@ -500,7 +500,7 @@ extension Logger {
 /// A query through the multiplex log handler the key `one` naturally returns `handler1`'s value, and a query for `two`
 /// naturally returns `handler2`'s value. Querying for the key `all` will return `handler1`'s value, as that handler was indicated
 /// "more important" than the second handler. The same rule applies when querying for the `metadata` property of the
-/// multiplex log handler - it constructs `Metadata` uniquing values
+/// multiplex log handler - it constructs `Metadata` uniquing values.
 public struct MultiplexLogHandler: LogHandler {
     private var handlers: [LogHandler]
     private var effectiveLogLevel: Logger.Level
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -458,12 +458,52 @@ extension Logger {
 
 /// A pseudo-`LogHandler` that can be used to send messages to multiple other `LogHandler`s.
 ///
-/// The first `LogHandler` passed to the initialisation function of `MultiplexLogHandler` control the `logLevel` as
-/// well as the `metadata` for this `LogHandler`. Any subsequent `LogHandler`s used to initialise a
-/// `MultiplexLogHandler` are merely to emit the log message to another place.
+/// ### Effective Logger.Level
+///
+/// When first initialized the multiplex log handlers' log level is automatically set to the minimum of all the
+/// passed in log handlers. This ensures that each of the handlers will be able to log at their appropriate level
+/// any log events they might be interested in.
+///
+/// Example:
+/// If log handler `A` is logging at `.debug` level, and log handler `B` is logging at `.info` level, the constructed
+/// `MultiplexLogHandler([A, B])`'s effective log level will be set to `.debug`, meaning that debug messages will be
+/// handled by this handler, while only logged by the underlying `A` log handler (since `B`'s log level is `.info`
+/// and thus it would not actually log that log message).
+///
+/// If the log level is _set_ on a `Logger` backed by an `MultiplexLogHandler` the log level will apply to *all*
+/// underlying log handlers. Allowing a logger to still select at what level it wants to log, regardless if the underlying
+/// handler is a multiplex or a normal one. If for some reason one might want to not allow changing a log level of a specific
+/// handled passed into the multiplex log handler, this is possible by wrapping it in a handler which ignores any log level changes.
+///
+/// ### Effective Logger.Metadata
+///
+/// Since a `MultiplexLogHandler` is a combination of multiple log handlers, the handling of metadata can be non-obvious
+/// at first when the underlying log handlers have some metadata of their own set, before they are used to initialize the handler.
+///
+/// The multiplex log handler acts purely as proxy and does not make any changes to underlying handler metadata other than
+/// proxying writes that users made on a `Logger` instance backed by this handler.
+///
+/// Setting metadata is always proxied through to _all_ underlying handlers, meaning that if a modification like
+/// `logger[metadataKey: "x"] = "y"` is made, all underlying log handlers that this multiplex handler was initiated with
+/// will observe this change.
+///
+/// Reading metadata from the multiplex log handler MAY need to pick one of conflicting values if the underlying log handlers
+/// were already initiated with some metadata before passing them into the multiplex handler. The multiplex handler uses
+/// the order in which the handlers were passed in during its initialization as a priority indicator - the first handler's
+/// values are more important than the next handlers values, etc.
+///
+/// Example:
+/// If the multiplex log handler was initiated with two handlers like this: `MultiplexLogHandler([handler1, handler2])`.
+/// The handlers each have some already set metadata: `handler1` has metadata values for keys `one` and `all`, and `handler2`
+/// has values for keys `two` and `all`.
+///
+/// A query through the multiplex log handler the key `one` naturally returns `handler1`'s value, and a query for `two`
+/// naturally returns `handler2`'s value. Querying for the key `all` will return `handler1`'s value, as that handler was indicated
+/// "more important" than the second handler. The same rule applies when querying for the `metadata` property of the
+/// multiplex log handler - it constructs `Metadata` uniquing values
 public struct MultiplexLogHandler: LogHandler {
-    private var effectiveLogLevel: Logger.Level
     private var handlers: [LogHandler]
+    private var effectiveLogLevel: Logger.Level
 
     /// Create a `MultiplexLogHandler`.
     ///
@@ -471,9 +511,9 @@ public struct MultiplexLogHandler: LogHandler {
     ///    - handlers: An array of `LogHandler`s, each of which will receive the log messages sent to this `Logger`.
     ///                The array must not be empty.
     public init(_ handlers: [LogHandler]) {
-        assert(!handlers.isEmpty)
+        assert(!handlers.isEmpty, "MultiplexLogHandler.handlers MUST NOT be empty")
         self.handlers = handlers
-        self.effectiveLogLevel = MultiplexLogHandler.effectiveLogLevel(handlers.map { $0.logLevel })
+        self.effectiveLogLevel = handlers.map { $0.logLevel }.min() ?? .trace
     }
 
     public var logLevel: Logger.Level {
@@ -481,15 +521,9 @@ public struct MultiplexLogHandler: LogHandler {
             return self.effectiveLogLevel
         }
         set {
-            self.mutatingForEachHandler {
-                $0.logLevel = newValue
-            }
+            self.mutatingForEachHandler { $0.logLevel = newValue }
             self.effectiveLogLevel = newValue
         }
-    }
-
-    private static func effectiveLogLevel(_ levels: [Logger.Level]) -> Logger.Level {
-        return levels.min() ?? .trace
     }
 
     public func log(level: Logger.Level,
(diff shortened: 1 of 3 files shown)


Tests/LoggingTests/LoggingTest.swift
```diff
@@ -74,6 +74,73 @@ class LoggingTest: XCTestCase {
...
+                return [:]
+            }()
+        )
+
+        logging1.history.assertExist(level: .info, message: "info")
+        logging2.history.assertExist(level: .info, message: "info")
+
+        XCTAssertEqual(messageMaterializations, 1)
+        XCTAssertEqual(metadataMaterializations, 1)
+    }
```
ktoso, line 142: this is kind of enforced "by construction" but wanted to have the test in anyway

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -74,6 +74,73 @@ class LoggingTest: XCTestCase {
...
+        logging1.history.assertNotExist(level: .trace, message: "trace")
+        logging1.history.assertNotExist(level: .debug, message: "debug")
+        logging1.history.assertExist(level: .info, message: "info")
+        logging1.history.assertExist(level: .warning, message: "warning")
+
+        logging2.history.assertNotExist(level: .trace, message: "trace")
+        logging2.history.assertExist(level: .debug, message: "debug")
+        logging2.history.assertExist(level: .info, message: "info")
+        logging2.history.assertExist(level: .warning, message: "warning")
+    }
```
ktoso, line 106: this test explains the feature

Sources/Logging/Logging.swift
```diff
@@ -472,24 +473,30 @@ public struct MultiplexLogHandler: LogHandler {
...
     public var logLevel: Logger.Level {
         get {
-            return self.handlers[0].logLevel
+            return self.effectiveLogLevel
         }
         set {
             self.mutatingForEachHandler {
                 $0.logLevel = newValue
             }
+            self.effectiveLogLevel = MultiplexLogHandler.effectiveLogLevel(self.handlers.map { $0.logLevel })
```
tomerd, line 487: since we are setting all the handlers in the move loop, computing effectiveLogLevel seems redundant

Sources/Logging/Logging.swift
```diff
@@ -472,24 +473,30 @@ public struct MultiplexLogHandler: LogHandler {
     public init(_ handlers: [LogHandler]) {
         assert(!handlers.isEmpty)
         self.handlers = handlers
+        self.effectiveLogLevel = MultiplexLogHandler.effectiveLogLevel(handlers.map { $0.logLevel })
```
tomerd, line 476: +1

Sources/Logging/Logging.swift
```diff
@@ -472,24 +473,30 @@ public struct MultiplexLogHandler: LogHandler {
...
         }
         set {
             self.mutatingForEachHandler {
                 $0.logLevel = newValue
             }
+            self.effectiveLogLevel = MultiplexLogHandler.effectiveLogLevel(self.handlers.map { $0.logLevel })
         }
     }
 
+    static func effectiveLogLevel(_ levels: [Logger.Level]) -> Logger.Level {
```
tomerd, line 491: private?

Sources/Logging/Logging.swift
```diff
@@ -472,24 +473,30 @@ public struct MultiplexLogHandler: LogHandler {
...
+        return levels.min() ?? .trace
+    }
+
     public func log(level: Logger.Level,
                     message: Logger.Message,
                     metadata: Logger.Metadata?,
                     file: String, function: String, line: UInt) {
-        self.handlers.forEach { handler in
+        for handler in self.handlers where handler.logLevel <= level {
```
tomerd, line 499: wouldn't delegating to the underlying `handler.log` already filter by level? what am I missing?

Sources/Logging/Logging.swift
ktoso: ok will do

Sources/Logging/Logging.swift
ktoso: that was "problem", example:

```
logger1 at .info
logger2 at .debug

multiplex.debug() // would never be emitted
```

because:

`multiplex.loglevel` was `return self.handlers[0].logLevel` arbitrary using the "first" as the handlers log level.

How logger decides if it should pass value to handler is: 

```
if handler.logLevel <= level {
```

so even though the `logger2` _does_ want to log things on debug level, it would not, because `logger1` is set to info; That's a bit weird.

----

Solutions:

- (this PR) on construction find "lowest loglevel this multiplex handler should accept, because one of its parts wants to get it"
- OR: make it like a normal logger/handler that it defaults to `.info` and if we want to log at other levels we need to change that on the multiplex handler?
- OR: (how it works today) the "first logger" is special and decides at which log level we're logging.
  - note that `multiplex.logLevel = .trace` then is a bit weird -- we've just changed _all_ loggers to .trace even though before we were saying the "first one only matters"; so the current impl is confusing since without setting a level: handlers log on VARIOUS levels; with setting a level, ALL handlers log at the same level.
  - rather what we'd want to do IMO when setting `multiplex.logLevel = .trace` is that "this handler will now take trace logs; and if the log handlers it contains WANT TO (were passed in as such) to log at .trace they now have the chance to do so" (this PR) -- so we retain the log level configurations that were passed in...
  - perhaps we'd then need a way to _really set_ the underlying levels if we wanted to? not sure.

ktoso: Oh wait, sorry didn't have my coffee yet! You mean that the looping here is redundant, thanks yes that's true.

☕ ☕ ☕

Sources/Logging/Logging.swift
ktoso: See the analysis here: https://github.com/apple/swift-log/pull/140/files#r447320481

ktoso: > the underlying handler.log already filter by level? what am I missing?

No, a handler is not normally performing the log level check -- only the Logger.log() API does that:

```
    @inlinable
        if self.logLevel <= level {
            self.handler.log(level: level,
                             message: message(),
                             metadata: metadata(),
                             file: file, function: function, line: line)
        }
    }
```

and handlers are naive and "if i am called, it seems I should log." This is why we have to make the `multiplex.logLevel` smarter as we do in this PR:

- so the change to logLevel in this PR causes the `Logger.log()` to dispatch to this multiplex handler when _at least one of the log handlers would like to log this message_
- but since now we don't know which logger that was... we have to check.

There's an open question how `multiplex.logLevel = .info` relates to this all, since then all levels are made equal and all will log. But what about if we only do:

```
logger1 at debug
logger2 at info

multiplex.debug() // we only want it to be logged at debug at logger1, not at logger2 as well
```

ktoso: For context, `StreamLogHandler`:

```swift
                    message: Logger.Message,
                    metadata: Logger.Metadata?,
                    file: String, function: String, line: UInt) {
        let prettyMetadata = metadata?.isEmpty ?? true
            ? self.prettyMetadata
            : self.prettify(self.metadata.merging(metadata!, uniquingKeysWith: { _, new in new }))
```

tomerd: ah you are correct, I forgot the filter was at the Logger level not backend one. this is 100%

ktoso:
I now realized the current behavior is documented:

```
/// The first `LogHandler` passed to the initialisation function of `MultiplexLogHandler` control the `logLevel` as
/// well as the `metadata` for this `LogHandler`. Any subsequent `LogHandler`s used to initialise a
/// `MultiplexLogHandler` are merely to emit the log message to another place.
```

but it's admittably a bit weird... not sure what to do here tbh.

The current behavior has some upsides (detailed control) but it is somewhat unexpected as witnessed by #139 as well as my going down this rabbit hole.

Do we have opinions what do do here @glbrntt @Lukasa (optionally, only if you have time and want to chime in @weissi) @tomerd ?

tomerd:
> Do we have opinions what do do here @glbrntt @Lukasa (optionally, only if you have time and want to chime in @weissi) @tomerd ?

for the log level, I think the semantics proposed by this PR is an improvement, but it does make the metadata story more murky since its still pinned to handler[0]

Lukasa:
Yeah, I think this behavioural change needs to happen. The logical semantics of MultiplexLogHandler, in my mind, would be that its effective log level is the lowest of all log handlers it wraps, _and_ that it only dispatches the messages to log handlers whose actual log level is lower than that of the message. This makes it maximally useful.

ktoso:
>  The logical semantics of MultiplexLogHandler, in my mind, would be that its effective log level is the lowest of all log handlers it wraps, and that it only dispatches the messages to log handlers whose actual log level is lower than that of the message.

Cool, very much agreed, that's what the current PR brings it to 👍 

I'll sanity check the metadata handling again here soon; I guess the semantics could be:

- do not just expose any of the underlying ones
- when set, set onto all underlying ones

or something like that; rather than the [0] based one

glbrntt:
I agree with what's been said above: the current behaviour is weird; this change introduces the behaviour I would have expected.

Sources/Logging/Logging.swift
```diff
@@ -458,40 +458,74 @@ extension Logger {
...
+/// Example:
+/// If log handler `A` is logging at `.debug` level, and log handler `B` is logging at `.info` level, the constructed
+/// `MultiplexLogHandler([A, B])`'s effective log level will be set to `.debug`, meaning that debug messages will be
+/// handled by this handler, while only logged by the underlying `A` log handler (since `B`'s log level is `.info`
+/// and thus it would not actually log that log message).
+///
+/// If the log level is _set_ on a `Logger` backed by an `MultiplexLogHandler` the log level will apply to *all*
+/// underlying log handlers. Allowing a logger to still select at what level it wants to log, regardless if the underlying
+/// handler is a multiplex or a normal one. If for some reason one might want to not allow changing a log level of a specific
+/// handled passed into the multiplex log handler, this is possible by wrapping it in a handler which ignores any log level changes.
```
ktoso, line 476: docs and examples on how the log level works

Sources/Logging/Logging.swift
```diff
@@ -458,40 +458,74 @@ extension Logger {
...
+///
+/// Example:
+/// If the multiplex log handler was initiated with two handlers like this: `MultiplexLogHandler([handler1, handler2])`.
+/// The handlers each have some already set metadata: `handler1` has metadata values for keys `one` and `all`, and `handler2`
+/// has values for keys `two` and `all`.
+///
+/// A query through the multiplex log handler the key `one` naturally returns `handler1`'s value, and a query for `two`
+/// naturally returns `handler2`'s value. Querying for the key `all` will return `handler1`'s value, as that handler was indicated
+/// "more important" than the second handler. The same rule applies when querying for the `metadata` property of the
+/// multiplex log handler - it constructs `Metadata` uniquing values
```
ktoso, line 503: docs and examples how metadata works

ktoso: I actually had a "smart and configurable" `EffectiveMetadataMode` based impl but I stepped back from it for a few hours to re-think and it was getting too complex IMO for what it is -- realistically this is not a "big problem that needs fancy solving" -- normally users just care about setting metadata on things and maybe some handlers have something specific to them, the conflicting cases should be rare and this simple rule is better than slamming more complexity into it.

// I even had an impl which would "collect all values if the same key is present in a few handlers into a logger metadata value Array etc... Way too fancy for what we're looking at I then realized.

Sources/Logging/Logging.swift
```diff
@@ -458,40 +458,74 @@ extension Logger {
...
+    private var effectiveLogLevel: Logger.Level
 
     /// Create a `MultiplexLogHandler`.
     ///
     /// - parameters:
     ///    - handlers: An array of `LogHandler`s, each of which will receive the log messages sent to this `Logger`.
     ///                The array must not be empty.
     public init(_ handlers: [LogHandler]) {
-        assert(!handlers.isEmpty)
+        assert(!handlers.isEmpty, "MultiplexLogHandler.handlers MUST NOT be empty")
```
ktoso, line 514: explicit messages always better IMO

Sources/Logging/Logging.swift
```diff
@@ -458,40 +458,74 @@ extension Logger {
...
         self.handlers = handlers
-        self.effectiveLogLevel = MultiplexLogHandler.effectiveLogLevel(handlers.map { $0.logLevel })
+        self.effectiveLogLevel = handlers.map({ $0.logLevel }).min() ?? .trace
```
ktoso, line 516: the ?? will never trigger but it's semantically "aha makes sense, we take the lowest" so I prefer that to writing `!` here

Sources/Logging/Logging.swift
```diff
@@ -503,16 +537,27 @@ public struct MultiplexLogHandler: LogHandler {
...
     public var metadata: Logger.Metadata {
         get {
-            return self.handlers[0].metadata
+            var effectiveMetadata: Logger.Metadata = [:]
+            // as a rough estimate we assume that the underlying handlers have a similar metadata count,
+            // and we use the first one's current count to estimate how big of a dictionary we need to allocate:
+            effectiveMetadata.reserveCapacity(self.handlers.first!.metadata.count) // !-safe, we always have at least one handler
+            return self.handlers.reduce(into: effectiveMetadata) { effectiveMetadata, handler in
+                effectiveMetadata.merge(handler.metadata, uniquingKeysWith: { l, _ in l })
+            }
```
ktoso, line 546: We pick the "left" over any following ones "first handlers in the list are more important"

Sources/Logging/Logging.swift
```diff
@@ -503,16 +537,27 @@ public struct MultiplexLogHandler: LogHandler {
...
-    public subscript(metadataKey metadataKey: String) -> Logger.Metadata.Value? {
+    public subscript(metadataKey metadataKey: Logger.Metadata.Key) -> Logger.Metadata.Value? {
         get {
-            return self.handlers[0].metadata[metadataKey]
+            for handler in handlers {
+                if let value = handler[metadataKey: metadataKey] {
+                    return value
+                }
+            }
+            return nil
```
ktoso, line 560: Again, implements the "first is most important"

ktoso:
I poked at the metadata quite a bit over the weekend and think I've arrived at what is a good tradeoff here.

The metadata handling change is: cdb48692e8bd7aab029651336733da70fde17e6d

And please see the comments in-line. Short version though: we should not magically rely on the _first_ handler, but we should _prefer_ more-first passed in handler metadata values over latter ones. This is simple to explain and implement without some crazy things (I had some advanced implementations with merge strategies and whatnot, but they ended up feeling wrong in their own ways, so I believe a simple rule like this (see added docs) is a good tradeoff, esp since we should not expect to be hitting this in reality much -- it is the underlying handlers doing their job, and barely anyone should be reading from the logger metadata to "just have a look at it" after all).

Sources/Logging/Logging.swift
```diff
@@ -458,54 +458,106 @@ extension Logger {
...
+/// Example:
+/// If log handler `A` is logging at `.debug` level, and log handler `B` is logging at `.info` level, the constructed
+/// `MultiplexLogHandler([A, B])`'s effective log level will be set to `.debug`, meaning that debug messages will be
+/// handled by this handler, while only logged by the underlying `A` log handler (since `B`'s log level is `.info`
+/// and thus it would not actually log that log message).
+///
+/// If the log level is _set_ on a `Logger` backed by an `MultiplexLogHandler` the log level will apply to *all*
+/// underlying log handlers. Allowing a logger to still select at what level it wants to log, regardless if the underlying
+/// handler is a multiplex or a normal one. If for some reason one might want to not allow changing a log level of a specific
+/// handled passed into the multiplex log handler, this is possible by wrapping it in a handler which ignores any log level changes.
```
glbrntt, line 476:
```suggestion
/// handler passed into the multiplex log handler, this is possible by wrapping it in a handler which ignores any log level changes.
```

Sources/Logging/Logging.swift
```diff
@@ -458,54 +458,106 @@ extension Logger {
...
+///
+/// ### Effective Logger.Metadata
+///
+/// Since a `MultiplexLogHandler` is a combination of multiple log handlers, the handling of metadata can be non-obvious
+/// at first when the underlying log handlers have some metadata of their own set, before they are used to initialize the handler.
```
glbrntt, line 481: I found this paragraph hard to parse:

```suggestion
/// Since a `MultiplexLogHandler` is a combination of multiple log handlers, the handling of metadata can be non-obvious.
/// For example, the underlying log handlers may have metadata of their own set before they are used to initialize the multiplex log handler.
```

Sources/Logging/Logging.swift
```diff
@@ -458,54 +458,106 @@ extension Logger {
...
+/// any log events they might be interested in.
+///
+/// Example:
+/// If log handler `A` is logging at `.debug` level, and log handler `B` is logging at `.info` level, the constructed
+/// `MultiplexLogHandler([A, B])`'s effective log level will be set to `.debug`, meaning that debug messages will be
+/// handled by this handler, while only logged by the underlying `A` log handler (since `B`'s log level is `.info`
+/// and thus it would not actually log that log message).
+///
+/// If the log level is _set_ on a `Logger` backed by an `MultiplexLogHandler` the log level will apply to *all*
+/// underlying log handlers. Allowing a logger to still select at what level it wants to log, regardless if the underlying
```
Lukasa, line 474: Nit, sentence fragment. Consider revising to:

```suggestion
/// underlying log handlers, allowing a logger to still select at what level it wants to log regardless of if the underlying
```

Sources/Logging/Logging.swift
```diff
@@ -458,54 +458,106 @@ extension Logger {
...
+///
+/// Example:
+/// If the multiplex log handler was initiated with two handlers like this: `MultiplexLogHandler([handler1, handler2])`.
+/// The handlers each have some already set metadata: `handler1` has metadata values for keys `one` and `all`, and `handler2`
+/// has values for keys `two` and `all`.
+///
+/// A query through the multiplex log handler the key `one` naturally returns `handler1`'s value, and a query for `two`
+/// naturally returns `handler2`'s value. Querying for the key `all` will return `handler1`'s value, as that handler was indicated
+/// "more important" than the second handler. The same rule applies when querying for the `metadata` property of the
+/// multiplex log handler - it constructs `Metadata` uniquing values
```
glbrntt, line 503:
```suggestion
/// multiplex log handler - it constructs `Metadata` uniquing values.
```

Sources/Logging/Logging.swift
ktoso: Thanks, I always over do it with long sentences 😅

ktoso:
Thanks for the review and cleanups folks 👍

PR #142, opened 2020-07-01
Add NoOpLogHandler for when you don't want logging

adam-fowler:

Provider a LogHandler that outputs nothing

I've just started looking into adding logging to aws-sdk-swift and I've found I am going to be replicating the `NoOpLogHandler` from async-http-client in aws-sdk-swift-core. It's a small bit of code but I think worthwhile to bring into swift-log to avoid multiple systems replicating this functionality when they want to provide a no logging option.  

Add struct `NoOpLogHandler` which conforms to `LogHandler` and implements all the methods as empty functions.

Systems will be able to provide a no logging option, without having to test against an optional `Logger`.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -646,23 +646,23 @@ public struct StreamLogHandler: LogHandler {
 
 /// No operation LogHandler, used when no logging is required
 public struct NoOpLogHandler: LogHandler {
-    public func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, file: String, function: String, line: UInt) {}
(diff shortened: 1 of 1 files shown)


ktoso:

Lukasa:
I think I'm ok adding a new name in a minor if we mangle it, e.g. `SwiftLogNoOpLogHandler`.

adam-fowler:
> I think I'm ok adding a new name in a minor if we mangle it, e.g. `SwiftLogNoOpLogHandler`.

Happy to go with that

ktoso:
That sounds like the safe bet to take here, thanks both :-)

PR #144, opened 2020-07-06
=doc #82 avoid showcasing internal lock in public docs

### LINKED ISSUE #82: Lock is internal
I'd like to use `Lock` in a custom LogHandler implementation but am unable to because `Lock` is declared as `internal` in `Locks.swift`. Was `Lock` intended to be used like this?

### Expected behavior

Under "Special cases" in the API docs [here](https://apple.github.io/swift-log/docs/current/Logging/Protocols/LogHandler.html), the example implementation `LogHandlerWithGlobalLogLevelOverride` has a property `overrideLock = Lock()`.

### Actual behavior

This is not possible if the custom LogHandler implementation (e.g. `LogHandlerWithGlobalLogLevelOverride`) is defined in another module as `Lock` is internal and not visible there.

`version: 0.0.0`

`Apple Swift version 4.2.1 (swiftlang-1000.11.42 clang-1000.11.45.1)
Target: x86_64-apple-darwin18.6.0`

ktoso:
### Motivation:

Example in public type was using our internal Lock() so it was not copy paste friendly.

### Modifications:

Use foundation's lock in the example.

### Result:

- resolves #82 
- copy paste friendly example in docs

### DIFF
--- a/Sources/Logging/LogHandler.swift
+++ b/Sources/Logging/LogHandler.swift
@@ -56,9 +56,11 @@
 /// `LogHandlerWithGlobalLogLevelOverride.overrideGlobalLogLevel = .debug`, for example.
 ///
 /// ```swift
+/// import class Foundation.NSLock
+///
 /// public struct LogHandlerWithGlobalLogLevelOverride: LogHandler {
 ///     // the static properties hold the globally overridden log level (if overridden)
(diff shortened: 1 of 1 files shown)


ktoso:
Thanks @tomerd :)

PR #146, opened 2020-07-07
Make SwiftLogNoOpLogHandler.init() public

adam-fowler:
I want to cry

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -733,6 +733,8 @@ public struct StreamLogHandler: LogHandler {
 
 /// No operation LogHandler, used when no logging is required
 public struct SwiftLogNoOpLogHandler: LogHandler {
+    public init() {}
+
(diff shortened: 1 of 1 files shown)


ktoso:
Oh my that's silly 😢 😂

ktoso:
We can make a 1.3.1 soon, I'd count that as fixup rather than new API after all.

Thanks for the PR @adam-fowler, that was so easy to miss... I didn't think we needed a test for a noop heh but that would have helped spot it I guess -- lesson learnt 🤷‍♂️

adam-fowler:
@ktoso cheers thanks

ktoso:
Heads up notice, we'll cut a 1.3.1 tomorrow - since there's not much reason not to

ktoso:
Meh, let's call it 1.4.0 since technically we're "adding API".

ktoso:
And it's out: https://github.com/apple/swift-log/releases/tag/1.4.0

PR #151, opened 2020-08-13
Sort metadata by keys when prettifying

### LINKED ISSUE #128: Metadata key/value pairs don't maintain order between runs using StreamLogHandler
When adding metadata to loggers and using `StreamLogHandler`, the order of key/value pairs are not maintained. Should this be the case or is it intentional?

glbrntt:
Motivation:

The stream log handler doesn't print metadata in a stable order. While
swift-log is an API package and shouldn't provide all the bells and
whistles for formatting logs, it should make the default log handler
at least somewhat sensible.

Modifications:

- Sort the metadata by key when 'prettifying' it

Result:

- The stream log handler emits logs with metadata in a stable order
- Resolves #128

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -715,7 +715,9 @@ public struct StreamLogHandler: LogHandler {
     }
 
     private func prettify(_ metadata: Logger.Metadata) -> String? {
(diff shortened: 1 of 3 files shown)


ktoso:

ktoso:
No reason to wait further, thanks a lot @glbrntt !

ktoso:
Whoops oh no, this was still against master and it seems I forgot to re-target it... will submit against `main`

PR #153, opened 2020-08-17
Make it possible to configure `LogHandler`s when creating `Logger`s

gwynne:
It is now possible to supply additional configuration to a `Logger`'s `LogHandler` without having to specify an entire handler factory.

### Motivation:

It is sometimes desirable to configure a new `Logger`'s `LogHandler` specially for the particular `Logger` (such as by changing its handler-specific metadata). Currently, the only way to accomplish this is to provide a new handler factory which creates a new `LogHandler`, with no reference to the default handler type that was used to bootstrap the logging system. This means that library code wishing to provide specialized logging for itself must currently hardcode use of, e.g., `StreamLogHandler` instead of configuring whichever handler the app using the library bootstrapped.

### Modifications:

A new initializer for `Logger` has been added which accepts a configuration closure; this closure is provided the `LogHandler` which will be used by the new `Logger`, and may modify it as needed.

### Result:

Code such as this:

```swift
// calling app has configured, e.g., a typical `MultiplexLogHandler`
let logger = Logger(label: "something-something-dark-side", factory: { label in
    // oh well, forced to use the wrong handler, hope the app doesn't mind
    var handler = HypotheticalLogHandler.standardOutput(label: label)
    handler.hoardsKrendars = true
    return handler
})
```

can now instead be written in this fashion:

```swift
// calling app might have configured a multiplex handler, or it might not have
let logger = Logger(label: "something-something-dark-side", configuringWith: { handler in
    guard handler is HypotheticalLogHandler else { return }
    (handler as! HypotheticalLogHandler).hoardsKrendars = true // can't assign through `as?`
})
```

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -430,7 +430,7 @@ extension Logger {
     public init(label: String, factory: (String) -> LogHandler) {
         self = Logger(label: label, factory(label))
     }
-    
+
     /// Construct a `Logger` given a `label` identifying the creator of the `Logger` and configuring the instantiated
     /// log handler before using it.
     ///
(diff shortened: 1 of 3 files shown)


ktoso:
We caught up on chat and the initial use case was actually changing log level which can be done but was not noticed by OP :) 

We're thinking thought if a general "configure my handler" capability can be useful...

PR #154, opened 2020-08-21
Make Logger.Metadata.Key strongly typed

ffried:
Make `Logger.Metadata.Key` strongly typed by using a separate struct.

### Motivation:

Using `String` as keys makes it hard to use constants for metadata keys. They have to be defined in some arbitrary namespace. Or if they don't, they're messy to change and even harder to "look up" (e.g. what keys exist on that logger? what key stores what information) since there is no central definition place.

```swift
enum SomeMetadataKeyNamespace {
    static let myMetadataKey = "awesome-key"
}

var logger = Logger(label: "some.great.logger")
logger[metadataKey: SomeMetadataKeyNamespace. myMetadataKey] = "what a beautiful value"

// Later on in e.g. a different file (a `LogHandler` implementation for example)
if case .string(let str)? = logger[metadataKey: SomeMetadataKeyNamespace. myMetadataKey], str.contains("beautiful") {
    print("It's beautiful!")
} else {
    print("Not beautiful...")
}
```

### Modifications:

- Add `Logger.MetadataKey` and change the `typealias` for `Logger.Metadata` to use `Logger.MetadataKey` as key instead of String.
- Change the protocol requirement in `LogHandler` to take `Logger.Metadata.Key` instead of `String` in `subscript(metadataKey:)`.
- `Logger.MetadataKey` conforms to:
    - `RawRepresentable` (`RawValue = String`)
    - `Hashable`
    - `CustomStringConvertible` (`description` being `rawValue`)
    - `ExpressibleByStringLiteral` (`StringLiteralType = String`)
    - `ExpressibleByStringInterpolation` (`StringInterpolation` = `DefaultStringInterpolation`)

### Result:

Using a separate struct for `Logger.Metadata.Key` allows static constants on it making metadata "nicer" to use:

```swift
extension Logger.Metadata.Key {
     static let myMetadataKey = Logger.Metadata.Key(rawValue: "awesome-key")
}

var logger = Logger(label: "some.great.logger")
logger[metadataKey: .myMetadataKey] = "what a beautiful value"

// Later on in e.g. a different file (a `LogHandler` implementation for example)
if case .string(let str)? = logger[metadataKey: .myMetadataKey], str.contains("beautiful") {
    print("It's beautiful!")
} else {
    print("Not beautiful...")
}
```

###  Source Compatibility
I'm not sure this change can be made in a source compatible way - especially since the `subscript(metadataKey:)` was directly requiring `String` instead of `Logger.Metadata.Key`. Most implementations would probably be fine with the deprecated overload I added (similar than what has been done for `source` - although I've read that this wasn't perfect either). But those implementations that use `String` APIs will definitively break. Thus this will likely need a new major version - in which case I could remove the "attempted backwards compatibility" 🙂

### DIFF
--- a/Sources/Logging/LogHandler.swift
+++ b/Sources/Logging/LogHandler.swift
@@ -146,6 +146,11 @@ public protocol LogHandler {
     ///
     /// - parameters:
     ///    - metadataKey: The key for the metadata item
+    subscript(metadataKey _: Logger.Metadata.Key) -> Logger.Metadata.Value? { get set }
+
+    /// SwiftLog 1.4 compatibility method. Please do _not_ implement, implement
+    /// the subscript that uses `Logger.Metadata.Key` instead.
+    @available(*, deprecated, message: "Please use `Logger.Metadata.Key` instead of String-based keys")
     subscript(metadataKey _: String) -> Logger.Metadata.Value? { get set }
 
     /// Get or set the entire metadata storage as a dictionary.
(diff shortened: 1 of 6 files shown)


ktoso:
I'm going to close this as we're working on proposing baggage (from tracing) integration into swift-log, we'll share a proposal soon.

ffried:
@ktoso I'm sorry that I haven't found time to take another look at this. And thanks for taking this on!

ktoso:
No worries at all, we'll get to a comprehensive solution soon 👍

PR #167, opened 2020-10-22
Logging: repair the Windows build after MSVCRT rename

compnerd:
Import `CRT` instead of `MSVCRT`.

### Motivation:

### Modifications:

### Result:


ktoso:

ktoso (approved):
We don't have a windows CI but if you say that's correct I'll believe you @compnerd 👍

drexin (approved):

PR #169, opened 2020-11-12
Add sushichop/Puppy to list of backends

sushichop:
Updates the list of SwiftLog backend implementations to include [Puppy](https://github.com/sushichop/Puppy).

### Motivation:

**I think this PR helps people to be able to find the compatible backends more easily.**
_[Details]_
I built the library that supports multiple transports(console, file, syslog, etc.) as loggers. It not only works alone, but also as a backend for swift-log. 
Furthermore, it has file log rotation feature and you can also customize the log format as you like. And it supports both Linux and Darwin.

### Modifications:

Append `sushichop/Puppy` to the list of implementations in the README.

### Result:

This PR makes no changes to code, just the README.


ktoso:

ktoso (approved):
Excellent! Fine work @sushichop, thanks for adding to the list!

PR #177, opened 2021-01-11
Add support for logging in SwiftWasm (via `WASILibc`)

helje5:
This is a port of Logging to the SwiftWasm `WASILibc`.

### Motivation:

I'd like to log things in Wasm as well.

### Modifications:

1. The library was using a lot of `import Glibc` as a fallback (you might want to consider to move to `canImport`). Added additional checks for `WASILibc` and emit an `#error` when no known runtime is available.
2. `WASILibc` (well, WASI) doesn't support threading and hence locks. `#if` guarded the Locks.swift file to not build the Lock classes w/ WASILibc. Then adjust the few invocations on the callsites to not use locks on WASILibc.
3. `WASILibc` doesn't have stdio, but it does have Foundation `print`. Defines a `TextOutputStream` using that when on WASILibc, and do not use the Stdio variant. (implies using the same stream for both stdout/err)
4. `WASILibc` doesn't have `strftime`, but it does have a working Foundation `DateFormatter`/`Date`. Use that instead of timestamp formatting.

### Result:

swift-log can be used in Wasm environments.

Source:
```swift
import Logging
let logger = Logger(label: "zz.log")
logger.info("Hello World!")
```

```sh
helge@M1ni cowsa $ swift build --triple wasm32-unknown-wasi
[3/3] Linking cowsa.wasm
helge@M1ni cowsa $ wasmer .build/debug/cowsa.wasm 
2021-01-11 17:44:06.926 info zz.log : Hello World!
```

### DIFF
--- a/Sources/Logging/Locks.swift
+++ b/Sources/Logging/Locks.swift
@@ -30,7 +30,7 @@
 import Darwin
 #elseif os(Windows)
 import WinSDK
-#else
+#elseif canImport(Glibc)
 import Glibc
 #endif
 
@@ -44,6 +44,8 @@ internal final class Lock {
     #if os(Windows)
     fileprivate let mutex: UnsafeMutablePointer<SRWLOCK> =
         UnsafeMutablePointer.allocate(capacity: 1)
+    #elseif canImport(WASILibc)
+    fileprivate let mutex : Void
     #else
     fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t> =
         UnsafeMutablePointer.allocate(capacity: 1)
@@ -53,6 +55,8 @@ internal final class Lock {
     public init() {
         #if os(Windows)
         InitializeSRWLock(self.mutex)
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         var attr = pthread_mutexattr_t()
         pthread_mutexattr_init(&attr)
@@ -66,11 +70,14 @@ internal final class Lock {
     deinit {
         #if os(Windows)
         // SRWLOCK does not need to be free'd
+        self.mutex.deallocate()
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_mutex_destroy(self.mutex)
         precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
-        #endif
         self.mutex.deallocate()
+        #endif
     }
 
     /// Acquire the lock.
@@ -80,6 +87,8 @@ internal final class Lock {
     public func lock() {
         #if os(Windows)
         AcquireSRWLockExclusive(self.mutex)
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_mutex_lock(self.mutex)
         precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
@@ -93,6 +102,8 @@ internal final class Lock {
     public func unlock() {
         #if os(Windows)
         ReleaseSRWLockExclusive(self.mutex)
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_mutex_unlock(self.mutex)
         precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
@@ -136,6 +147,9 @@ internal final class ReadWriteLock {
     fileprivate let rwlock: UnsafeMutablePointer<SRWLOCK> =
         UnsafeMutablePointer.allocate(capacity: 1)
     fileprivate var shared: Bool = true
+    #elseif canImport(WASILibc)
+    // no threading on WASI yet
+    fileprivate let rwlock: Void
     #else
     fileprivate let rwlock: UnsafeMutablePointer<pthread_rwlock_t> =
         UnsafeMutablePointer.allocate(capacity: 1)
@@ -145,6 +159,8 @@ internal final class ReadWriteLock {
     public init() {
         #if os(Windows)
         InitializeSRWLock(self.rwlock)
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_rwlock_init(self.rwlock, nil)
         precondition(err == 0, "\(#function) failed in pthread_rwlock with error \(err)")
@@ -154,11 +170,14 @@ internal final class ReadWriteLock {
     deinit {
         #if os(Windows)
         // SRWLOCK does not need to be free'd
+        self.rwlock.deallocate()
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_rwlock_destroy(self.rwlock)
         precondition(err == 0, "\(#function) failed in pthread_rwlock with error \(err)")
-        #endif
         self.rwlock.deallocate()
+        #endif
     }
 
     /// Acquire a reader lock.
@@ -169,6 +188,8 @@ internal final class ReadWriteLock {
         #if os(Windows)
         AcquireSRWLockShared(self.rwlock)
         self.shared = true
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_rwlock_rdlock(self.rwlock)
         precondition(err == 0, "\(#function) failed in pthread_rwlock with error \(err)")
@@ -183,6 +204,8 @@ internal final class ReadWriteLock {
         #if os(Windows)
         AcquireSRWLockExclusive(self.rwlock)
         self.shared = false
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_rwlock_wrlock(self.rwlock)
         precondition(err == 0, "\(#function) failed in pthread_rwlock with error \(err)")
@@ -201,6 +224,8 @@ internal final class ReadWriteLock {
         } else {
             ReleaseSRWLockExclusive(self.rwlock)
         }
+        #elseif canImport(WASILibc)
+        // no threading on WASI yet
         #else
         let err = pthread_rwlock_unlock(self.rwlock)
         precondition(err == 0, "\(#function) failed in pthread_rwlock with error \(err)")
--- a/Sources/Logging/Locks.swift
+++ b/Sources/Logging/Locks.swift
@@ -26,12 +26,18 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(WASILibc)
+// No locking on WASILibc
+#else
+
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
 import Darwin
 #elseif os(Windows)
 import WinSDK
 #elseif canImport(Glibc)
 import Glibc
+#else
+#error("Unsupported runtime")
 #endif
 
 /// A threading lock based on `libpthread` instead of `libdispatch`.
@@ -44,8 +50,6 @@ internal final class Lock {
     #if os(Windows)
     fileprivate let mutex: UnsafeMutablePointer<SRWLOCK> =
         UnsafeMutablePointer.allocate(capacity: 1)
-    #elseif canImport(WASILibc)
-    fileprivate let mutex : Void
     #else
     fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t> =
         UnsafeMutablePointer.allocate(capacity: 1)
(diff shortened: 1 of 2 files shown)


ktoso:

Sources/Logging/Logging.swift
```diff
@@ -721,6 +773,10 @@ public struct StreamLogHandler: LogHandler {
     }
 
     private func timestamp() -> String {
+        #if canImport(WASILibc)
+        // WASI doesn't have `strftime`, but Foundation `DateFormatter`
+        return sharedDateFormatter.string(from: Date())
```
weissi, line 778: `swift-log` isn't meant to depend on `Foundation` so we'd need a different solution here.

Sources/Logging/Logging.swift
```diff
@@ -603,6 +629,15 @@ public struct MultiplexLogHandler: LogHandler {
     }
 }
 
+#if canImport(WASILibc)
+// There is no stdio/FILE in WASI yet, neither stdout/err. But it does
+// support `print`.
```
weissi, line 634: how does WASILibc's `Swift.print` actually work?

weissi (changes requested):
I'm mostly happy with this but it does seem to add a Foundation dependency.

Sources/Logging/Logging.swift
helje5: No idea, I thought it might be OK for this specific use case.

Sources/Logging/Logging.swift
helje5: I skimmed over the source but couldn't find a specific change, maybe the WASILibc just doesn't export stdio.h for some reason.

Sources/Logging/Logging.swift
helje5: Looks like they use `strftime` themselves for the replacement DateFormatter, so it seems they just don't re-export that. Which is a little weird: https://github.com/swiftwasm/swift-corelibs-foundation/blob/swiftwasm/Sources/Foundation/DateFormatter.swift#L538

weissi: ha, that's weird. Let's ping @MaxDesiatov maybe he knows what's going on there?

MaxDesiatov: `strftime` does work, I don't see `import WASILibc` anywhere here though for these to be imported

Sources/Logging/Locks.swift
```diff
@@ -26,12 +26,18 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(WASILibc)
+// No locking on WASILibc
+#else
+
```
MaxDesiatov, line 32: This won't be needed if imports are structured the same way they are for other platforms.

```suggestion
```

Sources/Logging/Locks.swift
```diff
@@ -26,12 +26,18 @@
...
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
 import Darwin
 #elseif os(Windows)
 import WinSDK
-#else
+#elseif canImport(Glibc)
```
MaxDesiatov, line 37: This should make `strftime` and `stdout`/`stderr` available.

```suggestion
#elseif os(WASI)
import WASILibc
#elseif canImport(Glibc)
```

Sources/Logging/Locks.swift
```diff
@@ -66,11 +72,12 @@ internal final class Lock {
     deinit {
         #if os(Windows)
         // SRWLOCK does not need to be free'd
+        self.mutex.deallocate()
```
MaxDesiatov, line 75: unsure why the Windows branch is touched here?

Sources/Logging/Locks.swift
```diff
@@ -154,11 +161,12 @@ internal final class ReadWriteLock {
     deinit {
         #if os(Windows)
         // SRWLOCK does not need to be free'd
+        self.rwlock.deallocate()
```
MaxDesiatov, line 164: ditto

Sources/Logging/Logging.swift
```diff
@@ -16,8 +16,16 @@
...
-#else
+#elseif canImport(Glibc)
 import Glibc
+#elseif canImport(WASILibc)
+// WASILibc doesn't have stdio, e.g. things like `FILE` yet
+// WASILibc doesn't have `strftime`, but Foundation `DateFormatter`
+import struct Foundation.Date
+import class  Foundation.DateFormatter
+let sharedDateFormatter = DateFormatter()
+#else
```
MaxDesiatov, line 27:
```suggestion
#elseif os(WASI)
import WASILibc
#else
```

Sources/Logging/Logging.swift
```diff
@@ -281,7 +289,11 @@ extension Logger {
 /// configured. `LoggingSystem` is set up just once in a given program to set up the desired logging backend
 /// implementation.
 public enum LoggingSystem {
+    #if canImport(WASILibc)
```
MaxDesiatov, line 292:
```suggestion
    #if os(WASI)
```

Sources/Logging/Logging.swift
```diff
@@ -281,7 +289,11 @@ extension Logger {
...
+    // WASILibc is single threaded, provides no locks
```
MaxDesiatov, line 293:
```suggestion
    // WASI is currently single threaded, provides no locks
```

Sources/Logging/Logging.swift
```diff
@@ -292,18 +304,28 @@ public enum LoggingSystem {
     /// - parameters:
     ///     - factory: A closure that given a `Logger` identifier, produces an instance of the `LogHandler`.
     public static func bootstrap(_ factory: @escaping (String) -> LogHandler) {
+        #if canImport(WASILibc)
```
MaxDesiatov, line 307:
```suggestion
        #if os(WASI)
```

Sources/Logging/Logging.swift
```diff
@@ -292,18 +304,28 @@ public enum LoggingSystem {
...
             precondition(!self.initialized, "logging system can only be initialized once per process.")
             self.factory = factory
             self.initialized = true
         }
+        #endif
     }
 
     // for our testing we want to allow multiple bootstraping
     internal static func bootstrapInternal(_ factory: @escaping (String) -> LogHandler) {
+        #if canImport(WASILibc)
```
MaxDesiatov, line 322:
```suggestion
        #if os(WASI)
```

Sources/Logging/Logging.swift
```diff
@@ -391,7 +413,11 @@ extension Logger {
     /// - parameters:
     ///     - label: An identifier for the creator of a `Logger`.
     public init(label: String) {
+        #if canImport(WASILibc)
```
MaxDesiatov, line 416:
```suggestion
        #if os(WASI)
```

Sources/Logging/Logging.swift
```diff
@@ -614,12 +649,16 @@ internal struct StdioOutputStream: TextOutputStream {
         string.withCString { ptr in
             #if os(Windows)
             _lock_file(self.file)
+            #elseif canImport(WASILibc)
```
MaxDesiatov, line 652:
```suggestion
            #elseif os(WASI)
```

Sources/Logging/Logging.swift
```diff
@@ -614,12 +649,16 @@ internal struct StdioOutputStream: TextOutputStream {
...
+            // no file locking on WASI
             #else
             flockfile(self.file)
             #endif
             defer {
                 #if os(Windows)
                 _unlock_file(self.file)
+                #elseif canImport(WASILibc)
```
MaxDesiatov, line 660:
```suggestion
                #elseif os(WASI)
```

Sources/Logging/Logging.swift
```diff
@@ -654,22 +694,34 @@ let systemStdout = Darwin.stdout
 #elseif os(Windows)
 let systemStderr = CRT.stderr
 let systemStdout = CRT.stdout
-#else
+#elseif canImport(Glibc)
 let systemStderr = Glibc.stderr!
 let systemStdout = Glibc.stdout!
+#elseif canImport(WASILibc)
+// no stdio on WASI yet
```
MaxDesiatov, line 701:
```suggestion
#elseif os(WASI)
let systemStderr = WASILibc.stderr!
let systemStdout = WASILibc.stdout!
#elseif canImport(Glibc)
let systemStderr = Glibc.stderr!
let systemStdout = Glibc.stdout!
```

Sources/Logging/Logging.swift
```diff
@@ -654,22 +694,34 @@ let systemStdout = Darwin.stdout
...
     }
 
     /// Factory that makes a `StreamLogHandler` to directs its output to `stderr`
     public static func standardError(label: String) -> StreamLogHandler {
-        return StreamLogHandler(label: label, stream: StdioOutputStream.stderr)
+        #if canImport(WASILibc)
+            return standardOutput(label: label)
+        #else
+            return StreamLogHandler(label: label, stream: StdioOutputStream.stderr)
+        #endif
```
MaxDesiatov, line 724: I don't think these two changes will be needed if suggestions above are applied.

Sources/Logging/Logging.swift
MaxDesiatov: `stdio`/`stderr` should be available if you add `import WASILibc`, most probably this `PrintOutputStream` won't be needed.

helje5:
OK, lets put that on hold. Of course I tried w/ importing WASILibc first (obviously). As mentioned it didn't carry either `FILE` nor `strftime`. That was w/ the 5.3.1 release.

If those _do_ work, 95% of my modifications do not make any sense.

helje5:
`FILE` isn't available though, file'd: https://github.com/swiftwasm/swift/issues/2480

Sources/Logging/Locks.swift
helje5: I originally had another case for WASI here (which had no mutex), the change is not necessary anymore anymore since I then excluded the whole Lock class.

Sources/Logging/Logging.swift
helje5: Is `os(WASI)` available in all supported swift-log Swift versions, i.e. starting w/ Swift 5.0?

helje5:
OK, I had another look at this, this is much simpler now:
- no more PrintOutputStream, instead modified the `StdioOutputStream` to use an `OpaquePointer` on WASI (instead of a proper `UnsafeMutablePointer<FILE>`)
- use `strftime`, it is available
- also reverted the changes within Lock, for WASI we just don't use it at all

I didn't do the `os(WASI)` checks suggested (though I agree they are more correct), as I'm not sure whether those are available on Swift 5.0 and `canImport(WASILibc)` should be OK.

This should be worth another review, I tried it and it works w/ the SwiftWasm 5.3.1 toolchain.

weissi:

weissi:

weissi:
> I didn't do the `os(WASI)` checks suggested (though I agree they are more correct), as I'm not sure whether those are available on Swift 5.0 and `canImport(WASILibc)` should be OK.

You can do

```
#if compiler(>=5.3)
#if os(WASIlibc)
```

helje5:
> > I didn't do the `os(WASI)` checks suggested (though I agree they are more correct), as I'm not sure whether those are available on Swift 5.0 and `canImport(WASILibc)` should be OK.
> 
> You can do
> 
> ```
> #if compiler(>=5.3)
> #if os(WASIlibc)
> ```

Yes, but do you really want to add that much `#if` cruft everywhere? I think we better stick to the simpler `#if canImport` then?

weissi:
> Yes, but do you really want to add that much `#if` cruft everywhere? I think we better stick to the simpler `#if canImport` then?

fine by me

helje5:
@MaxDesiatov Is this good now from your perspective?

MaxDesiatov (approved):
Great stuff, thanks!

Sources/Logging/Locks.swift
weissi: resolving all these because @MaxDesiatov approved

weissi (approved):
Looks good to me!

weissi:
@ktoso / @tomerd want to have a look here or good to merge?

ktoso (approved):
Went through and it's looking good, thanks @helje5 -- let's merge.

ktoso:
// Had some ancient review comment from the time of the initial commit still there; deleted, ignore that :)

PR #179, opened 2021-02-09
Suggest more robust commit template

vrat28:
This template focusses on adding more information at the same time limiting the subject because github truncates subject greater than 72 with (...)
This can be useful when using git log,shortlog commands.

### Motivation:

### Modifications:

### Result:


ktoso:
Thanks but I don't think we necessarily need these in the template. 
Thanks again, but closing

Issue #180, opened 2021-02-22
should use fwrite not fputs


weissi:
https://github.com/apple/swift-log/blob/12d3a8651d32295794a850307f77407f95b8c881/Sources/Logging/Logging.swift#L828

We should use `fwrite` and not `fputs` here... Usually doesn't matter unless we print a 0 byte which could be a hard to debug thing if we lose everything after the 0 byte with `fputs`.

weissi:
fixed by #188

himanshu007-creator:
Please close this issue

PR #181, opened 2021-02-25
Use fwrite instead of puts in the Stdio logging

dhruvshah8:
Change fputs to fwrite

 ### Motivation:

https://github.com/apple/swift-log/issues/180

Printing  a 0 byte could be a hard to debug if we lose everything after the 0 byte with fputs or if we accidentally pass a string that's not properly null-terminated, thus using fwrite, it solve that problem.

### Modifications:

replaced fputs with fwrite

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -825,7 +825,7 @@ internal struct StdioOutputStream: TextOutputStream {
                 funlockfile(self.file)
                 #endif
             }
-            _ = fputs(ptr, self.file)
+            _ = fwrite(ptr, 1, strlen(ptr), self.file)
             if case .always = self.flushMode {
                 self.flush()
             }


ktoso:

Sources/Logging/Logging.swift
```diff
@@ -825,7 +825,7 @@ internal struct StdioOutputStream: TextOutputStream {
                 funlockfile(self.file)
                 #endif
             }
-            _ = fputs(ptr, self.file)
+            _ = fwrite(ptr, 1, strlen(ptr), self.file)
```
weissi, line 828: That's not quite enough because this will also terminate on `0` bytes. Try for example (that's actually worth a unit test) logging this string `"hello\u{0} world"`. It'll only print the `hello` not the `world` I think.

The right fix is to change how we obtain the pointer from `withCString` to

```
var contiguousString = string
#if compiler(>=5.1)
contiguousString.makeContiguousUTF8()
#else
contiguousString = string + "" // this is the Swift 5.0 hack to make the string contiguous
#endif

contiguousString.utf8.withContiguousStorageIfAvailable { utf8Bytes /* this has a pointer and length */ in
   // exising code
}! /* the bang is safe because it's guaranteed contiguous */
```

weissi (changes requested):
Thank you! This is a great start, we're almost there, just need the string length and a test for the fix.

ktoso:
Thanks for the PR, I changed title to be less generic.

ktoso:
Hi there @dhruvshah8 would you like to follow up on @weissi's comment?

We would like to cut a 1.4.2 release and this could be included in it, thanks!

weissi:
closing in favour of #188

PR #185, opened 2021-03-03
Fix function being dropped on source-less trace

### LINKED ISSUE #184: Function parameter dropped in trace
### Expected behavior
When I perform `someLogger.trace("hello", function: "someFunction()")` I expect that the function parameter that reaches the underlying LogHandler will be, `someFunction()`.

### Actual behavior
In practice, the function parameter will be `trace(_:metadata:file:function:line:)`.

SwiftLog 1.4.1

```
Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)
Target: x86_64-apple-darwin19.6.0
Darwin Sauls-MacBook-Pro-16-inch-2019 19.6.0 Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64 x86_64
```

### Further information

When adding back the accidentally removed functions on #152, the source-less `trace` call doesn't pass the value of `function` when calling `trace` with the `source` parameter. As a result, the value of `function` is always `trace(_:metadata:file:function:line:)` for all calls to `trace` where a source is not set.

saulbaro:
### Motivation:

When using the recently re-introduced source-less version of trace, the function parameter that reaches the underlying LogHandler is wrong.

### Modifications:

Pass the function parameter to trace with source parameter.

### Result:

* With these changes, the function parameter that reaches the underlying LogHandler has the correct value.
* Fixes #184

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -176,7 +176,7 @@ extension Logger {
     public func trace(_ message: @autoclosure () -> Logger.Message,
                       metadata: @autoclosure () -> Logger.Metadata? = nil,
                       file: String = #file, function: String = #function, line: UInt = #line) {
-        self.trace(message(), metadata: metadata(), source: nil, file: file, line: line)
+        self.trace(message(), metadata: metadata(), source: nil, file: file, function: function, line: line)
     }
 
     /// Log a message passing with the `Logger.Level.debug` log level.


ktoso:

ktoso (approved):
Ouch! Thanks for spotting this and the nice report and fix.

I skimmed the other functions, the remaining ones seem fine...

We could cut a quick 1.4.2 if you want, WDYT?

saulbaro:
Hey @ktoso, thanks to you for reviewing this so quickly :) 

Cutting a new release would be great. We've avoided the issue on our end for now by pinning the previous release, but we'd love to revert that back.

Regarding the other functions, this issue was flagged by our unit tests, and it was the only failing one, so I am confident that it is the only one with this issue 🤞

ktoso:
FYI 1.4.2 released 👍 https://github.com/apple/swift-log/releases/tag/1.4.2

seanfoltz:
Hi

PR #186, opened 2021-03-04
Remove references to Swift 4

steipete:
I found https://github.com/apple/swift-log/releases/tag/0.0.0 however these are just shims.
Seems unlikely that this is still needed.


ktoso:
Sounds good, thanks for the cleanup

PR #187, opened 2021-03-17
use #fileID for Swift 5.3+

weissi:
Motivation:

Right now, we use the full path (`#file`) for the logging as well as for
parsing out the module name (which goes into `source` by default).

This has two problems:

1. The file names contain the build server's full path, ie. aren't
   portable across different compiles (and are very long).
2. The module name parsing fails if we log from a file that's not in a
   directory named just like the module.

Modification:

Switch `#file` for `#fileID`. This changes a bunch of things:

- `source` is always the module name by default (great!)
- `file` changes from `/the/full/path/on/disk` to `ModuleName/FileBaseName.swift` (I think better, but it's a change)

Result:

`source` is more correct, `file` will be different.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -49,6 +49,7 @@ public struct Logger {
 }
 
 extension Logger {
+    #if compiler(>=5.3)
     /// Log a message passing the log level as a parameter.
     ///
     /// If the `logLevel` passed to this method is more severe than the `Logger`'s `logLevel`, it will be logged,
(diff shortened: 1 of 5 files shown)


weissi:
@ktoso / @tomerd this is mostly an RFC

ktoso (approved):
LGTM, it's a change in logged info but I don't think that's something we need to major bump for. For what it's worth, it was kind of sometimes broken before as you explain... 👍

weissi:
@ktoso I agree that we don't need a major. In theory it could break some backends (which parse the path) but I don't think we really guaranteed that it's the full file path...

weissi:

tomerd:
late LGTM

PR #188, opened 2021-04-04
Improve StdioOutputStream with fwrite (#180)

felipejinli:
# Replaced `fputs` with `fwrite`, added test and made spelling corrections

### _WIP help_
_(for reference, see the code changes)_
On the testing side, I'm currently using `String(validatingUTF8: <unsafe-mutable-pointer>` to turn a UnsafeMutablePointer<Int8> into a String. However, this seems to return a string that stops at the first 0 byte E.g. for a "hello\u{0} world" log it outputs hello only. NB. I have checked that `fwrite` correctly logs a message longer than "hello" through count but testing this is hard. @weissi @ktoso Have you got any suggestions?

### Motivation:

_If a 0 byte is logged, `fputs` would not output the content after the 0 byte making it harder to debug (see #180). `fwrite` uses a count argument so content after 0 byte can be logged._

### Modifications:

_Replaced `fputs` with the correct call to `fwrite`.   Added helper internal func `contiguousUTF8(_ string: String) -> String.UTF8View`. Added `testStdioOutputStreamWrite()` to test ability to log content after 0 byte._

### Result:

_Improved debugging experience_

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -808,7 +808,7 @@ internal struct StdioOutputStream: TextOutputStream {
     internal let flushMode: FlushMode
 
     internal func write(_ string: String) {
-        string.withCString { ptr in
+        contiguousUTF8(string).withContiguousStorageIfAvailable { utf8Bytes in
             #if os(Windows)
             _lock_file(self.file)
             #elseif canImport(WASILibc)
@@ -825,17 +825,27 @@ internal struct StdioOutputStream: TextOutputStream {
                 funlockfile(self.file)
                 #endif
             }
-            _ = fputs(ptr, self.file)
+            _ = fwrite(utf8Bytes.baseAddress!, 1, utf8Bytes.count, self.file)
             if case .always = self.flushMode {
                 self.flush()
             }
-        }
+        }!
     }
 
     /// Flush the underlying stream.
     /// This has no effect when using the `.always` flush mode, which is the default
     internal func flush() {
         _ = fflush(self.file)
+    }
+
+    internal func contiguousUTF8(_ string: String) -> String.UTF8View {
+        var contiguousString = string
+        #if compiler(>=5.1)
+        contiguousString.makeContiguousUTF8()
+        #else
+        contiguousString = string + ""
+        #endif
+        return contiguousString.utf8
     }
 
     internal static let stderr = StdioOutputStream(file: systemStderr, flushMode: .always)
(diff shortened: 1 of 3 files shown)


ktoso:

Sources/Logging/Logging.swift
```diff
@@ -825,7 +825,7 @@ internal struct StdioOutputStream: TextOutputStream {
                 funlockfile(self.file)
                 #endif
             }
-            _ = fputs(ptr, self.file)
+            _ = fwrite(utf8Bytes.baseAddress!, 1, utf8Bytes.count, self.file)
             if case .always = self.flushMode {
                 self.flush()
             }
```
weissi, line 831: mind adding a `!` to the line below this one (github doesn't let me comment there)? We need to make sure that the `withContiguousStorageIfAvailable` actually worked. It will always work (because your `contiguousUTF8` is correct) but the compiler can't prove this. And just to prevent mistakes by refactorings in the future, we should make sure this actually works.

weissi (changes requested):
Thank you! This is pretty much perfect! Just a `!` missing, then we can ship this.

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -723,6 +723,25 @@ class LoggingTest: XCTestCase {
...
+        self.withWriteReadFDsAndReadBuffer { writeFD, readFD, readBuffer in
+            let logStream = StdioOutputStream(file: writeFD, flushMode: .always)
+            LoggingSystem.bootstrapInternal { StreamLogHandler(label: $0, stream: logStream) }
+            let log = Logger(label: "test")
+            let testString = "hello\u{0} world"
+            log.critical("\(testString)")
+
+            read(readFD, readBuffer, 256)
+
+            if let output = String(validatingUTF8: readBuffer) {
```
weissi, line 736: I think a better way to write this would be

```swift
if let output = String(decoding: UnsafeRawBufferPointer(start: UnsafeRawPointer(readBuffer), count: returnValueOfRead), as: UTF8.self) {
```

because currently, you're relying on the initialisation of the read buffer.

felipejinli:
@weissi Makes sense - I've just incorporated the changes. It should work now!

weissi (approved):
Awesome, thank you!

weissi:

weissi:
Sorry @felipejinli , could you run `scripts/generate_linux_tests.rb` to generate the Linux test list, that should make the soundness bot happy.

felipejinli:
@weissi done!

weissi:

weissi:

PR #194, opened 2021-05-05
Add `SwiftLogNoOpLogHandler.init(_: String)`

glbrntt:
Motivation:

`LogHandler` facotories are given a label as input and return a
`LogHandler`. Since the no-op handler doesn't care about the label
callers must provide a thunk: `{ _ in SwiftLogNoOpLogHandler() }`.
That's fine but slightly more annoying to type than it needs to be.

Modifications:

- Add `SwiftLogNoOpLogHandler.init(_: String)`

Result:

Users can create a logger with the no-op handler with:

```swift
Logger(label: "foo", factory: SwiftLogNoOpLogHandler.init)
```

instead of:

```swift
Logger(label: "foo", factory: { _ in SwiftLogNoOpLogHandler() })
```

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1147,6 +1147,8 @@ public struct StreamLogHandler: LogHandler {
 public struct SwiftLogNoOpLogHandler: LogHandler {
     public init() {}
 
+    public init(_: String) {}
+
     @inlinable public func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, file: String, function: String, line: UInt) {}
 
     @inlinable public subscript(metadataKey _: String) -> Logger.Metadata.Value? {


ktoso (approved):
LGTM, makes sense to me -- happy to add 👍

weissi (approved):
makes sense, thank you!

ktoso:
Do people mind doing this on in 1.4.3 or should we 1.5.0? Technically it's new API

weissi:
> Do people mind doing this on in 1.4.3 or should we 1.5.0? Technically it's new API

Let's do 1.5.0, we don't need to release 1.4.3 at all.

glbrntt:
+1 for 1.5.0, version numbers are cheap!

ktoso:
Cool, I can cut that tomorrow morning, unless you're in a rush :-0

glbrntt:
No rush at all, this was just an itch I needed to scratch.

PR #198, opened 2021-06-06
Tests: enable tests on Windows

compnerd:
This adds support for running the tests on Windows.  However, running
the tests on Windows requires SPM currently.  While it is possible to
enable executing tests with CMake, the expectation is that SPM builds
would be the most common and CMake is largely used for bootstrapping.
The changes here require alterations to remove the POSIX threading to
get the thread ID.  Additionally, pipes on Windows are not similar to
pipes on Unix platforms. Finally, we must add Windows to the platform
list, which is temporary (it would be better to use test discovery in
SPM).

### Motivation:

### Modifications:

### Result:


compnerd:
CC: @tomerd

tomerd (approved):
lgtm

ktoso (approved):
Looks good to me, thanks!

I remain nervous about "we kind of _try to_ support windows but don't have CI for it" for it here 🤔 Hopefully we can address this someday

PR #199, opened 2021-06-06
Logging: avoid a deprecation warning on Windows

compnerd:
Windows recommends the safety extension usage which in this case
explicitly allow you to be informed of failures.  This just avoids an
annoying warning when using logging in projects.

### Motivation:

### Modifications:

### Result:


compnerd:
CC: @tomerd

tomerd (approved):

ktoso (approved):
LGTM, thanks!

PR #202, opened 2021-07-29
Change StreamLogHandler init to public

ye4241:
Change StreamLogHandler init to public

### Motivation:

If we just want to use TextOutputStream same as default StreamLogHandler, we can use the init to init ourself StreamLogHandler, eg, with FileHandle.

### Modifications:

Change to StreamLogHandler init from internal to public

### Result:

We can now use StreamLogHandler more fluently.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1103,8 +1103,7 @@ public struct StreamLogHandler: LogHandler {
         }
     }
 
-    // internal for testing only
-    internal init(label: String, stream: TextOutputStream) {
+    public init(label: String, stream: TextOutputStream) {
         self.label = label
         self.stream = stream
     }


ktoso:
I'll be the bearer of bad news I guess -- we do not intend to expand the capabilities of the provided handlers (or even, we'd rather not ship them at all (!)), and instead I would urge you to give one of the excellent implementation libraries a try: https://github.com/apple/swift-log#selecting-a-logging-backend-implementation-applications-only

swift-log is not intended to have any reasonable / largely useful feature set of the _handlers_ -- it is merely intended to be an API package that other handlers implement. We really do not want to allow using these stream log handlers in production etc. Please consider using one of the real backends.

ktoso:
Closing since stale since a while; please comment if you'd like to discuss more

PR #206, opened 2021-08-13
[WIP] Metadata providers

### LINKED ISSUE #205: Metadata providers
While it's currently possible to have "semi-global" metadata by attaching metadata values to a `Logger` itself there's no notion of globally stating that the **current value** of something should always be included in log statements.

This would be great to have for example for tracing where the user may want to include the current trace ID in each log statement. Instead of manually pulling out the trace ID (from task locals) before each log statement we could make it a bit easier by having a closure to be invoked for each statement automatically. In that closure the user could look up and return the current trace ID. Also, multiple of such closures could be provided so that e.g. a tracer implementation could take care of providing the trace ID, and the user could provide something application-specific which is also stored in a task local.

```swift
// example of a tracer invoking `provideMetadata`
// (`Baggage.current` is a task-local)

LoggingSystem.provideMetadata {
    return Baggage.current.traceID.map { ["trace-id": "\($0)"] }
}
```

slashmo:
Add `MetadataProvider` closure to be run on each log statement to attach the current value of something (e.g. a task-local) to the statement's metadata values (#205).

### Motivation:

Closes #205

### Modifications:

- `LoggingSystem` gains a new `provideMetadata` method
- `LoggingSystem` stores an array of `MetadataProvider`s
- `Logger` gains a new initializer taking an overwriting `MetadataProvider`
- `Logger` executes the `MetadataProvider` before logging a statement, attaching the resulting values to the statement's metadata

### Result:

It becomes easier to attach metadata to each log statement where the value can be derived without a method argument such as retrieving from the current task local storage. Instead of writing the code to retrieve such values per log statement it can instead be set up once.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -39,12 +39,19 @@ public struct Logger {
     @usableFromInline
     var handler: LogHandler
 
+    @usableFromInline
+    var metadataProvider: MetadataProvider
+
     /// An identifier of the creator of this `Logger`.
     public let label: String
 
-    internal init(label: String, _ handler: LogHandler) {
+    /// A closure invoked for each log statement to attach additional metadata.
+    public typealias MetadataProvider = () -> Metadata?
+
+    internal init(label: String, _ handler: LogHandler, metadataProvider: @escaping MetadataProvider) {
         self.label = label
         self.handler = handler
+        self.metadataProvider = metadataProvider
     }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/Logging/Logging.swift
```diff
@@ -76,9 +83,24 @@ extension Logger {
...
+                        return providerMetadata.merging(metadata) { _, b in
+                            b
+                        }
+                    } else {
+                        return metadata
+                    }
+                } else {
+                    return providerMetadata
+                }
+            }()
```
stevapple, line 100: I wonder if the implementation could be clearer using `guard-let`.

```swift
            let m: Metadata? = {
                guard let metadata = metadata() else {
                    return providerMetadata
                }
                guard let providerMetadata = providerMetadata else {
                    return metadata
                }
                // one-off metadata overwrites provider metadata
                return providerMetadata.merging(metadata) { $1 }
            }()
```

Sources/Logging/Logging.swift
```diff
@@ -92,9 +114,24 @@ extension Logger {
...
```
stevapple, line 131: Also use `guard-let` here.

Sources/Logging/Logging.swift
```diff
@@ -657,6 +694,23 @@ public enum LoggingSystem {
...
+                if let providedMetadata = provider() {
+                    if let existingMetadata = metadata {
+                        return existingMetadata.merging(providedMetadata) { _, b in
+                            b
+                        }
+                    }
+                    return providedMetadata
+                }
+                return metadata
+            }
```
stevapple, line 709: Also here:
```swift
            _metadataProviders.reduce(nil) { metadata, provider in
                guard let providedMetadata = provider() else {
                    return metadata
                }
                guard let existingMetadata = metadata else {
                    return providedMetadata
                }
                return existingMetadata.merging(providedMetadata) { $1 }
            }
```

slashmo:
This PR is now superseded by https://github.com/apple/swift-log/pull/235

Issue #216, opened 2022-01-28
make Logger Sendable


weissi:
`Logger` should be `Sendable` because that's required almost everywhere now :).

ktoso:
Yeap sounds good -- can you send in a pr?

weissi:
It's complicated. `Logger` obvs needs a ref to the `LogHandler` and `LogHandler` is the `public protocol`. So we'd need to require `LogHandler` to be `Sendable` which is SemVer major...

So want me to make it `@unchecked Sendable`? That kinda feels not right either. Said that I think the API docs require thread-safetey so we could do that 😬 . WDYT?

Joannis:
I don't think that `@unchecked Sendable` is a good solution here, and I strongly believe that a semver major that requires your log handlers to be `Sendable` is a good move.

ffried:
With Swift 5.6, the LogHandler could inherit from `Sendable` and be annotated with `@preconcurrency`. This would spit out a warning on `LogHandler` implementations that don't fulfill the `Sendable` requirements.

tomerd:
for visibility: https://github.com/apple/swift-log/pull/218

tomerd:
resolved via https://github.com/apple/swift-log/pull/218

PR #217, opened 2022-02-04
Enforce Sendable conformance for LogHandlers Swift-Log, and make Logger Sendable

Joannis:
Enforce Sendable conformance for LogHandlers Swift-Log, and make Logger Sendable. See #216 

### Motivation:

Swift-Log is adopted in many libraries, and as the language has recently involved in exciting new directions, these new features have been the subject of improvement and innovation. Many libraries aim at implementing `Sendable`, which means that many types that carry swift-log's `Logger` will quickly see warnings pop up all over these projects.

As swift-log is a loved and essential part of the swift ecosystem, it too should adopt the new Swift standards.

### Modifications:

- Require `LogHandler` implementations to conform to `Sendable`
- Conform `Logger` to `Sendable`
- Updated tests to support the new Sendable requirements

### Result:

Swift-Log supports Sendable. Requires iOS 13+, macOS 10.15+ and Swift 5.5+.
This'll obviously require a semver major.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.5
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the Swift Logging API open source project
@@ -17,6 +17,10 @@ import PackageDescription
 
 let package = Package(
     name: "swift-log",
+    platforms: [
+        .iOS(.v13),
+        .macOS(.v10_15),
+    ],
     products: [
         .library(name: "Logging", targets: ["Logging"]),
     ],
--- a/Sources/Logging/LogHandler.swift
+++ b/Sources/Logging/LogHandler.swift
@@ -113,7 +113,7 @@
 /// Please note that the above `LogHandler` will still pass the 'log level is a value' test above it iff the global log
 /// level has not been overridden. And most importantly it passes the requirement listed above: A change to the log
 /// level on one `Logger` should not affect the log level of another `Logger` variable.
-public protocol LogHandler {
+public protocol LogHandler: Sendable {
     /// This method is called when a `LogHandler` must emit a log message. There is no need for the `LogHandler` to
     /// check if the `level` is above or below the configured `logLevel` as `Logger` already performed this check and
     /// determined that a message should be logged.
@@ -126,7 +126,7 @@ public protocol LogHandler {
     ///     - file: The file the log message was emitted from.
     ///     - function: The function the log line was emitted from.
     ///     - line: The line the log message was emitted from.
-    func log(level: Logger.Level,
+    @Sendable func log(level: Logger.Level,
              message: Logger.Message,
              metadata: Logger.Metadata?,
              source: String,
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -35,7 +35,7 @@ import WASILibc
 ///
 ///     logger.info("Hello World!")
 ///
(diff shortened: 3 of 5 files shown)


Joannis:
That bot is very eager for a review!

ktoso:

Package.swift
```diff
@@ -17,6 +17,10 @@ import PackageDescription
 
 let package = Package(
     name: "swift-log",
+    platforms: [
+        .iOS(.v13),
+        .macOS(.v10_15),
+    ],
```
ktoso, line 23: can't just do that like this; we need to support old platforms too

Sources/Logging/LogHandler.swift
```diff
@@ -113,7 +113,7 @@
 /// Please note that the above `LogHandler` will still pass the 'log level is a value' test above it iff the global log
 /// level has not been overridden. And most importantly it passes the requirement listed above: A change to the log
 /// level on one `Logger` should not affect the log level of another `Logger` variable.
-public protocol LogHandler {
+public protocol LogHandler: Sendable {
```
ktoso, line 116: can't just do it like this, we'd need to follow the trickery with the sendable availability as explained in https://github.com/swift-server/guides/blob/main/docs/concurrency-adoption-guidelines.md#backwards-compatibility-of-declarations-and-checked-swift-concurrency

So we'd need `_SwiftLog_Sendable` and the entire dance

Sources/Logging/Logging.swift
```diff
@@ -1074,6 +1074,8 @@ let systemStdout = WASILibc.stdout!
 /// `StreamLogHandler` is a simple implementation of `LogHandler` for directing
 /// `Logger` output to either `stderr` or `stdout` via the factory methods.
 public struct StreamLogHandler: LogHandler {
+    public typealias Stream = TextOutputStream & Sendable
```
ktoso, line 1077: has to use the compatibility trickery

ktoso (changes requested):
Not quite right, we need to treat this very carefully so older clients can still adopt even if they are on old swift IMHO

ktoso:
Sendable conformance has been done in https://github.com/apple/swift-log/pull/218 in a more compatible way.

Thank you for the PR though!

easydev991:
> Sendable conformance has been done in #218 in a more compatible way.
> 
> Thank you for the PR though!

Hello! I still get warning about Logger not being `Sendable` inside a `Sendable`-class. Is it a new feature in Xcode 15.3, or is it a bug?

Lukasa:
Can you show your code? Are you storing it as `var`?

easydev991:
> Can you show your code? Are you storing it as `var`?

Cannot show the code now.
I store it as a `let` inside a `@MainActor`-class

Lukasa:
What version of swift-log have you resolved?

easydev991:
> What version of swift-log have you resolved?

I do not use swift-log, I `import OSLog` and then use `Logger` inside a class like this:
```swift
import OSLog

@MainActor
protocol SomeProtocol {}

final class MyClass: SomeProtocol {
  private let logger = Logger(subsystem: "Subsystem", category: "MyClass")
}
```

An interesting moment: in one file it helped me to make `@preconcurrency import OSLog`, but in other file it does not work. I have strict concurrency check (complete) in my app build settings.

Lukasa:
This repository is not for os log, it's for swift-log. Issues with os log should be taken to [Apple's developer forums](https://developer.apple.com/forums/).

PR #218, opened 2022-03-15
adopt sendable

tomerd:
motivation: adjust to swift 5.6

changes:
* define sendable shims for protocols and structs that may be used in async context
* adjust tests
* add a test to make sure no warning are emittted


tomerd:

Sources/Logging/Logging.swift
```diff
@@ -707,16 +721,19 @@ extension Logger {
...
+    public enum MetadataValue: _SwiftLogSendable {
         /// A metadata value which is a `String`.
         ///
         /// Because `MetadataValue` implements `ExpressibleByStringInterpolation`, and `ExpressibleByStringLiteral`,
         /// you don't need to type `.string(someType.description)` you can use the string interpolation `"\(someType)"`.
         case string(String)
 
         /// A metadata value which is some `CustomStringConvertible`.
+        #if compiler(>=5.6)
+        case stringConvertible(CustomStringConvertible & Sendable)
```
fabianfett, line 733: Why does the value need to be Sendable here?

ffried: Because otherwise `MetadataValue` cannot conform to `Sendable`, which in term means `Logger.Metadata` cannot conform to `Sendable`, which in the end means `Logger` can't conform to `Sendable`.

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -889,3 +889,13 @@ extension Logger {
     }
     #endif
 }
+
+// Sendable
+
+#if compiler(>=5.6)
+// @unchecked Sendable since not thread-safe by definition
+extension LoggingTest.LazyMetadataBox: @unchecked Sendable {}
```
fabianfett, line 897: We should not mark things that are `"not thread-safe by definition"` as Sendable, should we?

Sources/Logging/Logging.swift
fabianfett: > Because otherwise MetadataValue cannot conform to Sendable, which in term means Logger.Metadata cannot conform to Sendable, which in the end means Logger can't conform to Sendable.

Yeah, I get that. However we could also transform `CustomStringConvertible` to a string directly on the call-site. This would allow us to not require `Sendable`. However a class that might change its string representation at a later point, would always be represented by its original value, if it is added to the metadata. 

> This wouldn't work either, because this would be a new protocol. Currently, it only requires `CustomStringConvertible`.

Very good point. Going directly to string might potentially allow us to work around the semver major here.

tomerd: @DougGregor any advice here to avoid the need in a major API breakage?

Tests/LoggingTests/LoggingTest.swift
tomerd: what do you have in mind to deal with this?

Sources/Logging/Logging.swift
ktoso: > Yeah, I get that. However we could also transform CustomStringConvertible to a string directly on the call-site. 

That's not an option; the entire purpose of this case is to delay rendering the string; @ffried mentioned this already I just noticed.

Maybe a controversial take but... the `& Sendable` will not cause errors (!), until Swift 6... so, I'm not sure we need to consider it breaking -- rather, this gives people the right `warning` and they have time to fix the issue before it turns into an error.

tomerd: @ktoso are you saying that adding the Sendable protocol requirement will only yield warnings until Swift 6?

ktoso: That's right, I just confirmed with Doug as well. If you were seeing an error in beta-1 (maybe?), then Doug says this could have been a bug and is still just a warning nowadays and will continue to be.

tomerd: in that case I am also comfortable with this soft break. 

@ffried @fabianfett @weissi @Lukasa wdyt?

ktoso (approved):

tomerd:
lets take https://github.com/apple/swift-log/pull/222 first

Sources/Logging/Logging.swift
ffried: Same here. A warning is fine. 👍
It's the same as when for example a new deprecation is added. There would also be a warning then.

Lukasa: I'm ok with a warning here.

Tests/LoggingTests/LoggingTest.swift
Lukasa: If it's not used in a thread-crossing context, why does it need to be `Sendable` at all?

Lukasa: Ok in this case I think it might be best to explain that in this comment.

tomerd:
@yim-lee @weissi @Lukasa @ktoso @fabianfett  this is ready to go. ptal

Tests/LoggingTests/TestSendable.swift
```diff
@@ -0,0 +1,43 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Logging API open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the Swift Logging API project authors
```
yim-lee, line 5:
```suggestion
// Copyright (c) 2022 Apple Inc. and the Swift Logging API project authors
```

yim-lee (approved):

Lukasa (approved):

PR #222, opened 2022-06-28
better abstration for LoggingSystem state

tomerd:
motivation: prepare for sendable checks

changes:
* abstract the LoggingSystem state into a "boxed" class that handles the locking
* adjust call sites


Sources/Logging/Logging.swift
```diff
@@ -665,28 +659,54 @@ public enum LoggingSystem {
...
         #endif
+        fileprivate var _underlying: (String) -> LogHandler
+        private var initialized = false
+
+        init(_ underlying: @escaping (String) -> LogHandler) {
+            self._underlying = underlying
+        }
+
+        func replaceFactory(_ factory: @escaping (String) -> LogHandler, validate: Bool) {
+            #if canImport(WASILibc)
```
weissi, line 688: shouldn't we push that into the ReaderWriterLock and just do nothing in WASI?

tomerd: l kept this "as is" based on the comment https://github.com/apple/swift-log/pull/222/files#diff-f5985f666e1e706d02b520068f87ee07c97e376122fbd685314e088e0d532bb0R676

weissi: Yeah, but we can just put the do-nothing code into the `ReaderWriterLock`. That has less logic and is already `#if`-def'd for Windows etc already, right?

tomerd: i see, yes that is true

weissi (approved):
LGTM

PR #226, opened 2022-07-26
[docs] Minimal docc setup and landing page

ktoso:
Minimal setup for docc documentation.

Honestly I'm very tempted to rip out most of the `README.md` and put it into docc, any voices against doing that?

README could still contain the list of implementations, as that's perhaps easier to just look up in the README while browsing github, but all the other information in README should be moved over to the docc docs -- sounds good?

Screenshot 2022-07-26 at 16 08 04


Package@swift-5.6.swift
```diff
@@ -20,6 +20,9 @@ let package = Package(
     products: [
         .library(name: "Logging", targets: ["Logging"]),
     ],
+    dependencies: [
```
ktoso, line 23: To preview docs: ` swift package --disable-sandbox preview-documentation`

Sources/Logging/Logging.swift
```diff
@@ -103,7 +104,7 @@ extension Logger {
 
     /// Log a message passing the log level as a parameter.
     ///
-    /// If the `logLevel` passed to this method is more severe than the `Logger`'s `logLevel`, it will be logged,
+    /// If the ``logLevel`` passed to this method is more severe than the `Logger`'s `logLevel`, it will be logged,
```
yim-lee, line 107:
```suggestion
    /// If the ``logLevel`` passed to this method is more severe than the ``Logger``'s ``logLevel``, it will be logged,
```

Sources/Logging/Logging.swift
```diff
@@ -27,14 +27,15 @@ import WASILibc
 /// A `Logger` is the central type in `SwiftLog`. Its central function is to emit log messages using one of the methods
```
yim-lee, line 27: All the \`Logger\` references should be \`\`Logger\`\`?

Sources/Logging/Documentation.docc/Logging.md
```diff
@@ -0,0 +1,30 @@
+# ``Logging``
```
tomerd, line 1: how about we standardize on `Sources/<Module>/Documentation.docc/index.md` or `Sources/<Module>/Docs.docc/index.md`  for the "landing page"

ktoso: I'll check if that works... docc has a lot of weird assumptions about "same name".

Filename of the md file can be different indeed though -- index.md could be good! 👍

ktoso: Ok yeap that works -- sounds good to me.

Sources/Logging/Logging.swift
ktoso: This I left purposefully like this since this IS the `Logger` docs, so I don't think we need to link TO the `Logger` 🤔

ktoso:
Need to swap the Package.swift style here

ktoso:
Swapped the package style, merging

Issue #228, opened 2022-08-12
Swift 5.6 and 1.4.3 cause sendable errors on stringConvertible


drekka:
### Expected behavior

Packages that depend on SwiftLog 1.4.n should compile when package resolution updates SwiftLog from 1.4.2 to 1.4.3.

The discussion in https://github.com/apple/swift-log/pull/218 indicates to me that the developers thought adding Sendable would only produce a warning. Yet over here we are seeing compilation failures and errors being issued.

### Actual behavior

Packages using `Logger.MetadataValue.stringConvertable(CustomStringConvertible)` fail compilation if the passed `CustomStringConvertible` is also not `Sendable`.

### Steps to reproduce

With an established package that uses SwiftLog 1.4.2 and something like:

```swift
logger.log(level: self.logLevel,"...",metadata: ["abc": .stringConvertible(value)])
```

where `value` has not had `Sendable` applied. 

With 1.4.2 this would compile. With 1.4.3 it fails to compile. Placing an error on `value` and indicating it has to conform to `Sendable`.

### If possible, minimal yet complete reproducer code (or URL to code)

Sorry, don't have anything I can give at this time.

### SwiftLog version/commit hash

https://github.com/apple/swift-log/pull/218

342d57b6928a982ab36a43597e252caa529cef8e

ktoso:
This only emits a warning on 5.7, I double checked: 

What exact versions and flags are you using?
Are you using -warn-concurrency which would turn warnings like these into errors? If so, you're a bit jumping ahead; there will be many warnings what there are no workaround for in today's language, so I'd suggest not using this flag in projects unless "poking around".

drekka:
Error: 

![Screen Shot 2022-08-12 at 5 31 27 pm](https://user-images.githubusercontent.com/153837/184306468-798c15a5-b3ed-48f6-8bc4-9df56da1cd8e.png)

* SwiftLog 1.4.3, Xcode 13.4.1: - 
   ```
   derekclarkson@DereksDigIOMBP Edna % xcrun --find swift
   /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift
   derekclarkson@DereksDigIOMBP Edna % xcrun swift -version
   swift-driver version: 1.45.2 Apple Swift version 5.6.1 (swiftlang-5.6.0.323.66 clang-1316.0.20.12)
   ```
* Brand new SPM based project so there's no project file. 
[...]

PR #229, opened 2022-08-12
[sendable] Sendable conformance checks cause errors on 5.6 but warnings on 5.7

ktoso:
5.6 specific follow up to https://github.com/apple/swift-log/pull/218

This would be one way to solve https://github.com/apple/swift-log/issues/228 because a missing conformance here is diagnosed as a WARNING on 5.7, but as an ERROR on 5.6.

Other checks are fine on 5.6 AFAICS though.

cc @DougGregor

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1276,9 +1276,14 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
 
 // MARK: - Sendable support helpers
 
+#if compiler(>=5.7.0)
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

Lukasa:
Note the 5.6 failure, you'll need to make this `@unchecked Sendable` for 5.6.

ktoso:
Hah right, CI just reminded me of that as well. Painful but it seems that'll help...

Thanks for having a look @Lukasa

ktoso:
Okey, finally got the #if dance right.

tomerd (approved):

PR #231, opened 2022-08-23
=docc Move to multiple Package.swift files

ktoso:
As we agreed in offline discussions we'll adopt this style for the time being.

Also added the preview script


ktoso:
This is addressing review comments form https://github.com/apple/swift-log/pulls?q=is%3Apr+is%3Aclosed that I missed -- moved to multiple package files

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.7
```
yim-lee, line 1: Keep 5.6 for now?

scripts/preview_docc.sh
```diff
@@ -0,0 +1,16 @@
...
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.md for the list of Swift Logging API project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+xcrun swift package --disable-sandbox preview-documentation --target Logging
```
yim-lee, line 16: Can remove `xcrun`.

ktoso: Follow up in https://github.com/apple/swift-log/pull/231

Package.swift
ktoso: You're right thanks https://github.com/apple/swift-log/pull/231

PR #235, opened 2022-09-30
RFC: Metadata Providers & Distributed Tracing Support in swift-log

slashmo:
Adds a new construct to Logging called `MetadataProvider`, used for automatically including context-dependent metadata such as trace IDs.

> **Note**
> See [`proposals/0001-metadata-providers.md`](https://github.com/apple/swift-log/blob/921cb5344cf0e59b9fdb963c2003b564d51e9932/proposals/0001-metadata-providers.md) for more information

### Motivation:

While global metadata attributes may be manually set on a `LogHandler` level, there's currently no way of reliably providing contextual, automatically propagated, metadata when logging with swift-log.

To benefit from tools such as [Distributed Tracing](https://github.com/apple/swift-distributed-tracing) it is necessary for libraries to make use of the trace information.

Most notably, loggers should participate in tracing by including some trace metadata (such as e.g. a `trace-id`) when logging, as it transparently enables developers to benefit from log correlation using those IDs.

Today, the only supported way of providing such metadata is to pass them along to each log call explicitly:

```swift
logger.info("first this ...", metadata: ["trace-id": MyTracingLibrary.currentTraceID])
logger.info("... now this", metadata: ["trace-id": MyTracingLibrary.currentTraceID])
```

### Modifications:

- Logging depends on `swift-distributed-tracing-baggage`
- Logging exposes a new `Logger.MetadataProvider` struct
- `log` APIs pick up task-local `Baggage` and automatically populate log statements with contextual metadata
- Logging exposes new `log` overloads for explicitly passing `Baggage` if code runs on an event loop e.g. where task-locals cannot be picked up, or to disable task-local pick-ups altogether by passing `nil`

### Result:

A tracing library could now expose a metadata provider:

```swift
extension Logger.MetadataProvider {
    static let tracer = Logger.MetadataProvider { baggage in
        guard let spanContext = baggage?.spanContext else { return nil }
        return [
            "traceID": "\(spanContext.traceID)",
            "spanID": "\(spanContext.spanID)",
        ]
    }
}
```

Users of that library could instruct `LoggingSystem` to always use this metadata provider:

```swift
LoggingSystem.bootstrap(metadataProvider: .tracer, StreamLogHandler.standardOutput)
```

With this simple change, log statements would automatically include contextual metadata such as trace and span IDs if a task-local Baggage was set:

```swift
let logger = Logger(label: "org.swift")
logger.info("Request received.")
// info [traceID: 123, spanID: 456] Request received.
```

For more examples, in-depth explanations, and alternatives considered please read the proposal [proposals/0001-metadata-providers.md](https://github.com/apple/swift-log/blob/921cb5344cf0e59b9fdb963c2003b564d51e9932/proposals/0001-metadata-providers.md) attached to this PR.


ktoso:
As we discuss this, and verify it all is working as expected we'll also be announcing a 1.0 of distributed tracing along side with this. So swift-log would be depending on the stable 1.0 of baggage of course.

Package.swift
```diff
@@ -22,11 +22,14 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-distributed-tracing-baggage", from: "0.3.0"),
```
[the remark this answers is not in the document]
ktoso: Yes:

> So swift-log would be depending on the stable 1.0 of baggage of course.


https://github.com/apple/swift-log/pull/235#issuecomment-1263399324

Basically if this approach looks good we’ll do a final double check on APIs, make sure baggage can support 5.0 (today it does 5.2) and start 1.0-ing both baggage and tracing right after.

ktoso:
Cory also added a comment that I can't find inline now for some reason:

> We can't add a new parameter like this unless we're planning to emit a new semver major.

Right, @slashmo -- we need to keep source compatibility here, so we can't add a parameter to existing methods but need to add entirely new method implementations for:

```diff
     public func trace(_ message: @autoclosure () -> Logger.Message,
                       metadata: @autoclosure () -> Logger.Metadata? = nil,
+                      baggage: Baggage?,
```

in such way that the old implementations:

```
     public func trace(_ message: @autoclosure () -> Logger.Message,
                       metadata: @autoclosure () -> Logger.Metadata? = nil, ...) {
         self.trace(message, metadata: metadata, baggage: nil, ...)
```

in order to keep source compatibility.

There's also some bugs in Swift 5.0 it seems where the `#if` guards don't prevent it from trying (and failing to) parse task local use... So while we definitely should keep source compat, I'm not sure if we can manage to keep 5.0 around (5.1 works fine)

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+logger.info("... now this", metadata: ["trace-id": MyTracingLibrary.currentTraceID])
+```
+
+This comes with a couple of downsides:
+
+### Error-prone and repetetive
+
+It's easy to forget passing this metadata to _all_ log statements, resulting in an inconsistent debugging experience as these log statements cannot be found using correlation IDs.
+
+The repetetiveness and verboseness of logging multiple metadata in-line quickly becomes annoying and vastly decreases the signal-to-noise ratio of Swift code trying to be a good citizen and making use of log correlation techniques such as distributed tracing.
```
yim-lee, line 28:
```suggestion
The repetitiveness and verboseness of logging multiple metadata in-line quickly becomes annoying and vastly decreases the signal-to-noise ratio of Swift code trying to be a good citizen and making use of log correlation techniques such as distributed tracing.
```

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+
+### Proposal: Using `Baggage` as the execution context dependent metadata container
+
+We propose that swift-log should use [Swift Distributed Tracing Baggage](https://github.com/apple/swift-distributed-tracing-baggage) for its contextual metadata propagation type, meaning that swift-log would depend on this package.
+
+The `Baggage` type is modeled after the [Propagation format for distributed context: Baggage](https://www.w3.org/TR/baggage/), a W3C working draft, and is intended for all kinds of instrumentation of Swift applications both within and across processes. It is also used by [Swift Distributed Tracing](https://github.com/apple/swift-distributed-tracing), however it is usable even without tracing infrastructure. The baggage package contains the single `Baggage` type, and is expected to have a stable, unchanging API, fit for swift-log to depend upon.
+
+> NOTE: We could potentially even rename the baggage package, if the tracing wording in it is causing confusion here. But realistically, this is its primary purpose: to enable tracing in logging systems, so we think swift-log depending on `swift-distributed-tracing-baggage` is fine.
```
yim-lee, line 155: For me, the word "distributed" causes more confusion since logging doesn't necessarily have anything to do with "distributed".

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+}
+```
+
+Passing `nil` as the `Baggage` will prevent task-local look-ups in these performance sensitive situations.
+
+> NOTE: Most applications shouldn't need to worry about this and just rely on automatic Baggage propagation using task-locals
+
+> WARNING: The context-dependent metadata becomes static which makes it hard to recover the dynamic information back into `Baggage`, e.g. if you wanted to create a child span in one of those callbacks.
+
+## Alternatives considered
```
yim-lee, line 419: Have you considered defining an abstraction for baggage (without actually using `Baggage`) in swift-log and then have swift-distributed-tracing provide a concrete implementation?

ktoso: that would mean we need another package, to define some "abstract baggage like thing" which then both depend on (somewhat painful, baggage _is_ the "currency" for these things already)... and we'd end up using existentials in swift-log which is likely a bad idea (for performance reasons).

If it's yet again about baggage naming... perhaps we should reconsider the package name, but that already was through so many iterations of naming hell that I'm trying to be careful with that... I don't think we can do "swift-baggage" for example...

ktoso: Though perhaps another option would be to make providers:

```
() -> Metadata 
```

and every provider has to obtain the baggage _inside it_.

Upside: 
 - log does not have to depend on baggage for metadata providers

Downsides:
 - it becomes impossible to manually propagate baggage and libraries _must_ always wrap all code that should set have a trace-id in logs in a task local's `withValue` which has a performance impact -- so the high performance use-cases are hurt by this
 - if multiple metadata providers are used, it is likely they are going to be repeatedly looking up the same baggage task local... in some situations that's fine, in others perhaps somewhat painful (again, for performance reasons)
   - this is possible to work around by carefully designing libraries that offer metadata providers to also accept a `(Baggage) -> Metadata` but that's up to library authors then...

tomerd: generally speaking, I think it is okay to set a somewhat higher bar for library authors. ,and especially authors of core libraries such as async-http-client. would love to understand the exact performance issue with a closure / auto-closure since adding a dependency to swift-log seems like a fairly serious implication.

tomerd: cc @weissi

ktoso: What this design is try to mitigate is tons of thread-local accesses,  since every task-local value "pick up" is a thread-local hit, first and foremost. So if there's multiple hits to the same "contextual metadata container" (the baggage), per single log statement that's adding up. If there's one that's just life -- that's how locals work and we're only ever picking them up when _actually_ going to emit a log statement -- where the formatting of strings and other things absolutely dominate the pickup anyway.

The other reason to have a way to manually pass the baggage, and therefore _must_ have the type known to the library, is that libraries like NIO literarily can't rely on task-locals properly, since the same thread is shared across other callbacks/handlers, and just setting a scoped local works perfectly within swift concurrency, just isn't the right thing in a NIO system -- and picking up the local will also never be right; so, there has to be a way to pass into the same "extract values to log them" using the container manually.

ktoso: As far as benchmarking, we kicked off some initial benchmarks here: https://github.com/slashmo/swift-log/tree/feature/baggage-benchmarks though we need to make sure we're measuring the right things there, I'll look at it before our call tonight a bit more.

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+Most notably, loggers should participate in tracing by including some trace metadata (such as e.g. a `trace-id`) when logging, as it transparently enables developers to benefit from log correlation using those IDs.
+
+Today, the only supported way of providing such metadata is to pass them along to each log call explicitly:
+
+```swift
+logger.info("first this ...", metadata: ["trace-id": MyTracingLibrary.currentTraceID])
+logger.info("... now this", metadata: ["trace-id": MyTracingLibrary.currentTraceID])
+```
+
+This comes with a couple of downsides:
```
FranzBusch, line 22: Just as a general note I think this is not the only supported way. The other way, which is also the one that I see more often, is copying the `Logger` once you decoded the `trace-id` and from that point on pass that `Logger` forward.

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+
+### Error-prone and repetitive
+
+It's easy to forget passing this metadata to _all_ log statements, resulting in an inconsistent debugging experience as these log statements cannot be found using correlation IDs.
+
+The repetitiveness and verboseness of logging multiple metadata in-line quickly becomes annoying and vastly decreases the signal-to-noise ratio of Swift code trying to be a good citizen and making use of log correlation techniques such as distributed tracing.
```
FranzBusch, line 28: Similar point as the above as long as the libraries make sure to set the correct metadata on the logger and then pass it along it is not as error prone.

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+Metadata keys are unique, so in case multiple metadata providers return the same key,
+the last provider in the array "wins" and provides the value.
```
FranzBusch, line 147: Should we do this silently here? I think some of our `LogHandler`s actually precondition or fatalError on duplicates to avoid silent duplicates that override something.

proposals/0001-metadata-providers.md
FranzBusch: I agree. With this proposal the primary purpose of the package somehow changes since now it is not only used for tracing but also for providing log metadata through a task local. Can we imagine other use-cases where we want to use the baggage?

Just thinking out loud here what about `swift-task-baggage`? Though that might be a bit to abstract.

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+    logger.info("Such logging, wow!", metadata: ["one-off": "example"])
+    // info [my-test-key: my-awesome-trace-id, one-off: example] Such logging, wow!
+}
+```
+
+This has the advantage of not having to specify the global metadata explicitly anymore.
+
+The task-local baggage is often going to be set by e.g. tracing libraries or frameworks, so end-users do not need to do anything to benefit from tracing,
+other than setting up the metadata providers when they bootstrap their logging and tracing systems. For example, assuming some imaginary http framework where http requests are forwarded to an actor for handling, all the end-users of this framework would benefit from contextual metadata, propagated with baggage by logging, as usual:
+
```
FranzBusch, line 263: Just for my full understanding. Let's take the example of a gRPC-server. What we expect the gRPC library folks to implement is once they decode the trace-id from the incoming requests headers to set them on the current task local baggage right?
What key would they use to set on the `Baggage`? Is that key coming from `swift-distributed-tracing`?

proposals/0001-metadata-providers.md
tomerd: if we end up deciding we _must_ add an additional container type, I think we should pause and consider where it is best"located". swift-log is designed to be a low level, zero dependencies library and is used across many use cases that do not include distributed tracing, which triggered the above reaction. as such, if we _must_ add such container type, maybe there is a lower level where we can introduce this like the standard library as this is potentially useful outside this immediate context / in other context propagation scenarios. alternatively, we can explore reversing the dependency and add this type to swift-log and have the higher level distributed tracing libraries depend on swift-log. these are just ideas, the point is that we need to explore all options

proposals/0001-metadata-providers.md
slashmo: Good point, maybe the example lacks a bit. In practice it wouldn't just be about passing around a trace ID, but also a span ID. A span ID can't be defined wherever you create the `Logger` which you pass around, because it's changing during the lifetime of a single request.

proposals/0001-metadata-providers.md
slashmo: This approach works well with more static metadata, but not with dynamically changing information like a span ID.

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+```swift
+let logger = Logger(label: "example", metadataProvider: .init { baggage in
+    guard let operationID = baggage?.operationID else { return nil }
+    return ["extra/opID": "\(opID)"]
+})
+```
+
+> NOTE: Setting the metadata provider on the logger directly means the `LoggingSystem` metadata provider
+> is skipped (if defined), following how an explicitly passed handler `factory`
+> overrides the `LoggingSystem`s `factory`.
```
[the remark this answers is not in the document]
ktoso: Hm yeah that'd be an option if we want; it'd be `addMetadataProvider` rather than "`set`" I guess. That's probably better for most use-cases I suppose.

Sources/Logging/Logging.swift
```diff
@@ -70,16 +75,34 @@ extension Logger {
...
     ///            defaults to `#line`).
+    #if compiler(>=5.3)
     @inlinable
     public func log(level: Logger.Level,
                     _ message: @autoclosure () -> Logger.Message,
                     metadata: @autoclosure () -> Logger.Metadata? = nil,
                     source: @autoclosure () -> String? = nil,
                     file: String = #fileID, function: String = #function, line: UInt = #line) {
         if self.logLevel <= level {
+            let taskLocalBaggage: Baggage?
```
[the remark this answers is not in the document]
ktoso: Thanks for asking in the thread here; I did some benchmarking but the numbers were so all over the place that I didn't trust them yet. I'll follow up with more measurements soon, thanks!

ktoso:
A few TODOs I took notes for myself just so I have it all in the same thread:
- make the "dont pick up baggage" an explicit API, rather than "don't pass a value" so it is possible to write configurable wrappers
- get back with benchmarks
- re-consider where baggage can live
- write a server example

proposals/0001-metadata-providers.md
ktoso: I'll add more examples for the server side, tho tbh that perhaps is time better spent on revamping the tracing "what you should do" guide and I'll link to it from here rather in this proposal itself. I'll add more examples tho, good question - thanks.

Baggage has _ONE_ "blessed" task-local key and it is the `Baggage.current`; that's what an instrumented system would be setting.

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+    func channelActive(context: ChannelHandlerContext) {
+        logger.debug("Channel active.", baggage: self.baggage)
+        // debug [trace-id: 42] Channel active.
+    }
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        logger.debug("Received request part.", baggage: self.baggage)
+        // debug [trace-id: 42] Received request part.
+    }
+}
```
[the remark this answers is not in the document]
ktoso: I'm not sure I understand your examples here, we can catch up a bit on this specific question perhaps?

proposals/0001-metadata-providers.md
ktoso: Personally I'd be _very_ fine with that. To focus on a package that is a type good for "carrying around" whatever you want, and the existing infrastructure to propagate baggage -- regardless what is contained in it makes a lot of sense; callbacky frameworks picking it up and restoring it.

We had quite some naming drama bout the package last time, but I very much would like to make sure we all like what we end up with -- so that's on the table. Another idea was by @tomerd to sink it into swift-log, which would work, but I'm a bit worried about marrying it to logging _only_... Let's weigh our options here -- certianly the name you propose @FranzBusch would be quite good!

ktoso:
Another approach we need to consider: what if we made only log handlers aware of baggage -- though I believe it'll have "interesting" tradeoffs 🤔

tomerd:
> what if we made only log handlers aware of baggage 

the obvious upside is that the logging API stays narrowly focused on its main goal. the other upside is that the handler can make it a configurable option, which feel like the natural place. the downside is that its less DRY ie all handlers (collectors) that are interested in this would need to implement it. IMO, given that we dont expects to have tons of collectors it probably the right tradeoff

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,489 @@
...
+
+```swift
+class PerformanceCriticalRequestStateMachine: ChannelInboundHandler {
+    let logger: Logger
+
+    init(logger: Logger) {
+        var logger = logger
+        // in here we can still access the task-local Baggage
+        if let baggage = Baggage.current, let metadataProvider = logger.metadataProvider {
+            let metadata = metadataProvider(baggage)
```
adam-fowler, line 393: A helper function that does this for you would be nice

adam-fowler (commented):
Finally got around to looking at this. Not much new to add on top of what has already been said. 

Has combining `swift-distributed-tracing-baggage` and `swift-log` been considered. `swift-distributed-tracing-baggage` is fairly small and having the ability to add metadata via Task locals is a useful feature. I guess the question is then do you want `swift-distributed-tracing` to be dependent on `swift-log` instead of the smaller `swift-distributed-tracing-baggage` package.

ktoso:
I pushed my revision which pushes baggage pickup and printing et al into `LogHandler`: https://github.com/apple/swift-log/pull/237

We keep the baggage "pick up strategy" in API though, in order to support usage that AHC has requested a while ago, to be able to manually pass when necessary.

Mordil:
I know I’m coming really late to the party with lots of other discussions already have happened, but as a library author looking to really assist all use-cases I'm trying to find the easiest way forward.

I’ve been looking at the pattern and philosophy @ktoso outlined in this [forum post about “logs vs span events”](https://forums.swift.org/t/should-libraries-use-logs-or-span-events/60586/2), and I wonder if there’s a natural hierarchy of packages to formally support the proposed pattern more easily across the ecosystem:

```
- swift-instrumentation (swift-distributed-tracing-baggage-core or swift-distributed-tracing-baggage)
  - swift-metrics
  - swift-log
  - swift-tracing
```

with the **Instrumentation** providing `Baggage`  as a standard way of propagating metadata, `BaggageStrategy` (as proposed in #237), and the concept of an Instrumentation System as outlined partially by Konrad's post, but also borrowing some of Metrics/Logging/Tracing's common API which each of those packages will then plug into

Then vendors such as New Relic or Sentry could pull in all 3 (**Tracing**, **Logging**, **Metrics**) and make their system work within those protocols & frameworks to give app developers the freedom and expressiveness of the individual observability features.

And library authors only need to rely on **Instrumentation**, with their own unique spin on how it's supported (a library specific Instrumentation delegate)

ktoso:
Thanks @Mordil for chiming in -- having instrumentation baggage the root of the observability dependencies indeed would be quite right, though for better or worse we've not gotten agreement on that; and we had to prioritize keeping swift-log zero dependency.

I completed the third rewrite of this all, and updated the proposal as well as all tests are green: https://github.com/apple/swift-log/pull/238

The latest attempt is what I believe can be the final one here, please have a look over there.

PR #236, opened 2022-10-30
Fix build for wasm

ahti:
Adds necessary \#ifs to fix the WASM build.

### Motivation:

Seems like a few places were missed in the locking refactor, which made the WASM build fail due to missing pthread types and unknown `rwlock` variable.

### Modifications:

\#if-ed out remaining mentions of pthread types and variables not available on WASM.

### Result:

`swift-log` is usable on WASM again.

### DIFF
--- a/Sources/Logging/Locks.swift
+++ b/Sources/Logging/Locks.swift
@@ -45,7 +45,9 @@ import Glibc
 /// one used by NIO. On Windows, the lock is based on the substantially similar
 /// `SRWLOCK` type.
 internal final class Lock {
-    #if os(Windows)
+    #if canImport(WASILibc)
+    // WASILibc is single threaded, provides no locks
+    #elseif os(Windows)
     fileprivate let mutex: UnsafeMutablePointer<SRWLOCK> =
         UnsafeMutablePointer.allocate(capacity: 1)
     #else
@@ -55,7 +57,9 @@ internal final class Lock {
 
     /// Create a new lock.
     public init() {
-        #if os(Windows)
+        #if canImport(WASILibc)
+        // WASILibc is single threaded, provides no locks
+        #elseif os(Windows)
         InitializeSRWLock(self.mutex)
         #else
         var attr = pthread_mutexattr_t()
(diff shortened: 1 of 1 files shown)


Sources/Logging/Locks.swift
```diff
@@ -168,7 +181,10 @@ internal final class ReadWriteLock {
         let err = pthread_rwlock_destroy(self.rwlock)
         precondition(err == 0, "\(#function) failed in pthread_rwlock with error \(err)")
         #endif
+
+        #if !canImport(WASILibc)
         self.rwlock.deallocate()
+        #endif
```
tomerd, line 187: should `self.rwlock.deallocate()` be moved into the condition instead?

Sources/Logging/Locks.swift
```diff
@@ -68,21 +72,28 @@ internal final class Lock {
...
+        #elseif os(Windows)
         // SRWLOCK does not need to be free'd
         #else
         let err = pthread_mutex_destroy(self.mutex)
         precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
         #endif
+
+        #if !canImport(WASILibc)
         self.mutex.deallocate()
+        #endif
```
tomerd, line 86: should `self.mutex.deallocate()` be moved into the condition instead?

or alternatively change deallocate to handle WASI? I assume that is what is happening on Windows?

tomerd:

Sources/Logging/Locks.swift
ahti: Unless I'm mistaken, deallocate is from UnsafeMutablePointer, so I don't think that can be changed here. I put this in its own condition to avoid the duplication, but I can move it into the Windows and default cases above later today. Same for rwlock further down.

tomerd: oh I see. yes I think its best to be symmetrical and deallocate inside the same branches as allocating

ahti: okay, I updated the PR accordingly

tomerd:

tomerd (approved):

furby-tm:
@ahti Nice!! I'm happy to see that fix landed! Out of curiosity, I thought that if you were to `import WASILibc` that there was some support for pthreads, or, perhaps those were just stubbed for now? 

Somehow I vaguely remember seeing a commit for the removal of the stubs, but it's also entirely possible I'm just confusing myself with something else.

In either event, it's great to see all this progress! Thanks!

PR #237, opened 2022-11-14
RFC (rev 2): Metadata Providers & Distributed Tracing Support in swift-log

ktoso:
This builds on top of https://github.com/apple/swift-log/pull/235 but moves the `Baggage` handling into `LogHandler`s.

> NOTE: This CAN be fully source compatible; though I didn't fully do the copy-pasting of some declarations yet.

The API still does offer the ability to pass a baggage, or DISABLE its picking up, because of requirements stated earlier by AsyncHTTPClient. We should discuss this approach in parallel to  https://github.com/apple/swift-log/pull/235

Specifically, this is how a handler would be invoked (and yes, this is backwards compatible, the same way we added new parameters to handlers in previous iterations):

```
                     baggage: ContextualBaggage = .taskLocal,
         if self.logLevel <= level {
             self.handler.log(level: level,
                              message: message(),
                              metadata: metadata(),
                              baggage: baggage,
                              source: source() ?? Logger.currentModule(filePath: (file)),
                              file: file, function: function, line: line)
         }
```

where the modes are: 

```swift
    /// Defines where contextual `Baggage` metadata should be obtained from.
     /// Generally, the default of the ``ContextualBaggage/taskLocal`` strategy is the best to use in most applications.
     /// However, some systems may not be able to use task-locals, or want to pass baggage values explicitly.
     public enum BaggageStrategy {
         /// Instructs the ``LogHandler`` to pick-up the task-local `Baggage`,
         /// by querying `Baggage/current` when about to emit a log statement.
         case taskLocal

         /// Instructs the logger backend to use the explicitly passed Baggage when extracting metadata to include
         // in log statements.
         case explicitlyPassed(InstrumentationBaggage.Baggage)

         /// Completely ignore contextual Baggage, and do not attempt to pick-up,
         /// or extract values from a passed Baggage value.
         ///
         /// Effectively, this disables any configured ``Logger/MetadataProvider``.
         case ignore
     }
```

so this allows us to either turn off, pass explicitly, or keep the default "pick up from task-local".

Every log handler needs to be updated to do this, however if they want to, they can avoid "merging" dictionaries which can be more efficient.


ktoso:
Replaced by another redesign: https://github.com/apple/swift-log/pull/238

PR #238, opened 2022-11-24
Metadata Providers (e.g. for Distributed Tracing) in LogHandlers

ktoso:
This is the third, and hopefully final, revision of `MetadataProvider`. I believe this achieves everything we could come up with in the multiple rounds of reviews and calls, and introduces _zero_ dependencies or api breaks to the project.

Thank you to @slashmo, the NIO and SSWG teams for the continued discussions 👍 

This approach is summarized as:

- log handlers gain metadata providers
- providers, and whole of logging, remain unaware of `Baggage`
- a provider implementation, e.g. `metadataProvider: .otel` can and will pick up baggage from task-locals, since they do depend on baggage and tracing APIs.
- swift-distributed-tracing is able to implement `provideMetadata(from: Baggage)`

Updated proposal text: https://github.com/ktoso/swift-log/blob/wip-baggage-in-handlers-but-not-api/proposals/0001-metadata-providers.md

---

Distributed tracing is able to provide the following integration, in the `Instrumentation` module:

```swift
#if swift(>=5.5.0) && canImport(_Concurrency)
@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
extension Logger {

    /// Extract any potential ``Logger/Metadata`` that might be obtained by extracting the
    /// passed ``InstrumentationBaggage/Baggage`` to this logger's configured ``Logger/MetadataProvider``.
    ///
    /// Use this when it is necessary to "materialize" contextual baggage metadata into the logger, for future use,
    /// and you cannot rely on using the task-local way of passing Baggage around, e.g. when the logger will be
    /// used from multiple callbacks, and it would be troublesome to have to restore the task-local baggage in every callback.
    ///
    /// Generally prefer to set the task-local baggage using `Baggage.current` or `Tracer.withSpan`.
    public mutating func provideMetadata(from baggage: Baggage?) {
        guard let baggage = baggage else {
            return
        }

        Baggage.withValue(baggage) {
            let metadata = self.metadataProvider.provideMetadata()
            for (k, v) in metadata {
                self[metadataKey: k] = v
            }
        }
    }
}
#endif
```

which allows us to "capture all baggage into metadata ONCE" and then keep using this logger with the metadata; which was a feature request from NIO-heavy libraries.

--- 

The API break reported is:

```
22:45:48 /* Protocol Requirement Change */
22:45:48 Var LogHandler.metadataProvider has been added as a protocol requirement
22:45:48 
```

which is accounted for by a default implementation:

```swift
public extension LogHandler {
    /// Default implementation for `metadataProvider` which defaults to a "no-op" provider.
    var metadataProvider: Logger.MetadataProvider {
        get {
            return .noop
        }
        set {
            self.log(level: .warning, message: "Attempted to set metadataProvider on \(Self.self) that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.", metadata: nil, source: "Logging", file: #file, function: #function, line: #line)
        }
    }
}
```

and this will never be triggered _unless_ someone very actively in their end user library tries to make use of this, and then they'll get this warning. Existing libraries continue to work, but **SHOULD implement support for invoking the metadata providers** as otherwise the metadata will not be included in log statements. They can do so at their own pace though.

--- 

Since I wanted to include tests showing how task-locals are to be used with this, but Swift 5.0 cannot lex these properly even (confirmed this compiler bug with the compiler team; this will not be fixed), I had to make a new module for them.

This was quite painful, so either we keep this hacky separate module, or we drop those tests, or we drop 5.0 support. Either is fine with me.


Sources/Logging/LogHandler.swift
```diff
@@ -114,6 +114,12 @@
 /// level has not been overridden. And most importantly it passes the requirement listed above: A change to the log
 /// level on one `Logger` should not affect the log level of another `Logger` variable.
 public protocol LogHandler: _SwiftLogSendableLogHandler {
+    /// The metadata provider this `LogHandler` will use when a log statement is about to be emitted.
+    ///
+    /// A ``Logger/MetadataProvider`` may add a constant set of metadata,
+    /// or use task-local values to pick up contextual metadata ane add it to emitted logs.
```
t089, line 120:
```suggestion
    /// or use task-local values to pick up contextual metadata and add it to emitted logs.
```

ktoso: Thanks!

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -873,18 +920,20 @@ class LoggingTest: XCTestCase {
     }
 }
 
-extension Logger {
+public extension Logger {
```
ktoso, line 923: oh gosh mismatched format versions truly are fun... Going to fix all those

scripts/generate_linux_tests.rb
```diff
@@ -109,13 +116,26 @@ def createLinuxMain(testsDirectory, allTestSubDirectories, files)
...
 
+def testCaseNeedsIfGuard(testCase, allTestSubDirectories, major, minor)
+  for f in allTestSubDirectories
+    if File.exist? "Tests/" + f + "/" + testCase.split(/./, 1).first + ".swift" then
+      return f =~ /#{major}#{minor}/
+    end
+  end
+
+  false
+end
```
ktoso, line 137: This is horrible, though good enough to have the discussion if we want to make this workaround less horrible, remove those few tests entirely (and remove this hack), or what else.

FranzBusch (approved):
Just read through all of this again and I really like where we ended up now. It looks like an organic extension to swift-log which plays nicely with distributed tracing!
I left a few small comments inline but overall big +1!

W.r.r. the required hack, I would vote for increasing the min Swift version. NIO is setting precedent here already and IMO it should be fine to have the same rules for swift-log.

Sources/Logging/MetadataProvider.swift
```diff
@@ -0,0 +1,112 @@
...
+            return MetadataProvider { [:] }
+        }
+
+        /// Return the ``Logger/MetadataProvider-swift.struct`` that was configured during ``LoggingSystem/bootstrap(_:)-8ffrb``.
+        static var bootstrapped: MetadataProvider {
+            return LoggingSystem.metadataProvider
+        }
+
+        #if swift(>=5.5) && canImport(_Concurrency)
+        public typealias Function = @Sendable() -> Metadata
```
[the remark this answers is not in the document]
weissi: +1 against function typealiases in Swift, super hard to work with, especially with Xcode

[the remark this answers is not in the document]
ktoso: hmmm, I'll have to re-declare a bunch of entire function bodies unless I do a typealias like that -- so pretty painful tradeoff

Sources/Logging/MetadataProvider.swift
```diff
@@ -0,0 +1,112 @@
...
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(WASILibc)
+import WASILibc
+#else
+#error("Unsupported runtime")
+#endif
+
+#if compiler(>=5.6)
+@preconcurrency public protocol _SwiftLogSendable: Sendable {}
```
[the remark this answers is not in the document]
ktoso: Hmmm, I'll check but I think they need to be public -- it's for other libs to know this type is Sendable after all 🤔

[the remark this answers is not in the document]
FranzBusch: If I recall correctly this also works with an internal protocol. Maybe this changed but I tried to for one of our NIO repos. Somehow the compiler still transfers the Sendable constraint 😅

Sources/Logging/MetadataProvider.swift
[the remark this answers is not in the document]
FranzBusch: Yep painful but better end user experience 😅

Package.swift
```diff
@@ -32,5 +32,16 @@ let package = Package(
             name: "LoggingTests",
             dependencies: ["Logging"]
         ),
+        // Due to a compiler bug in parsing/lexing in 5.0,
```
yim-lee, line 35: I vote that we drop 5.0 support

ktoso: Yeah, honestly it's about time. We can go careful and _just_ drop 5.0, the others we can keep around still.

Sources/Logging/MetadataProvider.swift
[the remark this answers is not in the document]
ktoso: Okey I'll do that but can you explain what the issue is? I don't use Xcode and it's a bit baffling to me why a typealias can be hard to work with...?

Sources/Logging/Locks.swift
```diff
@@ -125,7 +125,7 @@ extension Lock {
     /// - Parameter body: The block to execute while holding the lock.
     /// - Returns: The value returned by the block.
     @inlinable
-    internal func withLock<T>(_ body: () throws -> T) rethrows -> T {
+    func withLock<T>(_ body: () throws -> T) rethrows -> T {
```
tomerd, line 128: could we undo (or move to another PR) these formatting changes? makes the diff too large and harder to review

Tests/LoggingTests/TestLogger.swift
```diff
@@ -23,7 +23,21 @@ internal struct TestLogging {
...
-        return TestLogHandler(label: label, config: self.config, recorder: self.recorder)
+        return TestLogHandler(
+            label: label,
+            config: self.config,
+            recorder: self.recorder,
+            metadataProvider: LoggingSystem.metadataProvider
+        )
+    }
+
+    func makeWithMetadataProvider(label: String, metadataProvider: Logger.MetadataProvider) -> LogHandler {
```
tomerd, line 34: collapse this into `make` and make `metadataProvider` optional with default of `.none`?

scripts/generate_linux_tests.rb
tomerd: what is this working around?

Sources/Logging/Logging.swift
```diff
@@ -43,13 +43,19 @@ public struct Logger {
     /// An identifier of the creator of this `Logger`.
     public let label: String
 
+    /// The metadata provider this logger was created with.
+    /// If no provider was configured, this will return a ``Logger/MetadataProvider/noop`` no-op metadata provider.
+    public var metadataProvider: Logger.MetadataProvider {
```
tomerd, line 48: does this need to be public?

Sources/Logging/Logging.swift
```diff
@@ -653,55 +661,165 @@ extension Logger {
...
+    // for our testing we want to allow multiple bootstrapping
     internal static func bootstrapInternal(_ factory: @escaping (String) -> LogHandler) {
+        self._metadataProviderFactory.replaceMetadataProvider(.noop, validate: false) // TODO: fixme remove this and fix tests
+        self._factory.replaceFactory({ label, _ in
+            factory(label)
+        }, validate: false)
+    }
+
+    // for our testing we want to allow multiple bootstraping
+    internal static func bootstrapInternal(_ factory: @escaping (String, Logger.MetadataProvider) -> LogHandler) {
```
tomerd, line 713: bootstrapInternalMetadataProvider? or maybe collapse the two internal bootstrap functions to one, with metadata provider optional?

Sources/Logging/Logging.swift
```diff
@@ -653,55 +661,165 @@ extension Logger {
...
         self._factory.replaceFactory(factory, validate: false)
     }
 
-    fileprivate static var factory: (String) -> LogHandler {
-        return self._factory.underlying
+    internal static func bootstrapInternal(metadataProvider: Logger.MetadataProvider) {
```
tomerd, line 717: bootstrapInternalMetadataProvider? or maybe collapse the two internal bootstrap functions to one, with metadata provider optional?

Sources/Logging/Logging.swift
```diff
@@ -653,55 +661,165 @@ extension Logger {
...
+        self._metadataProviderFactory.replaceMetadataProvider(metadataProvider, validate: false)
+    }
+
+    internal static func bootstrapInternal(metadataProvider: Logger.MetadataProvider,
```
tomerd, line 721: bootstrapInternalMetadataProvider? or maybe collapse the two internal bootstrap functions to one, with metadata provider optional?

Sources/Logging/Logging.swift
```diff
@@ -653,55 +661,165 @@ extension Logger {
...
+
+    /// System wide ``Logger/MetadataProvider`` that was configured during the logging system's `bootstrap`.
+    internal static var metadataProvider: Logger.MetadataProvider {
+        return self._metadataProviderFactory.metadataProvider
     }
 
     private final class FactoryBox {
         private let lock = ReadWriteLock()
-        fileprivate var _underlying: (String) -> LogHandler
+        fileprivate var _underlying: (String, Logger.MetadataProvider) -> LogHandler
```
tomerd, line 742: nit: add labels

Sources/Logging/Logging.swift
```diff
@@ -653,55 +661,165 @@ extension Logger {
...
-extension Logger {
+public extension LoggingSystem {
+    /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
+    /// implementation. `bootstrap` can be called at maximum once in any given program, calling it more than once will
+    /// lead to undefined behavior, most likely a crash.
+    ///
+    /// - parameters:
+    ///     - metadataProvider: The `MetadataProvider` used to inject runtime-generated metadata, defaults to a "no-op" provider.
+    ///     - factory: A closure that given a `Logger` identifier, produces an instance of the `LogHandler`.
+    static func bootstrap(metadataProvider: Logger.MetadataProvider,
```
tomerd, line 798: feels like the metadataProvider should be the second argument

Sources/Logging/Logging.swift
```diff
@@ -800,13 +918,51 @@ extension Logger {
...
+    /// - parameters:
+    ///     - label: An identifier for the creator of a `Logger`.
+    ///     - factory: A closure creating non-standard `LogHandler`s.
+    init(label: String, metadataProvider: MetadataProvider) {
+        self = Logger(label: label, factory: { label in
+            LoggingSystem.factory(label, metadataProvider)
+        })
+    }
+
+    init(label: String, metadataProvider: @escaping MetadataProvider.Function) {
```
tomerd, line 942: could auto-closure be useful here?

Sources/Logging/Logging.swift
```diff
@@ -1120,39 +1283,68 @@ public struct StreamLogHandler: LogHandler {
...
                     source: String,
                     file: String,
                     function: String,
                     line: UInt) {
-        let prettyMetadata = metadata?.isEmpty ?? true
-            ? self.prettyMetadata
-            : self.prettify(self.metadata.merging(metadata!, uniquingKeysWith: { _, new in new }))
+        let effectiveMetadata = StreamLogHandler.prepareMetadata(base: self.metadata, provider: self.metadataProvider, explicit: explicitMetadata)
+
+        let prettyMetadata: String?
```
tomerd, line 1301: `String!`?

Sources/Logging/Logging.swift
ktoso: Yes, this is public on purpose in order to facilitate external libraries, such as tracing, to implement the "provideMetadata" functions, like this:

```
#if swift(>=5.5.0) && canImport(_Concurrency)
@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
extension Logger {

    /// Extract any potential ``Logger/Metadata`` that might be obtained by extracting the
    /// passed ``InstrumentationBaggage/Baggage`` to this logger's configured ``Logger/MetadataProvider``.
    ///
    /// Use this when it is necessary to "materialize" contextual baggage metadata into the logger, for future use,
    /// and you cannot rely on using the task-local way of passing Baggage around, e.g. when the logger will be
    /// used from multiple callbacks, and it would be troublesome to have to restore the task-local baggage in every callback.
    ///
    /// Generally prefer to set the task-local baggage using `Baggage.current` or `Tracer.withSpan`.
    public mutating func provideMetadata(from baggage: Baggage?) {
        guard let baggage = baggage else {
            return
        }

        Baggage.withValue(baggage) {
            let metadata = self.metadataProvider.provideMetadata()
            for (k, v) in metadata {
                self[metadataKey: k] = v
            }
        }
    }
}
#endif
```

Tests/LoggingTests/TestLogger.swift
ktoso: That'll cause a series of ambiguity errors AFAIR, I'll give it a shot tho.

scripts/generate_linux_tests.rb
ktoso: A bug in Swift 5.0 that it cannot parse/lex code that has the shape of "Thing.$thing", which task-locals have to use - even when surrounded with `#if swift(>=5.1)` swift 5.0 will still fail to compile such code (yeah, it should ignore code in that #if, but it doesn't).

Options have are:
- don't have any tests which show how a task-local can be used with providers (we could not have those tests... and just static providers everywhere in our tests, though it's a bit annoying)
  - (no actual code is 5.0 breaking actually in this PR, it is only the tests)
- or use this separate module workaround

We can still have testing of this pattern in the tracing module, but it's a bit weird to test metadata provider patterns in a different package than the one declaring them.

Sources/Logging/Locks.swift
ktoso: I'll undo them; I have to stop using a pre-commit formatting hook to be honest; different versions of swift format, and projects not declaring all options which they rely on explicitly cause this :-(

Sources/Logging/Logging.swift
ktoso: A while ago we came to the order of parameters recommendations around "first strict, then closures, then optional params, then optional closures" so this follows that recommendation.

I'm not sure how many people actually write 

```
LoggingSystem.bootstrap {
  many
  lines
  return here
}
```

so it was intended to keep that shape 🤔 

```
LoggingSystem.bootstrap(metadataProvider: .coolTracer) {
  many
  lines
  return here
}
```

if we flip it, we end up with:

```
LoggingSystem.bootstrap({
  many
  lines
  return here
}, metadataProvider: .coolTracer)
```

potentially... but arguably most uses do look like this:

```
LoggingSystem.bootstrap(.coolLogger, metadataProvider: .coolTracer)
```

potentially... so, I'm not really sure about this one. The current shape kind of follows the "usual" order, with closures last. Open to either but current seemed more natural given the usage shapes hm

Sources/Logging/Logging.swift
ktoso: Oh good catch, just `String` perhaps

Sources/Logging/Logging.swift
```diff
@@ -43,13 +43,19 @@ public struct Logger {
...
+    }
+
     internal init(label: String, _ handler: LogHandler) {
         self.label = label
         self.handler = handler
     }
 }
 
-extension Logger {
+public extension Logger {
```
yim-lee, line 58: FYI, we can change SwiftFormat's behavior with `--extensionacl on-declarations`

ktoso: I'll add it to `.swiftformat`, that';s what we want 👏

Sources/Logging/Logging.swift
```diff
@@ -800,13 +918,51 @@ extension Logger {
...
+    ///
+    /// The `label` should identify the creator of the `Logger`. This can be an application, a sub-system, or even
+    /// a datatype.
+    ///
+    /// This initializer provides an escape hatch in case the global default logging backend implementation (set up
+    /// using `LoggingSystem.bootstrap` is not appropriate for this particular logger.
+    ///
+    /// - parameters:
+    ///     - label: An identifier for the creator of a `Logger`.
+    ///     - factory: A closure creating non-standard `LogHandler`s.
```
yim-lee, line 935: This description doesn't look right

Sources/Logging/Logging.swift
```diff
@@ -800,13 +918,51 @@ extension Logger {
...
```
yim-lee, line 958:
```suggestion
    ///     - factoryWithMetadataProvider: A closure creating non-standard `LogHandler`s.
```

Sources/Logging/Logging.swift
```diff
@@ -1081,6 +1237,11 @@ let systemStdout = WASILibc.stdout!
 
 /// `StreamLogHandler` is a simple implementation of `LogHandler` for directing
 /// `Logger` output to either `stderr` or `stdout` via the factory methods.
+///
+/// Metadata is merged in the following order:
+/// - metadata set on the log handler itself is used as the base metadata,
+/// - next, the handlers ``metadataProvider`` is invoked, and overrides any existing keys,
+/// - next, the per-log-statement metadata is merged, overriding any previously set keys.
```
yim-lee, line 1244:
```suggestion
/// 1. Metadata set on the log handler itself is used as the base metadata.
/// 2. The handler's ``metadataProvider`` is invoked, overriding any existing keys.
/// 3. The per-log-statement metadata is merged, overriding any previously set keys.
```

Sources/Logging/Logging.swift
ktoso: Heh so that's part of the problem; the version we use is too old to have this flag;
Other sswg projects use newer formatters, so it is hard to keep it all aligned.

We should bump the formatter version everywhere and fix this mess this way, and by adding more flags. I'll undo bad formats now and avoid reformatting automatically heh.

Sources/Logging/Logging.swift
ktoso: Ok, though they'll be pretty boring in the sense that not usable at call sites: `fileprivate var _underlying: (_ label: String, _ provider: Logger.MetadataProvider)`; swift does not allow for `_underlying: (label: String, provider: Logger.MetadataProvider)` :-/

Sources/Logging/Logging.swift
ktoso: hmm, yeah we could I guess; it'd allow a simple spelling for a "constant metadata provider" that is just the dictionary

Tests/LoggingTests/TestLogger.swift
ktoso: Yeah I'll push back on this one, it causes a lot more pain in our test code than necessary; 
We can't do `LoggingSystem.bootstrapInternal(logging.makeWithMetadataProvider)` style anymore and would have to spell out all the closures to disambiguate.

Note that this is only our test code, so the verbose name is not a concern IMHO

Sources/Logging/Logging.swift
```diff
@@ -255,7 +261,6 @@ extension Logger {
     ///    - line: The line this log message originates from (there's usually no need to pass it explicitly as it
     ///            defaults to `#line`).
     #if compiler(>=5.3)
-    @inlinable
```
glbrntt, line 258: Why did this `@inlinable` get removed?

Sources/Logging/Logging.swift
```diff
@@ -653,50 +659,160 @@ extension Logger {
...
-    private static let _factory = FactoryBox(StreamLogHandler.standardOutput)
+    private static let _factory = FactoryBox { label, _ in StreamLogHandler.standardOutput(label: label) }
+    private static let _metadataProviderFactory = MetadataProviderBox(.noop)
 
     /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
     /// implementation. `bootstrap` can be called at maximum once in any given program, calling it more than once will
     /// lead to undefined behavior, most likely a crash.
     ///
     /// - parameters:
+    ///     - metadataProvider: The `MetadataProvider` used to inject runtime-generated metadata, defaults to nil.
```
glbrntt, line 670: Should this be "defaults to a no-op provider" or something like that?

Sources/Logging/Logging.swift
ktoso: ticketified https://github.com/apple/swift-log/issues/239

Sources/Logging/Logging.swift
ktoso: (done)

Sources/Logging/Logging.swift
ktoso: The optional actually works out well here since the self.prettyMetadata may be optional so this is a nice way to handle it. Keeping as is

Sources/Logging/Logging.swift
```diff
@@ -1120,39 +1292,68 @@ public struct StreamLogHandler: LogHandler {
...
+
+        let provided = provider.provideMetadata()
+
+        guard !provided.isEmpty || !((explicit ?? [:]).isEmpty) else {
+            // all per-log-statement values are empty
+            return nil
+        }
+
+        if !provided.isEmpty {
+            metadata = metadata.merging(provided, uniquingKeysWith: { _, provided in provided })
```
glbrntt, line 1332: Probably better to use `merge` here (and below) to avoid creating a new intermediate dictionary

Sources/Logging/MetadataProvider.swift
```diff
@@ -0,0 +1,112 @@
...
+    ///
+    /// ### Merging conflicting keys
+    ///
+    /// `MetadataProvider`s are invoked left to right in the order specified in the `providers` argument.
+    /// In case multiple providers try to add a value for the same key, the last provider "wins" and its value is being used.
+    ///
+    /// - Parameter providers: An array of `MetadataProvider`s to delegate to. The array must not be empty.
+    /// - Returns: A pseudo-`MetadataProvider` merging metadata from the given `MetadataProvider`s.
+    public static func multiplex(_ providers: [Logger.MetadataProvider]) -> Logger.MetadataProvider {
+        precondition(!providers.isEmpty, "providers MUST NOT be empty")
```
glbrntt, line 101: Any reason to not just return `.noop` in this case?

Sources/Logging/Logging.swift
ktoso: oh no, it should not 😱 I will do another clear pass...

Sources/Logging/Logging.swift
ktoso: Great catch, thank you! It's been _years_ moving off scala but I still default to the functional style of some APIs it seems 🤣

Sources/Logging/MetadataProvider.swift
slashmo: I replicated the behavior of `MultiplexLogHandler` with this. It should be an `assert` instead of a `precondition` though: https://github.com/apple/swift-log/blob/d86803715f1251129063c6a93f3f1ff4dfc1b2aa/Sources/Logging/Logging.swift#LL938C9-L938C15

Package.swift
ktoso: Done, removed the horrible workaround with modules; 
We'll have to drop 5.0 though.

Sources/Logging/Logging.swift
ktoso: Thanks, fixed

```
    ///     - metadataProvider: The custom metadata provider this logger should invoke,
    ///                         instead of the system wide bootstrapped one, when a log statement is about to be emitted.
```

ktoso:
According to [everyone chiming in](https://github.com/apple/swift-log/pull/238/files#r1033052039) I've removed the 5.0 workaround: ad2fd8f2f1bebe67813131de443c69386cad2de7 & [`3e58207` (#238)](https://github.com/apple/swift-log/pull/238/commits/3e58207b1e1b8abe5a8a405ca567179f0dce0ddf)

I believe we are done here.

Please have a final look @tomerd, @weissi, @FranzBusch, @yim-lee @slashmo and I believe we could cut a minor with this.
I could then cut 1.0.0-beta1 releases of tracing libraries and we could do a similar "last call" before 1.0 on those libraries.

Sources/Logging/Locks.swift
ktoso: done.

tomerd (commented):
getting really close imo, couple of more points / questions

Sources/Logging/Logging.swift
```diff
@@ -1090,19 +1271,21 @@ public struct StreamLogHandler: LogHandler {
...
-        return StreamLogHandler(label: label, stream: StdioOutputStream.stderr)
+        return StreamLogHandler(label: label, metadataProvider: LoggingSystem.metadataProvider, stream: StdioOutputStream.stderr)
     }
 
     private let stream: _SendableTextOutputStream
     private let label: String
 
     public var logLevel: Logger.Level = .info
 
+    public var metadataProvider: Logger.MetadataProvider
```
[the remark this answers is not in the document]
ktoso: Yeah I don't mind going the optional route; will remove the noop impls then as well

Sources/Logging/MetadataProvider.swift
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        #else
+        public init(_ provideMetadata: @escaping () -> Metadata) {
+            self._provideMetadata = provideMetadata
+        }
+        #endif
+
+        /// Invoke the metadata provider and return the generated contextual ``Logger/Metadata``.
+        public func provideMetadata() -> Metadata {
```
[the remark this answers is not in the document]
ktoso: maybe `get()` then; value makes it sound a bit more "constant" to me.

ktoso:
Both good points, thanks 👍

ktoso:
Addressed both in https://github.com/apple/swift-log/pull/238/commits/1c3507c4895ad78787e7ca76fb75b3b768a07edb

Sources/Logging/Logging.swift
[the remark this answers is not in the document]
ktoso: [`1c3507c` (#238)](https://github.com/apple/swift-log/pull/238/commits/1c3507c4895ad78787e7ca76fb75b3b768a07edb)

Sources/Logging/MetadataProvider.swift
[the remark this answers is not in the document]
ktoso: [`1c3507c` (#238)](https://github.com/apple/swift-log/pull/238/commits/1c3507c4895ad78787e7ca76fb75b3b768a07edb)

Sources/Logging/LogHandler.swift
```diff
@@ -163,6 +169,19 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
...
 
+extension LogHandler {
+    /// Default implementation for `metadataProvider` which defaults to `nil`.
+    /// This default exists in order to facilitate source-compatible introduction of the metadataProvider protocol requirement.
+    public var metadataProvider: Logger.MetadataProvider? {
+        get {
+            nil
+        }
+        set {
+            self.log(level: .warning, message: "Attempted to set metadataProvider on \(Self.self) that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.", metadata: nil, source: "Logging", file: #file, function: #function, line: #line)
```
Lukasa, line 180: Uh, is this going to be ok? Can we safely recurse into `self` here?

Sources/Logging/MetadataProvider.swift
```diff
@@ -0,0 +1,118 @@
...
+        @usableFromInline
+        internal let _provideMetadata: @Sendable() -> Metadata
+        #else
+        @usableFromInline
+        internal let _provideMetadata: () -> Metadata
+        #endif
+
+        /// Create a new `MetadataProvider`.
+        ///
+        /// - Parameter provideMetadata: A closure extracting metadata from a given `Baggage`.
```
Lukasa, line 77: What `Baggage`?

Lukasa (approved):
Basically LGTM, a couple of notes.

Sources/Logging/MetadataProvider.swift
ktoso: Oh no, thank you. I'll skim if there's any leftover still

Sources/Logging/Logging.swift
```diff
@@ -264,6 +269,7 @@ extension Logger {
     }
 
     #else
```
weissi, line 271: Why the inlinables here? We use this exclusively as existentials, right?

Sources/Logging/Logging.swift
```diff
@@ -803,10 +919,69 @@ extension Logger {
...
+    /// - parameters:
+    ///     - label: An identifier for the creator of a `Logger`.
+    ///     - factoryWithMetadataProvider: A closure creating non-standard `LogHandler`s.
+    public init(label: String, factoryWithMetadataProvider: (String, MetadataProvider?) -> LogHandler) {
+        self = Logger(label: label, factoryWithMetadataProvider(label, LoggingSystem.metadataProvider))
+    }
 }
 
 extension Logger.Level {
-    internal var naturalIntegralValue: Int {
```
weissi, line 809: Seems unintentional

weissi (approved):
LGTM, left a few comments

Sources/Logging/Logging.swift
ktoso: Oh gosh don't tell me those slipped back in... we need to clean up how we do formatting, this is getting insane with the different versions of the tool in different projects... :|

Sources/Logging/Logging.swift
ktoso: All other methods (info, warning, error, critical) had them so it seemed to me those were omitted accidentally? maybe there's some history to this though?

Sources/Logging/MetadataProvider.swift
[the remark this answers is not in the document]
FranzBusch: I just tested this again. Since `Sendable` is only a marker protocol you don't have to make this `public`. It works without as well and propagates across modules.

FranzBusch (approved):
LGTM! Just this nit with the public protocol

Sources/Logging/Logging.swift
```diff
@@ -1201,7 +1412,7 @@ public struct SwiftLogNoOpLogHandler: LogHandler {
 
 extension Logger {
     @inlinable
-    internal static func currentModule(filePath: String = #file) -> String {
+    static func currentModule(filePath: String = #file) -> String {
```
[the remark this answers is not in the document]
ktoso: Right, thanks for spotting the few last ones.

ktoso:
Thanks for the comments @FranzBusch, @weissi - addressed!

Tests/LoggingTests/MetadataProviderTest.swift
```diff
@@ -0,0 +1,131 @@
...
+
+extension Logger.MetadataProvider {
+    static var simpleTestProvider: Logger.MetadataProvider {
+        return Logger.MetadataProvider {
+            return ["test": "provided"]
+        }
+    }
+}
+
+public struct LogHandlerThatDidNotImplementMetadataProviders: LogHandler {
```
slashmo, line 71: I like that name 😁

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Proposed solution
+
+To support this kind of runtime-generated metadata in `swift-log`, we need to extend the logging APIs in an open-ended way, to allow any kinds of metadata to be provided from the asynchronous context (i.e. task local values).
+
+We also have a desire to keep `swift-log` a "zero dependencies" library, as it intends only to focus on describing a logging API, and not incur any additional dependencies, so it can be used in the most minimal of projects.
+
+To solve this, we propose the extension of swift-log APIs with a new concept: _metadata providers_.
+
+`MetadataProvider` is struct nested in the `Logger` type, sitting alongside `MetadataValue` and the `Metadata`.
```
slashmo, line 48:
```suggestion
`MetadataProvider` is a struct nested in the `Logger` type, sitting alongside `MetadataValue` and the `Metadata`. 
```

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,328 @@
...
+```swift
+import Tracing // swift-distributed-tracing
+
+extension Logger.MetadataProvider {
+    static let myTracer = Logger.MetadataProvider {
+        guard let baggage = Baggage.current else {
+            return [:]
+        }
+        guard let spanContext = baggage.spanContext else { 
+            return nil
```
slashmo, line 80:
```suggestion
            return [:]
```

The return value is non-optional.

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,328 @@
...
+})
+```
+
+which _overrides_ the default bootstrapped metadata provider.
+
+> NOTE: Setting the metadata provider on the logger directly means the `LoggingSystem` metadata provider
+> is skipped (if defined), following how an explicitly passed handler `factory`
+> overrides the `LoggingSystem`s `factory`.
+
+Once a metadata provider was set up, when a log statement is about to be emitted, the a log handler implementation shall
```
slashmo, line 128:
```suggestion
Once a metadata provider was set up, when a log statement is about to be emitted, the log handler implementation shall
```

proposals/0001-metadata-providers.md
```diff
@@ -0,0 +1,328 @@
...
+    byID: storeID,
+    eventLoop: eventLoop,
+    logger: logger,
+    baggage: baggage
+)
+```
+
+### Explicitly passing `Logger` always
+
+Imagine we don't have metadata providers, we'd have to manually set trace IDs on loggers which doesn't really work as all libraries involved would need to know about the same specific trace ID. Event if we inverted the dependency to have `Tracing` depend on `Logging` so that we'd be able to define something like "Tracing, please populate this logger with metadata", we'd have to make sure this thing is called in all places to avoid dropping contextual metadata.
```
slashmo, line 299:
```suggestion
Imagine we don't have metadata providers, we'd have to manually set trace IDs on loggers which doesn't really work as all libraries involved would need to know about the same specific trace ID. Even if we inverted the dependency to have `Tracing` depend on `Logging` so that we'd be able to define something like "Tracing, please populate this logger with metadata", we'd have to make sure this thing is called in all places to avoid dropping contextual metadata.
```

slashmo: Not everything needs to be event-driven 😜

slashmo (approved):
I'm really happy about what we ended up with. Thank you @ktoso for further refining the APIs based on the very constructive feedback we received from everyone else involved ❤️ I'm extremely excited for this to be almost over the finish line now 🏁

Sources/Logging/Logging.swift
```diff
@@ -653,52 +660,161 @@ extension Logger {
...
+        }, validate: true)
+    }
+
+    /// `bootstrapMetadataProvider` is a one-time configuration function which globally selects the desired
+    /// ``Logger/MetadataProvider`` which should be used by all loggers created in this process from here onwards.
+    ///
+    /// `bootstrapMetadataProvider` can be called at maximum once in any given program, calling it more than once will
+    /// lead to undefined behavior, most likely a crash.
+    ///
+    /// It is also possible to `bootstrap(metadataProvider:_:)` to configure both a metadata provider,
```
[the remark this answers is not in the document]
ktoso: will do a docs cleanup pass separately maybe

Sources/Logging/MetadataProvider.swift
```diff
@@ -0,0 +1,123 @@
...
+    /// ```
+    ///
+    /// We recommend referring to [swift-distributed-tracing](https://github.com/apple/swift-distributed-tracing)
+    /// for metadata providers which make use of its tracing and metadata propagation infrastructure. It is however
+    /// possible to make use of metadata providers independently of tracing and instruments provided by that library,
+    /// if necessary.
+    public struct MetadataProvider: _SwiftLogSendable {
+        /// A default no-op metadata provider, which always returns empty metadata.
+        public static var noop: MetadataProvider {
+            return MetadataProvider { [:] }
```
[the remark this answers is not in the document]
slashmo: Oohh, I didn't know that! Cool 👍

ktoso:

ktoso:
This is ready and received a number of LGTMs -- the 5.0 failure is expected, we're dropping 5.0 (as discussed in the PR thread), the API break is the checker not realizing the new requirement has a default implementation (it is not an source/API break).

Want to give it a final look @tomerd? I'd love to merge and bump impls depending on this :-)

Sources/Logging/Logging.swift
```diff
@@ -662,41 +680,139 @@ public enum LoggingSystem {
...
+            }
+        }
+    }
+}
+
+extension LoggingSystem {
+    /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
+    /// implementation. `bootstrap` can be called at maximum once in any given program, calling it more than once will
+    /// lead to undefined behavior, most likely a crash.
+    ///
```
yim-lee, line 792:
```suggestion
    /// implementation. 
    ///
    /// - Warning:
    /// `bootstrap` can be called at maximum once in any given program, calling it more than once will
    /// lead to undefined behavior, most likely a crash.
    ///
```

yim-lee (approved):
Some suggestions on docs

ktoso:
Thank you, merged the suggestions! :)

Sources/Logging/Logging.swift
```diff
@@ -653,52 +660,161 @@ extension Logger {
...
+    /// `bootstrapMetadataProvider` can be called at maximum once in any given program, calling it more than once will
+    /// lead to undefined behavior, most likely a crash.
+    ///
+    /// It is also possible to `bootstrap(metadataProvider:_:)` to configure both a metadata provider,
+    /// and log handler backend at the same time, in a single call.
+    ///
+    /// - Parameter metadataProvider: The metadata provider to be used by this logging system.
+    public static func bootstrapMetadataProvider(_ metadataProvider: Logger.MetadataProvider?) {
+        self._metadataProviderFactory.replaceMetadataProvider(metadataProvider, validate: true)
```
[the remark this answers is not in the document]
tomerd: would be nice to remove this API if it is not necessary. the new bootstrapping API surface area does feel a bit too large and would be nice to simplify it if we can

Sources/Logging/Logging.swift
```diff
@@ -653,7 +660,21 @@ extension Logger {
 /// configured. `LoggingSystem` is set up just once in a given program to set up the desired logging backend
 /// implementation.
 public enum LoggingSystem {
-    private static let _factory = FactoryBox(StreamLogHandler.standardOutput)
+    private static let _factory = FactoryBox { label, _ in StreamLogHandler.standardOutput(label: label) }
+    private static let _metadataProviderFactory = MetadataProviderBox(.noop)
```
tomerd, line 664: I am a bit confused why we need these two boxes? seems like we are storing the metadata factory twice but maybe that can be simplified?

Sources/Logging/Logging.swift
```diff
@@ -662,41 +683,146 @@ public enum LoggingSystem {
...
+    /// implementation. 
+    ///
+    /// - Warning:
+    /// `bootstrap` can be called at maximum once in any given program, calling it more than once will
+    /// lead to undefined behavior, most likely a crash.
+    ///
+    /// - parameters:
+    ///     - metadataProvider: The `MetadataProvider` used to inject runtime-generated metadata, defaults to a "no-op" provider.
+    ///     - factory: A closure that given a `Logger` identifier, produces an instance of the `LogHandler`.
+    public static func bootstrap(metadataProvider: Logger.MetadataProvider?,
```
tomerd, line 803: metadataProvider should be second argument since optional?

Sources/Logging/Logging.swift
```diff
@@ -662,41 +683,146 @@ public enum LoggingSystem {
...
+    /// implementation. 
+    ///
+    /// - Warning:
+    /// `bootstrap` can be called at maximum once in any given program, calling it more than once will
+    /// lead to undefined behavior, most likely a crash.
+    ///
+    /// - parameters:
+    ///     - metadataProvider: The `MetadataProvider` used to inject runtime-generated metadata, defaults to nil.
+    ///     - factory: A closure that given a `Logger` identifier, produces an instance of the `LogHandler`.
+    public static func bootstrap(metadataProvider: Logger.MetadataProvider?,
```
tomerd, line 819: ditto

Sources/Logging/Logging.swift
```diff
@@ -1120,39 +1312,68 @@ public struct StreamLogHandler: LogHandler {
     }
 
     // internal for testing only
-    internal init(label: String, stream: _SendableTextOutputStream) {
+    internal init(label: String, metadataProvider: Logger.MetadataProvider?, stream: _SendableTextOutputStream) {
```
tomerd, line 1315: optional argument after required ones

Sources/Logging/Logging.swift
tomerd: can we get rid of '.noop' since it is optional?

Sources/Logging/Logging.swift
```diff
@@ -662,41 +683,146 @@ public enum LoggingSystem {
...
         self._factory.replaceFactory(factory, validate: false)
     }
 
-    fileprivate static var factory: (String) -> LogHandler {
-        return self._factory.underlying
+    internal static func bootstrapInternal(metadataProvider: Logger.MetadataProvider?) {
+        self._metadataProviderFactory.replaceMetadataProvider(metadataProvider, validate: false)
+    }
+
+    internal static func bootstrapInternal(metadataProvider: Logger.MetadataProvider?,
```
tomerd, line 723: optional argument after required ones

Sources/Logging/Logging.swift
```diff
@@ -803,6 +929,65 @@ extension Logger {
...
+    ///
+    /// The `label` should identify the creator of the `Logger`. This can be an application, a sub-system, or even
+    /// a datatype.
+    ///
+    /// This initializer provides an escape hatch in case the global default logging backend implementation (set up
+    /// using `LoggingSystem.bootstrap`) is not appropriate for this particular logger.
+    ///
+    /// - parameters:
+    ///     - label: An identifier for the creator of a `Logger`.
+    ///     - factoryWithMetadataProvider: A closure creating non-standard `LogHandler`s.
```
tomerd, line 961: nit: docs incorrect

Sources/Logging/Logging.swift
```diff
@@ -803,6 +929,65 @@ extension Logger {
     public init(label: String, factory: (String) -> LogHandler) {
         self = Logger(label: label, factory(label))
     }
+
+    /// Construct a `Logger` given a `label` identifying the creator of the `Logger` and a non-standard ``Logger/MetadataProvider``.
+    ///
+    /// The `label` should identify the creator of the `Logger`. This can be an application, a sub-system, or even
+    /// a datatype.
+    ///
+    /// This initializer provides an escape hatch in case the global default logging backend implementation (set up
```
tomerd, line 938: is the escape hatch because of real world need, or more theoretical? if the latter, maybe we can add later if/when we have a real need?

Sources/Logging/Logging.swift
```diff
@@ -786,7 +912,7 @@ extension Logger {
     /// - parameters:
     ///     - label: An identifier for the creator of a `Logger`.
     public init(label: String) {
-        self.init(label: label, LoggingSystem.factory(label))
+        self.init(label: label, LoggingSystem.factory(label, .bootstrapped))
```
tomerd, line 915: could you elaborate with this does? the Logger defining ".bootstrapped" state on the factory seems odd?

Sources/Logging/MetadataProvider.swift
```diff
@@ -0,0 +1,123 @@
...
+    /// }
+    /// ```
+    ///
+    /// We recommend referring to [swift-distributed-tracing](https://github.com/apple/swift-distributed-tracing)
+    /// for metadata providers which make use of its tracing and metadata propagation infrastructure. It is however
+    /// possible to make use of metadata providers independently of tracing and instruments provided by that library,
+    /// if necessary.
+    public struct MetadataProvider: _SwiftLogSendable {
+        /// A default no-op metadata provider, which always returns empty metadata.
+        public static var noop: MetadataProvider {
```
tomerd, line 62: can remove this now that provider is optional?

Sources/Logging/Logging.swift
ktoso: Ah yes, we can remove .noop -- thank you

Sources/Logging/Logging.swift
ktoso: Technically this was to answer @fabianfett's request of "don't try reading task locals, there won't be any here anyway" when they'd manually set metadata in a nio handler, but looking at it that would be pretty destructive since they would also disable any user-provided even constant metadata providers -- so they likely should not be doing that... I'll remove this but wonder if it may come back someday

Sources/Logging/Logging.swift
ktoso: right, thanks :)

Sources/Logging/Logging.swift
ktoso: We can remove the "bootstrapped" computed property but the logic needs to be here: get the metadata provider instance the system was bootstrapped with when we create a new logger.

I'll make this 

```
self.init(label: label, LoggingSystem.factory(label, LoggingSystem.metadataProvider))
```

if you think that's nicer

Sources/Logging/MetadataProvider.swift
ktoso: We can I guess, it'll force writing the same `MetadataProvider { [:] }` by us or users sometimes... if we frantically want to avoid a "noop" API then sure we can remove this...

Sources/Logging/Logging.swift
ktoso: Heh, so I did this, but the callsites become very ugly...

```
LoggingSystem.bootstrapInternal() { label, metadataProvider in
  logging.makeWithMetadataProvider(label: label, metadataProvider: metadataProvider)
}
LoggingSystem.bootstrapInternal(metadataProvider: .myCoolTracer) { label, metadataProvider in
  logging.makeWithMetadataProvider(label: label, metadataProvider: metadataProvider)
}
```

vs

```
LoggingSystem.bootstrapInternal() { label, metadataProvider in
  logging.makeWithMetadataProvider(label: label, metadataProvider: metadataProvider)
}

LoggingSystem.bootstrapInternal({ label, metadataProvider in
logging.makeWithMetadataProvider(label: label, metadataProvider: metadataProvider)
}, metadataProvider: .myCoolTracer)
```

So... I'd suggest to keep existing API but we can change if you feel strongly about it 🤔

Sources/Logging/Logging.swift
ktoso: thanks fixed

ktoso: And I think we can remove a few of those inits actually...

Sources/Logging/Logging.swift
ktoso: I did that then

ktoso:
Followed up on your comments in https://github.com/apple/swift-log/pull/238/commits/91437c0bf675feb0987eec60bb23968379414f5b @tomerd  -- the reordering of the params I'm doubtful about, see here: https://github.com/apple/swift-log/pull/238#discussion_r1034150298

but I could flip those around still... that's the final thing i guess?

Sources/Logging/Logging.swift
ktoso: Okey, let's do it 😄

Sources/Logging/Logging.swift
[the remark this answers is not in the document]
ktoso: Agreed, removed this one. We only gain one additional bootstrap method now.

Sources/Logging/Logging.swift
```diff
@@ -694,18 +706,6 @@ public enum LoggingSystem {
         }, validate: true)
     }
 
-    /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
-    /// implementation. `bootstrap` can be called at maximum once in any given program, calling it more than once will
-    /// lead to undefined behavior, most likely a crash.
-    ///
-    /// - parameters:
-    ///     - factory: A closure that given a `Logger` identifier, produces an instance of the `LogHandler`.
-    public static func bootstrap(_ factory: @escaping (String) -> LogHandler) {
```
ktoso, line 664: this just moved "up" a little in the file, to be organized together with the other ones

Sources/Logging/Logging.swift
ktoso: Dropped it entirely, we only have those *ONE* new shape of bootstraps:

```
    internal static func bootstrapInternal(_ factory: @escaping (String, Logger.MetadataProvider?) -> LogHandler,
                                           metadataProvider: Logger.MetadataProvider?) {
```


```
    public static func bootstrap(_ factory: @escaping (String, Logger.MetadataProvider?) -> LogHandler,
                                 metadataProvider: Logger.MetadataProvider?) {
```

Sources/Logging/Logging.swift
```diff
@@ -1090,19 +1206,21 @@ public struct StreamLogHandler: LogHandler {
 
     /// Factory that makes a `StreamLogHandler` to directs its output to `stdout`
     public static func standardOutput(label: String) -> StreamLogHandler {
-        return StreamLogHandler(label: label, stream: StdioOutputStream.stdout)
+        return StreamLogHandler(label: label, stream: StdioOutputStream.stdout, metadataProvider: LoggingSystem.metadataProvider)
```
ktoso, line 1209: default impls "do the right thing"™

ktoso:
Discussed offline with @tomerd and I did another pass and removed as much new public API as possible. 


We're _done_! I'm going to merge and release a minor version with this. This will cause an all our tracing repos to "fall into place" and we're soon going to 1.0-beta there 🥳

ktoso:
To make sure folks don't panic: The API breakage detector is wrong here; we add a protocol requirement, but we add a default impl as well, this is source-compatible and purely additive.

PR #246, opened 2023-01-08
Default subscript implementation.

nashysolutions:
The `LogHandler` protocol provides a default implementation of the metadata subscript function.

### Motivation:

The requirement for a subscript seems unnecessary, given that it's typically boilerplate.

### Modifications:

- Deleted the requirement from the protocol signature. 
- Installed a default implementation in the protocol extension.

### Result:

The library user no longer has to provide the boilerplate.

### DIFF
--- a/Sources/Logging/LogHandler.swift
+++ b/Sources/Logging/LogHandler.swift
@@ -139,15 +139,6 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
     @available(*, deprecated, renamed: "log(level:message:metadata:source:file:function:line:)")
     func log(level: Logging.Logger.Level, message: Logging.Logger.Message, metadata: Logging.Logger.Metadata?, file: String, function: String, line: UInt)
 
-    /// Add, remove, or change the logging metadata.
-    ///
-    /// - note: `LogHandler`s must treat logging metadata as a value type. This means that the change in metadata must
-    ///         only affect this very `LogHandler`.
-    ///
-    /// - parameters:
-    ///    - metadataKey: The key for the metadata item
-    subscript(metadataKey _: String) -> Logger.Metadata.Value? { get set }
-
     /// Get or set the entire metadata storage as a dictionary.
     ///
     /// - note: `LogHandler`s must treat logging metadata as a value type. This means that the change in metadata must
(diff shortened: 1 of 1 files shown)


ktoso:
Thanks for the PR but that'd change behavior in existing adopting libraries and potentially cause breaking issues, if they really did something else in those implementations -- I don't think this is a safe change to apply.

ktoso (changes requested):
I don't think this is a safe change to apply, the win from trying to make it happen also seems rather small so I don't think we need to pursue this further 🤔

nashysolutions:
I don't see why it would be unsafe. Happy to close, if you think that's best.

nashysolutions:
> Thanks for the PR but that'd change behavior in existing adopting libraries and potentially cause breaking issues, if they really did something else in those implementations -- I don't think this is a safe change to apply.

Here is an example implementation of how this change does not cause a break. The print statement 'setting' still prints in each scenario.

```swift
protocol Vehicle {
    var store: [String: String] { get set }
    
    // Remove this to test
    subscript(attribute key: String) -> String?
}

// Remove this to test
extension Vehicle {
    
    subscript(attribute key: String) -> String? {
        get {
            store[key]
        }
        set {
            store[key] = newValue
        }
    }
}

struct Car: Vehicle {
    
    var store: [String: String] = [:]
    
    subscript(attribute key: String) -> String? {
        get {
            store[key]
        }
        set {
            print("setting")
            store[key] = newValue
        }
    }
}

struct ContentView: View {
    
    init() {
        var car = Car()
        car[attribute: "fuel"] = "petrol"
        print(car.store)
    }
        
    var body: some View {
        Text("Hello, World!")
    }
}
```

ktoso:
Sure. 

- Removing the protocol requirement gets us to a non-working customization point, libs would compile, but not work as expected -- that's the state of this PR. The bad behavior is shown by my snippets above. 

You also mention to remove the extension, but if we do that, we have non-compiling code:

- Removing the protocol requirement AND the extension results in non-compiling swift-log, since we must be calling this subscript, but there is no declaration of it anywhere, i.e.:

```
error: repl.swift:75:8: error: value of type 'any Vehicle' has no subscripts
    car[attribute: "fuel"] = "petrol"
```

So... no, wa can't make this change, in any way you intended those "remove to test" to be interpreted, they result either in incorrect or non compiling code.

If you mean removing "just the extension" is just status quo of the library as it is today.

I'm going to close this PR if you don't mind, there's nothing actionable we can do here AFAICS.

nashysolutions:
The part I missed in your example, was the casting to vehicle.

```swift
var car: Vehicle = Car()
```

I didn't quite realise this behaviour when removing the requirement but keeping an extension. Apologies.

PR #249, opened 2023-01-10
Remove default implementation for protocol requirement

### LINKED ISSUE #248: Infinite recursion between deprecated API's
### Expected behavior
Print log to debugger

### Actual behavior

Screenshot 2023-01-09 at 07 41 06

Screenshot 2023-01-09 at 07 41 16

### Steps to reproduce

```swift
import Logging

struct Hello: LogHandler {
    
    var metadata: Logging.Logger.Metadata = [:]
    
    var logLevel: Logging.Logger.Level = .error
    
    init(_ label: String) {
    }
    
    subscript(metadataKey metadataKey: String) -> Logger.Metadata.Value? {
        get {
            return self.metadata[metadataKey]
        }
        set(newValue) {
            self.metadata[metadataKey] = newValue
        }
    }
}
```

```swift
import Logging

final class Client {
    
    static let shared = Client()
    
    private init() {
        LoggingSystem.bootstrap(Hello.init)
    }
}

@main
struct TestApp: App {
    
    init() {
        _ = Client.shared
    }
    
    var body: some Scene {
        WindowGroup {
            ContentView()
        }
    }
}
```

```swift
import Logging

struct ContentView: View {
    
    init() {
        var log = Logger(label: "unique.value")
        log[metadataKey: "mykey"] = "jelly"
        log.error("hi")
    }
        
    var body: some View {
        Text("Hello, World!")
    }
}
```

The `init` for SwiftUI view's may be called multiple times, but this doesn't seem to be the issue.

Latest release as of (9th Jan 2023) - v1.4.4

swift-driver version: 1.62.15 Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)
Target: arm64-apple-macosx13.0
Darwin Roberts-Personal-MacBook-Pro.local 22.1.0 Darwin Kernel Version 22.1.0: Sun Oct  9 20:14:30 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T8103 arm64

Xcode Version 14.1 (14B47b)

nashysolutions:
The `LogHandler` protocol demands an implementation of the following.

```swift
func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, source: String, file: String, function: String, line: UInt)
```

However, the compiler does not enforce this. 
This PR ensures that it does.

Fixes #248 

### Motivation:

If you do not implement the above protocol requirement, your code will crash. So the compiler should help you here.

### Modifications:

A default implementation of the protocol requirement is no longer provided.

### Result:

An error is now thrown if you do not implement this function.

Screenshot 2023-01-10 at 04 29 17

### DIFF
--- a/Sources/Logging/LogHandler.swift
+++ b/Sources/Logging/LogHandler.swift
@@ -134,11 +134,6 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
              function: String,
              line: UInt)
 
-    /// SwiftLog 1.0 compatibility method. Please do _not_ implement, implement
-    /// `log(level:message:metadata:source:file:function:line:)` instead.
-    @available(*, deprecated, renamed: "log(level:message:metadata:source:file:function:line:)")
-    func log(level: Logging.Logger.Level, message: Logging.Logger.Message, metadata: Logging.Logger.Metadata?, file: String, function: String, line: UInt)
-
     /// Add, remove, or change the logging metadata.
     ///
     /// - note: `LogHandler`s must treat logging metadata as a value type. This means that the change in metadata must
@@ -164,19 +159,17 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
 }
 
 extension LogHandler {
-    @available(*, deprecated, message: "You should implement this method instead of using the default implementation")
-    public func log(level: Logger.Level,
-                    message: Logger.Message,
-                    metadata: Logger.Metadata?,
-                    source: String,
-                    file: String,
-                    function: String,
-                    line: UInt) {
-        self.log(level: level, message: message, metadata: metadata, file: file, function: function, line: line)
-    }
-
+    
+    /// SwiftLog 1.0 compatibility method. Please do _not_ implement, implement
+    /// `log(level:message:metadata:source:file:function:line:)` instead.
     @available(*, deprecated, renamed: "log(level:message:metadata:source:file:function:line:)")
-    public func log(level: Logging.Logger.Level, message: Logging.Logger.Message, metadata: Logging.Logger.Metadata?, file: String, function: String, line: UInt) {
+    public func log(
+        level: Logging.Logger.Level,
+        message: Logging.Logger.Message,
+        metadata: Logging.Logger.Metadata?,
+        file: String,
+        function: String,
+        line: UInt) {
         self.log(level: level,
                  message: message,
                  metadata: metadata,
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1171,11 +1171,16 @@ public struct StreamLogHandler: LogHandler {
 
 /// No operation LogHandler, used when no logging is required
 public struct SwiftLogNoOpLogHandler: LogHandler {
+    
     public init() {}
 
     public init(_: String) {}
 
     @inlinable public func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, file: String, function: String, line: UInt) {}
+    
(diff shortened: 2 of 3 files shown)


Sources/Logging/LogHandler.swift
```diff
@@ -134,11 +134,6 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
              function: String,
              line: UInt)
 
-    /// SwiftLog 1.0 compatibility method. Please do _not_ implement, implement
-    /// `log(level:message:metadata:source:file:function:line:)` instead.
-    @available(*, deprecated, renamed: "log(level:message:metadata:source:file:function:line:)")
-    func log(level: Logging.Logger.Level, message: Logging.Logger.Message, metadata: Logging.Logger.Metadata?, file: String, function: String, line: UInt)
```
ktoso, line 140: As the comment states: `SwiftLog 1.0 compatibility method.` 

This method is here for compatibility reasons and we _cannot_ remove it without causing source-breaks in log handler implementations which adopted during 1.0 cycle and did not upgrade to implementing the new methods.

This was purposefully implemented this way in order to keep source compatibility, and while it is annoying to implement a new handler since you have to read the comment or note on the method, it guarantees that all 1.x libraries work well together.

We can only do such change once we do a source-breaking 2.0 release which at this point in time is not being considered.

Sources/Logging/LogHandler.swift
```diff
@@ -164,19 +159,17 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
...
-                    metadata: Logger.Metadata?,
-                    source: String,
-                    file: String,
-                    function: String,
-                    line: UInt) {
-        self.log(level: level, message: message, metadata: metadata, file: file, function: function, line: line)
-    }
-
+    
+    /// SwiftLog 1.0 compatibility method. Please do _not_ implement, implement
```
ktoso, line 163: this cannot be an extension method; this is a protocol requirement that existing `LogHandler` implementations implement; removing the requirement will not yield the expected behavior with such libraries.

ktoso (changes requested):
Thank you for the well intended PR, but we cannot accept this as it would be source compatibility breaking which we're not ready to do just now.

I explained more in-line.

Sources/Logging/Logging.swift
```diff
@@ -1171,11 +1171,16 @@ public struct StreamLogHandler: LogHandler {
...
 /// No operation LogHandler, used when no logging is required
 public struct SwiftLogNoOpLogHandler: LogHandler {
+    
     public init() {}
 
     public init(_: String) {}
 
     @inlinable public func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, file: String, function: String, line: UInt) {}
+    
+    @inlinable public func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata?, source: String, file: String, function: String, line: UInt) {
```
ktoso, line 1181: in other words, this specific forcing of the requirement impl breaks compilation of any LogHandler implementation that was implementing the 1.0 API (the one above), so we can't force such change without making a major version bump of swift-log.

ktoso:
Double checked and this would definitely be a source break -- thanks but we can't do these right now.

nashysolutions:
It's not a breaking change!!!

nashysolutions:
Thanks @ktoso for the explanation. Appreciate your time.

ktoso:
No worries, thanks for trying to improve the lib! It's tricky to maintain a source compatible lib like this :)

PR #250, opened 2023-01-18
Allow passing explicit provider into the stream handlers

ktoso:
Small follow up to https://github.com/apple/swift-log/pull/238 which went perhaps _too_ minimal with the API additions.

Normally you don't need to customize the default stdout handler, but when you do, we made it a bit awkward right now.
It'll pick up the configured metadata provider by default, but we should follow the pattern of passing it in, also in order to allow this syntax:

```swift
        LoggingSystem.bootstrap(
            StreamLogHandler.standardOutput,
            metadataProvider: .exampleMetadataProvider)
```

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1209,17 +1209,17 @@ public struct StreamLogHandler: LogHandler {
         return StreamLogHandler(label: label, stream: StdioOutputStream.stdout, metadataProvider: LoggingSystem.metadataProvider)
     }
 
-    /// Factory that makes a `StreamLogHandler` to directs its output to `stdout`
+    /// Factory that makes a `StreamLogHandler` that directs its output to `stdout`
     public static func standardOutput(label: String, metadataProvider: Logger.MetadataProvider?) -> StreamLogHandler {
         return StreamLogHandler(label: label, stream: StdioOutputStream.stdout, metadataProvider: metadataProvider)
     }
 
-    /// Factory that makes a `StreamLogHandler` to directs its output to `stderr`
+    /// Factory that makes a `StreamLogHandler` that directs its output to `stderr`
     public static func standardError(label: String) -> StreamLogHandler {
         return StreamLogHandler(label: label, stream: StdioOutputStream.stderr, metadataProvider: LoggingSystem.metadataProvider)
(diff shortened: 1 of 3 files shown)


Sources/Logging/Logging.swift
```diff
@@ -1209,11 +1209,21 @@ public struct StreamLogHandler: LogHandler {
         return StreamLogHandler(label: label, stream: StdioOutputStream.stdout, metadataProvider: LoggingSystem.metadataProvider)
     }
 
+    /// Factory that makes a `StreamLogHandler` to directs its output to `stdout`
```
yim-lee, line 1212:
```suggestion
    /// Factory that makes a `StreamLogHandler` that directs its output to `stdout`
```

Sources/Logging/Logging.swift
```diff
@@ -1209,11 +1209,21 @@ public struct StreamLogHandler: LogHandler {
...
+    public static func standardOutput(label: String, metadataProvider: Logger.MetadataProvider?) -> StreamLogHandler {
+        return StreamLogHandler(label: label, stream: StdioOutputStream.stdout, metadataProvider: metadataProvider)
+    }
+
     /// Factory that makes a `StreamLogHandler` to directs its output to `stderr`
```
yim-lee, line 1217:
```suggestion
    /// Factory that makes a `StreamLogHandler` that directs its output to `stderr`
```

Sources/Logging/Logging.swift
```diff
@@ -1209,11 +1209,21 @@ public struct StreamLogHandler: LogHandler {
...
     public static func standardError(label: String) -> StreamLogHandler {
         return StreamLogHandler(label: label, stream: StdioOutputStream.stderr, metadataProvider: LoggingSystem.metadataProvider)
     }
 
+    /// Factory that makes a `StreamLogHandler` to directs its output to `stderr`
```
yim-lee, line 1222:
```suggestion
    /// Factory that makes a `StreamLogHandler` that direct its output to `stderr`
```

yim-lee (approved):

tomerd (approved):

ktoso:
Thanks for the review! 🙏

PR #251, opened 2023-01-18
Emit correct metadata from StreamLogHandler

slashmo:
Return the correct metadata to log from the `StreamLogHandler`.

### Motivation:

At the moment, `StreamLogHandler` always returns the explicitly passed metadata as the metadata to log. This throws away any metadata provided by the registered `MetadataProvider`.

### Modifications:

Return the merged metadata, both explicit and provided.

### Result:

When using the `StreamLogHandler`, metadata set through `MetadataProvider`s will now correctly show up in the logs.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1287,7 +1287,7 @@ public struct StreamLogHandler: LogHandler {
             metadata.merge(explicit, uniquingKeysWith: { _, explicit in explicit })
         }
 
-        return explicit
+        return metadata
     }
 
     private func prettify(_ metadata: Logger.Metadata) -> String? {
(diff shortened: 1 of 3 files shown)


ktoso:
Ouch, nice catch -- thanks @slashmo.

We should have a test that covers this, can you please add the following test case:

```swift
    func testStreamLogHandlerWritesIncludeMetadataProviderMetadata() {
        let interceptStream = InterceptStream()
        LoggingSystem.bootstrapInternal({ _, metadataProvider in
            StreamLogHandler(label: "test", stream: interceptStream, metadataProvider: metadataProvider)
        }, metadataProvider: .exampleMetadataProvider)
        let log = Logger(label: "test")

        let testString = "my message is better than yours"
        log.critical("\(testString)")

        let messageSucceeded = interceptStream.interceptedText?.trimmingCharacters(in: .whitespacesAndNewlines).hasSuffix(testString)

        XCTAssertTrue(messageSucceeded ?? false)
        XCTAssertEqual(interceptStream.strings.count, 1)
        let message = interceptStream.strings.first!
        XCTAssertTrue(message.contains("example=example-value"), "message must contain metadata, was: \(message)")
    }
```

Thank you!

slashmo:
> can you please add the following test case in LoggingTests
> @ktoso 

Good point, I'll add it.

ktoso (approved):
LGTM, thanks for adding the test and spotting the issue

yim-lee (approved):

PR #252, opened 2023-01-21
Avoid logging warnings when handler does not support metadataproviders

ktoso:
Discovered by https://github.com/vapor/console-kit/pull/177

---

We were setting the metadata provider "always" so even without implementing the right factory function it would have tried to set a provider during logger creation. Since this happens in code which did not adopt metadata providers, it ends up as logging a warning for creating loggers.

This can be very noisy, e.g. a plain new vapor project would be like this:

```
[1357/1357] Linking Run
Build complete! (47.19s)
[ WARNING ] Attempted to set metadataProvider on ConsoleLogger that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.
[ NOTICE ] Server starting on http://127.0.0.1:8080
[ WARNING ] Attempted to set metadataProvider on ConsoleLogger that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.
[ WARNING ] Attempted to set metadataProvider on ConsoleLogger that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.
[ INFO ] GET / [request-id: AAF6BC94-109E-4492-B061-F26B4060DC9F]
[ WARNING ] Attempted to set metadataProvider on ConsoleLogger that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.
[ WARNING ] Attempted to set metadataProvider on ConsoleLogger that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.
[ WARNING ] Attempted to set metadataProvider on ConsoleLogger that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.
```

The change in https://github.com/apple/swift-log/commit/6f9b9d4cd7cce207b5256adb5ac6fc4c7f6e9040 here makes it such that the only way to make use of metadata providers is to make sure to bootstrap with a factory that accepts one. OR you can always set it manually later on, if you know your log handler will support it.

This avoids the noisy logging.

The warning logging still would happen here if you do yourself 

```
log.metadataProvider = ....
```

on a log handler that does not support it. I think this is fine (?), but it would happen for every instance of such call... 

I could be convinced to remove the warning entirely if we think this is still risking it too much.

---

For context, the origin of the warning is this:

```
extension LogHandler {
    /// Default implementation for `metadataProvider` which defaults to `nil`.
    /// This default exists in order to facilitate source-compatible introduction of the `metadataProvider` protocol requirement.
        get {
            nil
        }
        set {
            self.log(level: .warning, message: "Attempted to set metadataProvider on \(Self.self) that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.", metadata: nil, source: "Logging", file: #file, function: #function, line: #line)
        }
    }
}
```

so if someone tried to set a provider on a logger whose handler does not support it, the intent was to give a warning.

### DIFF
--- a/Sources/Logging/LogHandler.swift
+++ b/Sources/Logging/LogHandler.swift
@@ -177,7 +177,11 @@ extension LogHandler {
             nil
         }
         set {
-            self.log(level: .warning, message: "Attempted to set metadataProvider on \(Self.self) that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.", metadata: nil, source: "Logging", file: #file, function: #function, line: #line)
+            #if DEBUG
+            if LoggingSystem.warnOnceLogHandlerNotSupportedMetadataProvider(Self.self) {
+                self.log(level: .warning, message: "Attempted to set metadataProvider on \(Self.self) that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.", metadata: nil, source: "Logging", file: #file, function: #function, line: #line)
+            }
+            #endif
         }
     }
 }
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -663,6 +663,10 @@ public enum LoggingSystem {
     private static let _factory = FactoryBox { label, _ in StreamLogHandler.standardOutput(label: label) }
     private static let _metadataProviderFactory = MetadataProviderBox(nil)
 
+    #if DEBUG
+    private static var _warnOnceBox: WarnOnceBox = WarnOnceBox()
+    #endif
+
     /// `bootstrap` is a one-time configuration function which globally selects the desired logging backend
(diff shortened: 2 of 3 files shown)


ktoso:
cc @tim0x

yim-lee (approved):

tomerd:
another option is to try and emit once n first occurrence. this will have done performance cost but maybe static atomic check is not too bad. better maybe we only do it #if DEBUG

ktoso:
Only doing it in debug is definitely a good idea as well! Perhaps they and the atomic approach combined would be the best 🤔

ktoso:
Sounds like a plan, done: https://github.com/apple/swift-log/pull/252/commits/3efaf3773776995bdc911893138745bec3aeaeb4

Issue #253, opened 2023-01-23
MultiplexLogHandler to learn about metadata providers


ktoso:
We missed to update the `MultiplexLogHandler` to also deal with the providers.

> I was trying to write a metadata provider test for a [Discord logger](https://github.com/MahdiBM/DiscordBM/pull/25) i've made, then i noticed since log handlers like MultiplexLogHandler still don't support metadata providers, metadata providers won't even work with the logger (since it's using the MultiplexLogHandler).

The solution is just to implement the handling in the handler.

PR #254, opened 2023-01-23
Handle providers properly in multiplex log handler

ktoso:
Resolves https://github.com/apple/swift-log/issues/253

The handling is a bit tricky, but we uphold reasonable semantics now.
We could cache the calculation a bit more perhaps rather than each time in the `get{}` but this has the right semantics at least.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1052,7 +1052,7 @@ extension Logger {
 public struct MultiplexLogHandler: LogHandler {
     private var handlers: [LogHandler]
     private var effectiveLogLevel: Logger.Level
-    /// This provider is in addition to any providers the handlers this multiplexes to may have.
+    /// This metadata provider runs after all metadata providers of the multiplexed handlers.
     private var _metadataProvider: Logger.MetadataProvider?
 
     /// Create a `MultiplexLogHandler`.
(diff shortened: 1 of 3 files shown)


ktoso:
Please have a look @slashmo

Sources/Logging/Logging.swift
```diff
@@ -1050,6 +1052,8 @@ extension Logger {
 public struct MultiplexLogHandler: LogHandler {
     private var handlers: [LogHandler]
     private var effectiveLogLevel: Logger.Level
+    /// This provider is in addition to any providers the handlers this multiplexes to may have.
```
slashmo, line 1055:
```suggestion
    /// This metadata provider runs after all metadata providers of the multiplexed handlers.
```

slashmo (approved):
LGTM. In my opinion, the semantics make sense.

tomerd (approved):

Sources/Logging/Logging.swift
```diff
@@ -1072,6 +1083,43 @@ public struct MultiplexLogHandler: LogHandler {
...
+                    }
+                } else if let multiplexHandlerProvider = self._metadataProvider {
+                    return multiplexHandlerProvider
+                } else {
+                    return nil
+                }
+            } else {
+                var providers: [Logger.MetadataProvider] = []
+                let hasSpecifiedAdditionalMetadataProvider = (self._metadataProvider != nil ? 1 : 0)
+                providers.reserveCapacity(self.handlers.count + hasSpecifiedAdditionalMetadataProvider)
```
ktoso, line 1103:
```suggestion
                providers.reserveCapacity(self.handlers.count + additionalMetadataProviderCount)
```

PR #258, opened 2023-01-26
update recent addition to 5.0 syntax

tomerd:
motivation: support 5.0 clients

changes: return value explicitly

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -734,7 +734,7 @@ public enum LoggingSystem {
     /// but an attempt was made to set a metadata provider on such handler. In order to avoid flooding the system with
     /// warnings such warning is only emitted in debug mode, and even then at-most once for a handler type.
     internal static func warnOnceLogHandlerNotSupportedMetadataProvider<Handler: LogHandler>(_ type: Handler.Type) -> Bool {
-        self._warnOnceBox.warnOnceLogHandlerNotSupportedMetadataProvider(type: type)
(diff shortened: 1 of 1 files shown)


tomerd:
see https://github.com/swift-server/swift-service-lifecycle/pull/129

ktoso (commented):
Lgtm, but there's a discussion to have here. During https://github.com/apple/swift-log/pull/238 the consensus was to drop 5.0 support, so that's what we did and documented in release notes, removed the CI etc. So if we want to retract that move we need to do so properly, rather than with an adhoc fix that will regress again without CI.

The summary is that the 5.0 compiler can't survive parsing a task local, even when #if-ed out. There's no workaround in the language. So we'd have to do an very involved tricky dance to make some test into a separate module, only include it in build in new package files and modify the generate Linux tests script to handle this -- I had done all that but everyone in the review had voted against supporting 5.0 at the cost of this complexity so we dropped it.

So... 5.0 isn't supported, as documented in 1.5 release and adhoc fixing it is not the right way -- either we revive the workarounds I did, or we stick to our decision to drop 5.0

tomerd:
@ktoso right I remember that dicussion. it does means we need to remove the `Package@swift-5.0.swift` file so that SwiftPM will happily satisfy the dependency causing potential issues to downstream projects

ktoso:
Hah I see, we missed to remove the file, here's a PR to rectify that: https://github.com/apple/swift-log/pull/259

ktoso:
Superseded by https://github.com/apple/swift-log/pull/259

PR #259, opened 2023-01-30
Remove leftover 5.0 Package.swift; not supported since 1.5.0

ktoso:
We dropped Swift 5.0 support in 1.5.0, but forgot to remove the Package.swift file for that tools version, removing now.


yim-lee (approved):

ktoso:
Release wise we're not rushing to push this one; whenever we have more to release this'll come with it.

Context: The change in Swift 5.0 support was announced in 1.5.0: https://github.com/apple/swift-log/releases/tag/1.5.0 but we missed to remove the support explicitly; This only enforces what was already announced before.

PR #269, opened 2023-06-21
Use #if canImport(Darwin) where possible

FranzBusch:
### Motivation:

Instead of adding every Darwin OS we should rather check if we can import `Darwin`


glbrntt (approved):

tomerd (approved):

tomerd:

ktoso:
Please assign milestone when closing -- fixed now and going to release :+1:

PR #274, opened 2023-07-25
Add wynioux/DeveloperSuite

wynioux:
wynioux/DeveloperSuite added to logging backend implementation table.

### Motivation:

I am developing new open source QA & debug tool. I want to support swift-log, I'm adding my tool to the backend implementation table as alpha tests are started.

### Modifications:

Add a link to the package to logging backend implementation table in the README.


ktoso:
I'm a little confused by the project linked, as in, unsure if one would use it just as a logger backend -- seems not really? Only if already bought into this suite then one would be looking for the logger integration. 

I'm unsure if we should list it, WDYT @tomerd ( https://github.com/wynioux/DeveloperSuite ) ?

ktoso:
I'm going to close this PR for now, we've not heard back from the author for a while.

PR #280, opened 2023-11-16
Don't assume localtime() result is non-nil

gwynne:
Check for `localtime(3)` returning a `nil` result.

### Motivation:

This avoids calling `strftime(3)` with a `NULL` pointer in the odd case that `localtime(3)` fails. It also has the useful, if incidental, side effect of fixing building `swift-log` on Arch Linux (see https://github.com/vapor/toolbox/issues/430 for more info).

### Modifications:

Added a `guard` which returns a generic failure string if `localtime(3)` returns `nil`.

### Result:

There should be no visible impact on virtually any users, seeing as how this has not been previously encountered, and the use of `swift-log` on at least one unofficial (but apparently commonly used) platform will be fixed.


tomerd (approved):

tomerd:
fixing CI in https://github.com/apple/swift-log/pull/281

tomerd:

PR #283, opened 2023-11-28
Changes to build with Musl.

al45tair:
Added imports of Musl in a few places.  Also added definitions for `systemStderr` and `systemStdout`.

### Motivation:

We would like to be able to build swift-log on top of Musl, as well as the usual Glibc.

### Modifications:

Really just adds some `import Musl`s in conditionals, alongside the `import Glibc`s.

### Result:

swift-log will build against Musl.


yim-lee:

tomerd (approved):

PR #297, opened 2024-05-09
feat: Implement Copy-On-Write (CoW) behavior for Logger struct

ayushi2103:
Added CoW implementation for Logger struct to optimize performance and minimize memory overhead.

### Motivation:

LogHandler is an existential type and has a String requiring two words so copying a Logger instance duplicates 7 words and executes two ARC operations. By implementing CoW with a boxed LogHandler, we will duplicate 1 word and one ARC operation, enhancing efficiency particularly when Logger instances are frequently passed around.

### Modifications:

Added a private property `_handler` of type `Box<LogHandler>` to hold the `LogHandler` and updated the `handler` property to be computed, allowing transparent access to the `LogHandler`.

### Result:

When the Logger instances will be passed around, the change will reduce memory overhead and potentially improve performance.


ayushi2103:

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,41 @@ import WASILibc
...
     /// An identifier of the creator of this `Logger`.
     public let label: String
+    
+    /// A computed property to access the `LogHandler`.
+    public var handler: LogHandler {
+        get {
+            return _handler.value
+        }
+        set {
+            if !(isKnownUniquelyReferenced(&_handler)) {
```
ayushi2103, line 54: Should I use `isUniquelyReferenced` or `isUniquelyReferencedNonObjC`?

ayushi2103:
In which file should I add the unit test?

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,41 @@ import WASILibc
...
     }
 
     internal init(label: String, _ handler: LogHandler) {
         self.label = label
-        self.handler = handler
+        self._handler = Box(value: handler)
+    }
+}
+
+private final class Box<T> {
```
FranzBusch, line 73: Can we nest this type inside the `struct Logger` please. Additionally can we make it non generic and rather concrete so that it holds both the handler and the label.

Sources/Logging/Logging.swift
FranzBusch: `isKnownUniquelyReferenced` is correct

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,41 @@ import WASILibc
 /// logger.info("Hello World!")
 /// ```
 public struct Logger {
-    @usableFromInline
-    var handler: LogHandler
+    /// A private property to hold the boxed `LogHandler`.
+    private var _handler: Box<LogHandler>
 
     /// An identifier of the creator of this `Logger`.
     public let label: String
```
FranzBusch, line 46: We need to move this into the box as well

Sources/Logging/Logging.swift
weissi, line 54:
```suggestion
            if !isKnownUniquelyReferenced(&self._handler) {
```

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,41 @@ import WASILibc
...
+    /// A computed property to access the `LogHandler`.
+    public var handler: LogHandler {
+        get {
+            return _handler.value
+        }
+        set {
+            if !(isKnownUniquelyReferenced(&_handler)) {
+                _handler = Box(value: newValue)
+            } else {
+                _handler.value = newValue
```
weissi, line 57: please use explicit `self.` everywhere, otherwise these things are very hard to read especially in diffs which might not have enough context to spot where this is from.

Sources/Logging/Logging.swift
weissi: +1, often also called `Storage` but that's a nit :).

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,41 @@ import WASILibc
 /// logger.info("Hello World!")
 /// ```
 public struct Logger {
-    @usableFromInline
-    var handler: LogHandler
+    /// A private property to hold the boxed `LogHandler`.
```
weissi, line 42: as a unit test, you can put `XCTAssertEqual(MemoryLayout<UnsafeRawPointer>.size, MemoryLayout<Logger>.size)`. This should pass once this is complete (and it won't pass today).

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,44 @@ import WASILibc
...
+        return handler.metadataProvider
     }
 
     internal init(label: String, _ handler: LogHandler) {
         self.label = label
-        self.handler = handler
+        self._handler = Storage(label: label, handler: handler)
+    }
+
+    public class Storage: @unchecked Sendable {
```
ayushi2103, line 72: The CI is failing because of @unchecked Sendable, I am not able to reproduce on Xcode, not sure how to fix it.

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,44 @@ import WASILibc
 /// logger.info("Hello World!")
 /// ```
 public struct Logger {
-    @usableFromInline
-    var handler: LogHandler
+    /// A private property to hold the boxed `LogHandler`.
+    private var _handler: Storage
```
FranzBusch, line 43: Let's rename this and also add the inline annotation back. This is going to be important for performance reasons.
```suggestion
    @usableFromInline
    private var _storage: Storage
```

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,44 @@ import WASILibc
...
 
     /// An identifier of the creator of this `Logger`.
     public let label: String
+    
+    /// A computed property to access the `LogHandler`.
+    public var handler: LogHandler {
```
FranzBusch, line 49: Same here for perf reasons.
```suggestion
    @inlnable
    public var handler: LogHandler {
```

Sources/Logging/Logging.swift
FranzBusch, line 72: The below should hopefully fix this. Can we also move the class to top of this struct. So that it is the first line of code inside the structs body.
```suggestion
    @usableFromInline
    public final class Storage: @unchecked Sendable {
```

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,44 @@ import WASILibc
...
 
     internal init(label: String, _ handler: LogHandler) {
         self.label = label
-        self.handler = handler
+        self._handler = Storage(label: label, handler: handler)
+    }
+
+    public class Storage: @unchecked Sendable {
+        var label: String
+        var handler: LogHandler
```
FranzBusch, line 74: Can we add `@usableFromInline` to both of these properties and please copy the comment over from the `Logger` properties.

Sources/Logging/Logging.swift
ayushi2103: `@usableFromInLine` can only be applied to internal or declaration, should I change `_storage` from private?

weissi: @ayushi2103 Yes please! You already named it correctly, the `_` signifies "this is internal but should be private". And `public var _foo` means "this is public but should be private". 

Especially for `internal var _foo` it's almost always because of `@inlinableFromPublic`.

Sources/Logging/Logging.swift
weissi: This needs to be `internal class Storage` and then you also don't need the `final` but I would still recommend putting the `final`.

Also: A comment is warranted: `Storage` isn't actually `Sendable` but the compiler can't understand that we're using it with copy-on write... So it's never mutated unless we're making a copy. This is safe but compiler doesn't know. This makes it extra crucial that this is _not_ public, or else the `Sendable` conformance would be wrong.

I would write

```
internal final class Storage: @unchecked /* and not actually */ Sendable /* but safe if only used with CoW */ {
 ...
}
```

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -1053,6 +1053,11 @@ class LoggingTest: XCTestCase {
         LoggingSystem.bootstrap(StreamLogHandler.standardOutput, metadataProvider: .exampleMetadataProvider)
         LoggingSystem.bootstrap(StreamLogHandler.standardError, metadataProvider: .exampleMetadataProvider)
     }
+    
+    func testLoggerSize() {
+        let expectedSize = MemoryLayout<Logger.Storage>.size + MemoryLayout<String>.size
```
weissi, line 1058:
```suggestion
        let expectedSize = MemoryLayout<UnsafeRawPointer>.size
```

We need the `Logger` to become pointer-sized. So we should pull the `String` into the `Storage` too

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -1053,6 +1053,11 @@ class LoggingTest: XCTestCase {
         LoggingSystem.bootstrap(StreamLogHandler.standardOutput, metadataProvider: .exampleMetadataProvider)
         LoggingSystem.bootstrap(StreamLogHandler.standardError, metadataProvider: .exampleMetadataProvider)
     }
+    
+    func testLoggerSize() {
```
weissi, line 1057:
```suggestion
    func testLoggerIsJustHoldingASinglePointer() {
```

Sources/Logging/Logging.swift
ayushi2103: @weissi The [log](https://ci.swiftserver.group/job/swift-log-swift51-prb/351/console) is failing with the below errors
 error: unknown attribute 'unchecked'
error: use of undeclared type 'Sendable'
Do I need to to conform Storage to Sendable? (I was having warnings without that)

Sources/Logging/Logging.swift
ayushi2103: @FranzBusch As the properties are internal, I am not able to use `@inlinable`.

Sources/Logging/Logging.swift
weissi: > error: unknown attribute 'unchecked'
> error: use of undeclared type 'Sendable'

That means the compiler is too old. Ignore any failures that appear in Swift 5.0 through 5.7. We only care about Swift 5.8, 5.9, 5.10 and Swift 6 at this point. Those versions do have `Sendable` and `@unchecked`.

Sources/Logging/Logging.swift
FranzBusch: If you mark them as `@usableFromInline` this will work

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,57 @@ import WASILibc
...
+            self.handler = handler
+        }
+    }
+    
+    @usableFromInline
+    internal var _storage: Storage
 
-    /// An identifier of the creator of this `Logger`.
-    public let label: String
+    internal var label: String {
```
FranzBusch, line 59:
```suggestion
    @inlinable
    internal var label: String {
```

Sources/Logging/Logging.swift
FranzBusch: @ayushi2103 can you add the conformance to Sendable by putting it into a `#if compiler(` directive. There should be other examples in this repo that use `#if`

weissi: @FranzBusch we don't care about 5.0...5.7 anymore, so I don't think we should `#if` things that work on 5.8+, right?

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,75 @@ import WASILibc
 /// ```swift
 /// logger.info("Hello World!")
 /// ```
+
 public struct Logger {
+    #if compiler(>=5.7)
+        /// Storage class to hold the label and log handler/
+        @usableFromInline
+        internal final class Storage: @unchecked /* and not actually */ Sendable /* but safe if only used with CoW */ {
```
FranzBusch, line 46: I think we can make this easier by just moving the conformance to the bottom and compiler guard it like this
```
#if compiler(>=5.7)
extension Logger.Storage: @unchecked Sendable {}
#endif
```

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,75 @@ import WASILibc
...
+            if !isKnownUniquelyReferenced(&_storage) {
+                _storage = Storage(label: label, handler: newValue)
+            } else {
+                _storage.handler = newValue
+            }
+        }
+    }
 
     /// The metadata provider this logger was created with.
     public var metadataProvider: Logger.MetadataProvider? {
```
FranzBusch, line 104: We should probably mark this as `@inlinable` as well

FranzBusch:
@ayushi2103 Can you update this PR since we just landed #299  which drops support for old Swift versions. This should mean you no longer have to add the conditional conformance.

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
+    internal var _storage: Storage
     @usableFromInline
-    var handler: LogHandler
+    internal var label: String {
+        get {
+            return _storage.label
+        }
+        set {
+            if !isKnownUniquelyReferenced(&_storage) {
+                _storage = Storage(label: newValue, handler: _storage.handler)
```
weissi, line 66: it is usually beneficial to create a `func copy() -> Storage` on storage that just copies all variables over then you can write this as

```
[...]
self._storage = self._storage.copy()
self._storage.label = newValue
[...]
```

Also missing explicit `self.`

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
+    @inlinable
+    public var handler: LogHandler {
+        get {
+            return _storage.handler
+        }
+        set {
+            if !isKnownUniquelyReferenced(&_storage) {
+                _storage = Storage(label: label, handler: newValue)
+            } else {
+                _storage.handler = newValue
```
weissi, line 83:
```suggestion
                self._storage.handler = newValue
```

Please add explicit `self.`s everywhere. It's extremely dangerous to avoid using them because it's just not clear that you're mutating a property on self, especially in a CoW type. The most crucial bit of information is whether we're modifying something local or a shared reference. `self....` make this obivous, omitting `self` makes the user guess -- particularly in diffs which often don't have enough information to see this.

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
+        }
+    }
 
-    /// An identifier of the creator of this `Logger`.
-    public let label: String
+    /// A computed property to access the `LogHandler`.
+    @inlinable
+    public var handler: LogHandler {
+        get {
+            return _storage.handler
```
weissi, line 77:
```suggestion
            return self._storage.handler
```

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
-    var handler: LogHandler
+    internal var label: String {
+        get {
+            return _storage.label
+        }
+        set {
+            if !isKnownUniquelyReferenced(&_storage) {
+                _storage = Storage(label: newValue, handler: _storage.handler)
+            } else {
+                _storage.label = newValue
```
weissi, line 68:
```suggestion
                self._storage.label = newValue
```

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
 /// ```swift
 /// logger.info("Hello World!")
 /// ```
+
```
FranzBusch, line 41:
```suggestion
```

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
 public struct Logger {
+    /// Storage class to hold the label and log handler
+    @usableFromInline
+    internal final class Storage {
+        @usableFromInline
+        var label: String
```
FranzBusch, line 47:
```suggestion
        var label: String

```

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
+        @usableFromInline
+        var handler: LogHandler
```
FranzBusch, line 49:
```suggestion
        var handler: LogHandler

```

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
+        @usableFromInline
```
FranzBusch, line 50:
```suggestion
        @inlinable
```

Sources/Logging/Logging.swift
```diff
@@ -38,21 +38,61 @@ import WASILibc
...
+            self.label = label
+            self.handler = handler
+        }
+    }
+    
+    @usableFromInline
+    internal var _storage: Storage
     @usableFromInline
-    var handler: LogHandler
+    internal var label: String {
```
FranzBusch, line 60: I think this needs to be `public` like before
```suggestion
    public var label: String {
```

Sources/Logging/Logging.swift
```diff
@@ -1330,6 +1370,7 @@ private final class WarnOnceBox {
 // MARK: - Sendable support helpers
 
 extension Logger.MetadataValue: Sendable {}
+extension Logger.Storage: @unchecked Sendable {}
```
FranzBusch, line 1373: Can we move this conformance to the `Storage` at the again instead of an extension and add a common that it is safe since we manually implement CoW
```suggestion
extension Logger.Storage: @unchecked Sendable {}
```

Sources/Logging/Logging.swift
ayushi2103: ```'@usableFromInline' attribute can only be applied to internal or package declarations, but 'label' is public```, getting this error. 
Should I remove ```@usableFromInline``` or keep it ```internal```?

FranzBusch: Yeah remove `@usableFromInline` and keep it public

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,65 @@ import WASILibc
...
+    public var label: String {
+        get {
+            return self._storage.label
+        }
+        set {
+            if !isKnownUniquelyReferenced(&self._storage) {
+                self._storage = self._storage.copy()
+            }
+            self._storage.label = newValue // mutation is fine now because we just copied
+        }
```
FranzBusch, line 74: This just had a getter and was never settable
```suggestion
        return self._storage.label
```

FranzBusch (approved):
This looks good to me! @weissi you want to take a last look as well before we merge?

Sources/Logging/Logging.swift
```diff
@@ -39,20 +39,57 @@ import WASILibc
...
+    public var handler: LogHandler {
+        get {
+            return self._storage.handler
+        }
+        set {
+            if !isKnownUniquelyReferenced(&self._storage) {
+                self._storage = Storage(label: self.label, handler: newValue)
+            } else {
+                self._storage.handler = newValue
+            }
```
weissi, line 80:
```suggestion
            if !isKnownUniquelyReferenced(&self._storage) {
                self._storage = self._storage().copy()
            }
            self._storage.handler = newValue
```

This avoids repeating the important bits of the code

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -1039,6 +1039,11 @@ class LoggingTest: XCTestCase {
         LoggingSystem.bootstrap(StreamLogHandler.standardOutput, metadataProvider: .exampleMetadataProvider)
         LoggingSystem.bootstrap(StreamLogHandler.standardError, metadataProvider: .exampleMetadataProvider)
     }
+
+    func testLoggerIsJustHoldingASinglePointer() {
+        let expectedSize = MemoryLayout<UnsafeRawPointer>.size
+        XCTAssertEqual(MemoryLayout<Logger>.size, expectedSize)
+    }
```
weissi, line 1046: We should also add a test that shows that CoW works. Something like

```
var logger1 = Logger(label: "foo")
logger1.logLevel = .error
var logger2 = logger1
logger2.logLevel = .trace
XCTAssertEqual(.error, logger1)
XCTAssertEqual(.trace, logger2)
```

weissi (commented):
Thank you, this is almost there. Two little suggestions

weissi (approved):
Amazing! Thank you so much, this will be quite the perf boost for certain things!!

PR #312, opened 2024-06-24
Disable existential any build setting

glbrntt:
Motivation:

On 5.8, if the experiment existential any setting is enabled and a consuming package holds an existential of a type from swift-log without an explicit 'any' then the build will fail.

Modifications:

- Disable the setting

Result:

Users of 5.8 aren't broken

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,12 @@
 import PackageDescription
 
 let swiftSettings: [SwiftSetting] = [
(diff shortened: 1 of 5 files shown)


FranzBusch (approved):

Lukasa:
Can I suggest adding this flag to CI on these platforms?

FranzBusch (approved):

Sokartecnologi (commented):
7

Issue #242, opened 2022-12-22
Provide Test Helpers Publicly


Mordil:
I have my [own test helpers for validating logging behavior in my library](https://github.com/Mordil/RediStack/blob/master/Tests/RediStackIntegrationTests/RedisLoggingTests.swift)

after writing some tests in this repo's test suite... the test helpers are much nicer than what I wrote.

It'd be nice if the helpers were made public in a `LoggingTestUtils` module to be imported in my test targets...

ktoso:
Rather, we wanted to provide the LogCapture libs, as distributed actors uses it: https://github.com/apple/swift-distributed-actors/blob/main/Sources/DistributedActorsTestKit/LogCapture.swift and cluster membership: https://github.com/apple/swift-cluster-membership/blob/main/Tests/SWIMTestKit/LogCapture.swift

I forgot what that got stuck on, was it you @glbrntt or @fabianfett who was looking into this or am I misremembering?

glbrntt:
gRPC does similar. I don't recall looking into making this a feature of swift-log though.

It would be great if we did have this though as it's a useful utility and annoying for folks to have to re-invent it each time they want it.

tomerd:
+1

Mordil:
Where are we at with this? Is this something that just needs a PR?

ktoso:
Just someone needs to do it, yes. If you're interested help here would be welcome.

It'd live in a LoggingTestKit the same as we do for metrics test kit.
You can base the implementation on what the distributed cluster does.

Mordil:
Awesome. I'll look into creating a submission next week while I take extra time off for the US holiday weekend

brimworks:
This seems to have languished, so I put together this proposal: https://github.com/apple/swift-log/pull/316

If I get the thumbs up, then I'll put together a PR with that implementation. Please provide feedback in the above PR about the approach.

Thanks!

kukushechkin:
As @samuelmurray pointed out, the package now has a new product [InMemoryLogHandler](https://github.com/apple/swift-log/blob/main/Sources/InMemoryLogging/InMemoryLogHandler.swift), that allows exploring the recorded entries through a public `.entries { get }` property.

PR #316, opened 2024-08-19
Add a testing log handler proposal

brimworks:
Proposal for implementing a testing log handler which is provided by swift-log. For additional context, please see this github issue: https://github.com/apple/swift-log/issues/242

### DIFF
--- a/proposals/0002-testing-log-handler.md
+++ b/proposals/0002-testing-log-handler.md
@@ -1,0 +1,112 @@
+# Test Log Handler
+
+Authors: [Brian Maher](https://github.com/brimworks), [Konrad 'ktoso' Malawski](https://github.com/ktoso)
+
+## Introduction
+
+Software operators rely on logging to provide insight into what is being done,
+therefore, testing log output is an important facet.
+
+## Motivation
+
+A variety of implementations have cropped up organically, but having a single
+test log handler to build off of will reduce the re-write churn.
+
+### Existing Implementations
+
+* [swift-log-testing](https://github.com/neallester/swift-log-testing)
+* [swift-distributed-actors/LogCapture](https://github.com/apple/swift-distributed-actors/blob/main/Sources/DistributedActorsTestKit/LogCapture.swift)
+* [swift-cluster-membership/LogCapture](https://github.com/apple/swift-cluster-membership/blob/main/Tests/SWIMTestKit/LogCapture.swift)
+* [RediStack/RedisLoggingTests](https://github.com/swift-server/RediStack/blob/master/Tests/RediStackIntegrationTests/RedisLoggingTests.swift)
+* [swift-log/TestLogger](https://github.com/apple/swift-log/blob/main/Tests/LoggingTests/TestLogger.swift#L52)
+
+
+## Features
+
+A testing log handler needs the following features:
+* The ability to capture log entries at the start of a unit test.
+* The ability to assert the captured log entries contain a matching message.
+* Work correctly in the face of parallel test execution.
+* Print log messages to the console if the log entry does not match a log entry
+  that is being asserted.
+* Bootstrap the test log handler as early as possible so we don't miss any
+  `Logger()` creations which may happen during setup.
+
+## Proposed Solution
+
+A global function to bootstrap the test log handler:
+
+```swift
+TestingLogHandler.bootstrap()
+```
+
+A way to create a new "container" which will capture all log messages that match
+a filter. Note that multiple concurrent "container"s need to collect duplicate
+log messages, since tests may be ran in parallel.
+
+To create a new "container" we can have a function that takes a filter parameter
+and returns the new log container:
+
+```swift
+let container = TestingLogHandler.container { logMessage in
+    // Ideally, you would be as specific as possible in your filtering!
+    return logMessage.level == Logger.Level.debug &&
+        logMessage.label == "Test" &&
+        logMessage.message.description == "Matched"
+}
+// Emit a log that matches the filter:
+Logger(label: "Test").debug("Matched")
+
+// Emit a log that does NOT match the filter:
+Logger(label: "Test").debug("Printed to console")
+
+// Assert that the log message was found:
+XCTAssertFalse(container.messages.isEmpty)
+```
+
+It is recommended to create a separate container for each log assertion so that
+you can simply check that the messages in a container are non-empty. If you use
+this pattern you also protect yourself from accidental log message matches that
+occur due to a test ran in parallel. An accidental match could cause a failing
+test to spuriously succeed, but it would never cause a successful test to
+spuriously fail.
+
+If a log message is emitted that does **not** match any container's filter, it
+will be printed to the console in the same way as the default built-in log
+handler.
+
+In order to make filtering easier, the `LogMessage` struct contains a `match`
+function that takes the following optional parameters:
+* label: `String?`
+* level: `Logger.Level?`
+* message: `Regex<AnyRegexOutput>?`
+* metadata: `[(Logger.Metadata.Element) -> Bool]`
+* source: `Regex<AnyRegexOutput>?`
+* file: `Regex<AnyRegexOutput>?`
+* function: `String?`
+* line: `UInt?`
+
+Only if all specified parameters match will this function return true.
+
+In order to make debugging easier, the `LogMessage` struct will implement
+`CustomStringConvertible` which causes the log message to be formatted in the
+default built-in log handler format.
+
+Additionally, each field of the log message is public and can be used for
+custom filtering:
+* timestamp: `Date`
+* label: `String`
+* level: `Logger.Level`
+* message: `Logger.Message` (note that this is also `CustomStringConvertible`)
+* metadata: `Logger.Metadata`
+* source: `String`
+* file: `String`
+* function: `String`
+* line: `UInt`
+
+Note that the log handler implementation must maintain "weak" pointers to all
+the containers so as to avoid memory leaks.
+
+Also note that the `LogHandler.logLevel` is **always** `.trace`, otherwise log
+messages that are at to low of a level will not have a chance to be captured
+due to the `if self.logLevel <= level` guard on the `Logging.log()` method.


proposals/0002-testing-log-handler.md
```diff
@@ -0,0 +1,112 @@
...
+Logger(label: "Test").debug("Printed to console")
+
+// Assert that the log message was found:
+XCTAssertFalse(container.messages.isEmpty)
+```
+
+It is recommended to create a separate container for each log assertion so that
+you can simply check that the messages in a container are non-empty. If you use
+this pattern you also protect yourself from accidental log message matches that
+occur due to a test ran in parallel. An accidental match could cause a failing
```
ktoso, line 70: How so? The testing log handler is global, we may get multiple containers but it concurrently logging from various tests/threads, won't they end up in all currently active containers?

ktoso (commented):
Overall looks good to me, I think we could prepare a PR with the APIs / implementation. This has come up frequently and has been implemented independently in at least 3-4 projects I know of. We can fine tune the behavior once a PR with impl is present :)

proposals/0002-testing-log-handler.md
```diff
@@ -0,0 +1,112 @@
...
+  that is being asserted.
+* Bootstrap the test log handler as early as possible so we don't miss any
+  `Logger()` creations which may happen during setup.
+
+## Proposed Solution
+
+A global function to bootstrap the test log handler:
+
+```swift
+TestingLogHandler.bootstrap()
```
FranzBusch, line 40: Why don't we call the regular bootstrap method here? Furthermore, I think when we do this we should also make the init of `Logger` finally public that just takes a `LogHandler` so APIs that accept `Logger`s can be tested by just creating a `TestingLogHandler`, creating a logger with it and then calling the API with that logger.

proposals/0002-testing-log-handler.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+A way to create a new "container" which will capture all log messages that match
```
FranzBusch, line 43: I am not sure I follow the `container` principle yet. Should the testing log handler start by capturing all logs and metadata and then provide easy assertion APIs in the end? Potentially a connection to `swift-testing` that makes that as easy as possible would be great.

FranzBusch (commented):
Thanks for working on this! I agree that we should have a way for testing logging in this package. I am not yet sure I agree with the `container` principle outlined here instead of just having a "recorder" testing log handler that makes it easy to assert on what it recorded.

proposals/0002-testing-log-handler.md
brimworks: >  The testing log handler is global, we may get multiple containers but it concurrently logging from various tests/threads, won't they end up in all currently active containers?

Yes that is correct understanding. Maybe I need to elaborate a bit more here... but the scenario that I'm thinking of is where test A causes log message X to be printed and test B also causes log message X to be printed. If test A and B are ran such that the log containers for test A and B are both created before log message X is printed, and the log message X is printed before test A and B assert the container is non-empty, there could be multiple messages in the container. Therefore asserting that there is exactly one message could lead to spurious errors.

proposals/0002-testing-log-handler.md
brimworks: > Why don't we call the regular bootstrap method here?

As I recall, the `LoggingSystem.bootstrap()` must be called exactly once, so my thought was to encapsulate the logic for ensuring it is initialized exactly once into the `LoggingSystem.bootstrap()` method, but perhaps I am mistaken in my understanding?

> Furthermore, I think when we do this we should also make the init of Logger finally public that just takes a LogHandler so APIs that accept Loggers can be tested by just creating a TestingLogHandler, creating a logger with it and then calling the API with that logger.

I'm not opposed to that change, however I have found that it isn't to uncommon to find code which does not expose a way to "inject" the `Logger` instance (or trying to "inject" it with our custom log handler is rather onerous).

Overall, I do like the idea of making the `Logger` constructor that takes a `LogHandler` instance as public, but I'd like to also have a way to capture logs too.

Mordil: Honestly, I feel this situation needs to be resolved in a separate proposal where the bootstrap method is relaxed, perhaps only in debug mode.

proposals/0002-testing-log-handler.md
Mordil: I think I'm more in favor of giving more powerful assertion capabilities that solve the filtering, rather than having a robust filtering API, because the former is less work for the library to build, test, and maintain, than the filtering capabilities of the log handler.

It's also more familiar territory for those looking for test utilities, as it's a bit more discoverable due to the fact you're reaching for these tools as you're drafting your assertions rather than when you're in the mindset of bootstrapping the test (at least to me)

brimworks:
So, I started putting together an example implementation, and in the process I'm refactoring the existing `swift-log` tests so they don't use `LoggingSystem.bootstrapInternal()` since multiple calls to `LoggingSystem.boostrap()` triggers the "Logging.swift:617: Precondition failed: logging system can only be initialized once per process." error.

...and low and behold, the `LoggingTest testMultiplex` test is failing due to the fact that some extra medatadata (`["example": example-value]`) is getting injected as a side-effect of another concurrently executing test.

So, whatever solution we come up with needs to be a bit less fragile  in the face of concurrent test execution than the current solution. Note that I intend to do minimal changes to the current test solution so we can be assured that a bug in our test logger doesn't get hidden. Additionally, it will allow us to iterate on the test logger implementation provided by this proposal without impacting the other unit tests. Longer term we can see about removing that redundancy, but I'd rather do that as a separate PR.

I'll update this PR with my proposal soon.

brimworks:
PR has been updated with a proposed implementation. I also refactored the tests so they rely on `bootstrapInternal()` a bit less, although I couldn't completely remove the usage of this since some of the tests are testing the interactions at a "global" level.

Overall, I'd suggest that we minimize the usage of `bootstrapInternal()` though.

proposals/0002-testing-log-handler.md
brimworks: Thanks for the feeback @Mordil!

> because the former (more powerful assertion capabilities) is less work for the library to build, test, and maintain, than the filtering capabilities of the log handler.

Really? I would think it would be the opposite. More specifically, any assertion capability needs to have the exact same logic as the filtering capabilities... since an assertion is really just a fancy "filter" that not only returns true/false built in the event of failure it would print a nice debug message to assist.

> It's also more familiar territory for those looking for test utilities, as it's a bit more discoverable due to the fact you're reaching for these tools as you're drafting your assertions rather than when you're in the mindset of bootstrapping the test (at least to me)

Maybe the real problem is with the names I have chosen. Instead of having a "container" that has a pre-filter as I have done so far. What if we reduce the capabilities and instead have the exact same API but call this a "future log assertion"... and then later in the test you simply assert that this "future log assertion" got called (and remove the capability of introspecting any log messages that got captured). This would be similar to many "mock" frameworks which allow one to validate that certain methods on the "mock" got called. Do you think that would be more intuitive and discoverable?

proposals/0002-testing-log-handler.md
brimworks: BTW, @FranzBusch when putting together the sample implementation I noticed that there already is `Logger()` constructors (aka `init()`) that allow one to specify a log factory, so it looks like this comment is already done:

>  Furthermore, I think when we do this we should also make the init of Logger finally public that just takes a LogHandler so APIs that accept Loggers can be tested by just creating a TestingLogHandler, creating a logger with it and then calling the API with that logger.

@Mordil  said:

> Honestly, I feel this situation needs to be resolved in a separate proposal where the bootstrap method is relaxed, perhaps only in debug mode.

Perhaps... but I do think that if we have a reasonably good test logger which allows one to do initialization exactly once, then there is no need to allow for multiple initialization.

Actually, one way to get around the multiple initialization problem is to specify a log factory that delegates to an implementation which can be changed... actually, maybe I should use that idea in the current log tests since I'm trying to allow the current tests to be minimally modified (which means supporting multiple log handlers).

proposals/0002-testing-log-handler.md
ktoso: > I propose only non-matching log messages will be printed to the console.

I don't think this is right, at least not by default; A "log capture" log handler should capture all logs, then we assert on those (or filter out and are able to eagerly drop some), but we should not cause prints to happen.

If an assertion fails, there should be a way to "there was an error (the error); dump all the logs captured". This is how I use logging of _vast_ quantities of logs, and I get the "important" logs dumped when an assertion has failed.

Sources/Logging/TestingLogHandler.swift
```diff
@@ -0,0 +1,298 @@
+import Foundation
```
ktoso, line 1: (nitpick before we merge this eventually someday: needs license header)

proposals/0002-testing-log-handler.md
brimworks: > I don't think this is right, at least not by default; A "log capture" log handler should capture all logs, then we assert on those (or filter out and are able to eagerly drop some), but we should not cause prints to happen.

Let me tell you a story about when our team first attempted to add log assertions to our code base. Initially we added a dependency on [neallester/swift-log-testing](https://github.com/neallester/swift-log-testing) and we [followed the example](https://github.com/neallester/swift-log-testing/blob/master/Tests/SwiftLogTestingTests/ExampleTests.swift#L27-L34):

```swift
override func setUp() {
        TestLogMessages.bootstrap() 
}
```

We added in our log assertions and everything was great... until our CI pipeline broke and we realized that **none** of our log messages got printed, and since the logs were existing with signal 11 we had no idea what the problem was. In fact, this different design (print non-captured log messages) was partially motivated by this inoculating experience.

Therefore, I **highly** recommend that whatever final approach we land on will continue to default to printing logs at least when running tests that are not trying to use the log testing framework.

> If an assertion fails, there should be a way to "there was an error (the error); dump all the logs captured". This is how I use logging of vast quantities of logs, and I get the "important" logs dumped when an assertion has failed.

What if the code which performs the log assertion is never executed? I can think of many scenarios where I have experienced this: exception is thrown, process segfaults, etc.

> But I don't think "passed" execution should be logging anything by default

In the current design, if a log **matches** any container, then the message is not logged. Maybe I didn't make that part clear?

> in fact, tests which print much output are an anti pattern imho and stdout should be empty if nothing went wrong

Agreed... and if you use the test logging framework I propose, then any log message that matches an "expectation" won't be printed.

What do you think of this idea? Instead of having a container with a pre-filter as I currently have designed, what if I borrow some ideas from [java EasyMock](https://easymock.org/user-guide.html). More specifically, what if you could define a set of "expectations" for log messages, and then at the end of your test you can "verify" these expectations.

So, a typical usage might look like this:

```swift
class TestMyLogs: XCTestCase {
    override func setUp() {
        TestingLogHandler.bootstrap() 
    }
    func testExample() {
        // Logger may be created before the verifier:
        let logger = Logger(label: "Example")
        let verifier = LogVerifier()
        verifier.expect(label: "Example", level: .debug, message: Regex("^Message$"))
        logger.debug("Message")
        // Performs assertion(s) on all expectations. If any expectation fails, the expectation(s)
        // that fail are printed in the assertion failure. This call also resets all expectations.
        verifier.verify()
        // This call would fail, with an error indicating that there are no expectations set.
        verifier.verify()
    }
}
```

proposals/0002-testing-log-handler.md
Mordil: Obviously, I'll use whatever is settled on, and I think there's merit to both approaches.

My main feedback is just try to find the solution that requires the least amount of effort, has the least room for bugs, and perhaps has the most extensibility _outside_ of the module so this library doesn't need to see frequent bug/feature requests on this aspect.
---
I guess it's one of those "I'll know it if I see it" scenarios for me.

When I see the up-front filtering, I'm jumping (and this is where I may be wrong 😅) to the idea of an implementation that needs to track all of that, with a "nice" API for describing the filtering.

Vs. the assertions route is much more language primitives where we just store them, and then APIs for exposing those data structures and then relying on built-in methods like filter, map, etc.

In the former, I'd think you'd need more internal data structures, more internal logic. While in the latter, I'm thinking it's more of just a plain data structure of arrays.

proposals/0002-testing-log-handler.md
Mordil: Couldn't this behavior be a flag passed, so that people who need to debug/inspect their test executions more deeply can - and for those who don't, have their ideal clean stdout?

brimworks: > Couldn't this behavior be a flag passed, so that people who need to debug/inspect their test executions more deeply can - and for those who don't, have their ideal clean stdout?

In the current implementation I have proposed, just collect all the logs by specifying a pre-filter that selects all logs `{ true }`  and then you can get a "clean" stdout during that test run, and you can iterate over the container of logs collected to do all the normal assertion stuff you might want to do. I don't advocate for using it in that way, since I find it more useful to be selective up-front using a pre-filter, but the current implementation I have suggested supports both uses.

proposals/0002-testing-log-handler.md
brimworks: @Mordil  are you willing to put a little effort in and try out the implementation I have proposed? If you want to, you can have the pre-filter match all logs. When that container is active, no longs will be printed to stderr.

```swift
class TestMyLogs: XCTestCase {
    override func setUp() {
        TestingLogHandler.bootstrap() 
    }
    func testExample() {
        // Logger may be created before the container:
        let logger = Logger(label: "Example")
        let container = TestingLogHandler.container { true }
        logger.debug("Message")
        XCTAssertTrue(container.messages.find { $0.match(label: "Example", level: .debug) }, "Expected log message with label: Example, level: debug, but got \(container.messages.map { "\($0)" }.joined("\n"))")
    }
}
```

...and you could aways add extensions to the `TestingLogHandler.Container` if you want to make it easier to write assertions like the one above.

So, I think what I currently propose is the minimal code with maximal flexibility.

Personally, I wouldn't use it in this way, but everyone has their own preferences :).

proposals/0002-testing-log-handler.md
brimworks: > Actually, one way to get around the multiple initialization problem is to specify a log factory that delegates to an implementation which can be changed... actually, maybe I should use that idea in the current log tests since I'm trying to allow the current tests to be minimally modified (which means supporting multiple log handlers).

So, I went ahead and made an attempt at writing a log factory delegate here:

https://github.com/brimworks/swift-log/blob/testing-log-handler-proposal-delegated/Sources/Logging/DelegatedLogHandler.swift

Unfortunately, the current log unit tests rely on the copy on write behavior of structs, but my `DelegatedLogHandler` needs to be a class so we can keep track of any outstanding `LogHandler`s and update them (using a weak map). If I can find a way to replicate the copy on write behavior of structs in a class then I think I could get this to pass the current tests.

Mordil (commented):
I don't see anything wrong with this direction from my perspective

proposals/0002-testing-log-handler.md
FranzBusch: I am still not convinced this is the right thing to do in tests here. IMO if developers want to test their logs I don't think they should use bootstrapping of a test log handler but rather they should accept a `Logger` and we can create a `Logger(handler: TestLogerHandler()`. This would make this proposal a lot simpler since we don't need containers anymore.

Furthermore, it would push developers into the right pattern which is to accept a `Logger` at their APIs and potentially fallback to create a `Logger` if none is passed.

proposals/0002-testing-log-handler.md
FranzBusch: Similar to my comment above. I am not convinced we need this complicated container API if we should expect users to properly accept `Logger`s at the API level. We can then just store all log messages in the test log handler and assert on that. This should also work during parallel test execution

Sources/Logging/TestingLogHandler.swift
FranzBusch, line 1: I think we should introduce a new `LoggingTestKit` instead of adding it to the `Logging` module itself.

Sources/Logging/TestingLogHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+
+#if os(Linux)
+    import Glibc
+#else
+    import Darwin.C
+#endif
+
+public struct LogMessage: CustomStringConvertible {
```
FranzBusch, line 9: Naming NIT: What about `CapturedLogMessage`

Sources/Logging/TestingLogHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+    public let label: String
+
+    public let level: Logger.Level
+    public let message: Logger.Message
+    public let metadata: Logger.Metadata
+    public let source: String
+
+    public let file: String
+    public let function: String
+    public let line: UInt
```
FranzBusch, line 28: Can we make all of those `var`s. Since this is a `struct` this should be fine.

Sources/Logging/TestingLogHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+        self.message = message
+        self.metadata = metadata
+        self.source = source
+        self.file = file
+        self.function = function
+        self.line = line
+    }
+
+    // Easily check if log message matches certain characteristics:
+    @available(macOS 13.0, *)
```
FranzBusch, line 65: This should be available on all platforms we support

Sources/Logging/TestingLogHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+    public func match(
```
FranzBusch, line 66: I am not sure if we need this method at all. Why can't users just write

`#expected(capturedLogMessage.message == "foo")`

Sources/Logging/TestingLogHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+           self.line != line
+        {
+            return false
+        }
+
+        return true
+    }
+}
+
+public struct TestingLogHandler: LogHandler, Sendable {
```
FranzBusch, line 121: Naming NIT to align it with `TestMetrics` from `swift-metrics`
```suggestion
public struct TestLogHandler: LogHandler, Sendable {
```

Sources/Logging/TestingLogHandler.swift
```diff
@@ -0,0 +1,298 @@
...
+    public typealias Filter = (LogMessage) -> Bool
+
+    public final class Container: @unchecked Sendable {
+        // Guarded by queue:
+        private var _messages: [LogMessage] = []
+
+        private let queue: DispatchQueue
```
FranzBusch, line 128: Can we just use a `LockedValueBox` containing an array of messages?

Sources/Logging/TestingLogHandler.swift
brimworks: Excellent idea! Let me refactor the code into a new module.

Sources/Logging/TestingLogHandler.swift
brimworks: Wouldn't that mean this is no longer implementing the `Sendable` implicit protocol? ...or is that only `class`es that have that requirement?

FranzBusch: Correct, since this is a `struct` we can safely make everything mutable and still be `Sendable` since the compiler will copy the struct when mutated

Sources/Logging/TestingLogHandler.swift
brimworks: Like this `LockedValueBox`? https://github.com/apple/swift-distributed-tracing/blob/1.1.1/Sources/Instrumentation/Locks.swift#L129

I figured that `swift-log` would have a **no dependencies** policy, in case the swift distributed tracing code wants to use `swift-log`, but maybe I misunderstand your suggestion?

brimworks: ... maybe you mean this `RWLockedValueBox` which is not public (defined in `Logging.swift`)? If I move this code into a separate target, I don't think I will have access to that code, right? Did you want to make the code public?

Sources/Logging/TestingLogHandler.swift
brimworks: Typically you would be working with multiple log messages rather than a single captured one.

Perhaps the **real** desire here is to provide a better error message in the event that no single log message matches? For example, if you setup a container so that it only has log entries that match `message == "foo"`, then if no log messages matches, printing an error to the effect of `no log entry matches message == "foo"` would be desirable. Let me see if I can come up with a design for that.

Sources/Logging/TestingLogHandler.swift
brimworks: This was needed to support `Regex`

brimworks:
@FranzBusch I have been thinking over your comment over the course of the last week:

> I am still not convinced this is the right thing to do in tests here. IMO if developers want to test their logs I don't think they should use bootstrapping of a test log handler but rather they should accept a Logger and we can create a Logger(handler: TestLogerHandler(). This would make this proposal a lot simpler since we don't need containers anymore.
>
> Furthermore, it would push developers into the right pattern which is to accept a Logger at their APIs and potentially fallback to create a Logger if none is passed.

Although I agree with your sentiment (everyone should allow a `Logger` to be "injected" into their public APIs), I don't think this is a realistic requirement that we should be hoisting on anyone that wants to add some rigor to their logs by testing that the desired log output is achieved.

Therefore, I would rather ["embrace the genius of the and"](https://www.jimcollins.com/concepts/genius-of-the-and.html) and **encourage** users of the library to inject in their log handler manually, but still allow bootstrapping to a singular test log handler instance.

So, the way I take your feedback is that we need an easy way to inject the test log handler without requiring a bootstrap. Let me see if I can come up with a good solution to this problem.

Sources/Logging/TestingLogHandler.swift
brimworks: Done.

Sources/Logging/TestingLogHandler.swift
brimworks: Done

Sources/Logging/TestingLogHandler.swift
brimworks: Done

Sources/Logging/TestingLogHandler.swift
brimworks: Done

Sources/Logging/TestingLogHandler.swift
brimworks: I've updated the design proposal so that a new "assertion" method (under the covers it would really just wrap a log "container"). Let me know if you like that idea and I can add that wrapper.

brimworks:
> @brimworks said:
> So, the way I take your feedback is that we need an easy way to inject the test log handler without requiring a bootstrap. Let me see if I can come up with a good solution to this problem.

BTW, I changed the test case to use the "easily inject a TestLogHandler instance" way which avoids bootstrapping. Let me know if this addresses your concern.

Thanks!

brimworks:
For anyone else following along, the fate of this proposal is in limbo due to an open question about how to pass loggers around (should we be using task locals?). Based on the outcome of that decision, we may want to take a different tactic for testing (aka use task local logger handlers).

FranzBusch:
> For anyone else following along, the fate of this proposal is in limbo due to an open question about how to pass loggers around (should we be using task locals?). Based on the outcome of that decision, we may want to take a different tactic for testing (aka use task local logger handlers).

That's right but I think we could scope this down and still make progress with the PR. In my opinion, just creating one log handler that is recording all logged messages and allows us to easily assert what is recorded is valuable regardless of how we move forward with the logger passing. I would recommend that we focus on that for now to make this PR easier to move forward.

Sources/LoggingTestKit/TestLogHandler.swift
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        return true
+    }
+}
+
+public struct TestLogHandler: LogHandler, Sendable {
+    public typealias Filter = (CapturedLogMessage) -> Bool
+
+    public final class Container: @unchecked Sendable {
```
FranzBusch, line 139: I am not yet convinced that we need this type or concept. The thing that I am most worried about is that the container is a reference type. I can understand that after a test you would want to filter all the logged messages by a predicate and then assert on top of the filtered messages but this type is live-updated.

How strongly do you think that this type is needed?

Sources/LoggingTestKit/TestLogHandler.swift
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        public var factory: (String) -> TestLogHandler {
+            { label in
+                TestLogHandler(label: label, container: self)
+            }
+        }
+    }
+
+    private static let queue = DispatchQueue(label: "TestLogHandler")
```
FranzBusch, line 183: Can we instead use a lock to protect this state here? I don't recall of this project contains a locked value box but that or the `Mutex` from the stdlib would be preferable

Sources/LoggingTestKit/TestLogHandler.swift
```diff
@@ -0,0 +1,320 @@
...
+
+    // Guarded by queue:
+    private static var isInitialized = false
+    // Guarded by queue:
+    private static var logLevel: Logger.Level = .info
+    // Guarded by queue:
+    private static var _containers: [Weak<Container>] = []
+
+    public static func bootstrap() {
```
FranzBusch, line 192: I am leaning towards not offering this method and rather let the user in their tests do this. If they already have code that accepts a logger we just need to make the init public and they should be able to avoid bootstrapping completely.

Sources/LoggingTestKit/TestLogHandler.swift
```diff
@@ -0,0 +1,320 @@
...
+    public static func container(_ filter: @escaping Filter) -> Container {
+        let container = Container(filter)
+        queue.sync {
+            Self._containers.append(Weak(container))
+        }
+        return container
+    }
+
+    // Override the default global log level (defaults to .info)
+    public static func setLevel(_ level: Logger.Level) {
```
FranzBusch, line 212: Instead of a method can we expose the property publicly instead

FranzBusch (commented):
Thanks for working on this! I left some code comments inline. My biggest meta level feedback at this point is if we need the concept of `containers` at all. I already see great value by landing a log handler that just records all messages to get started with and then build it out from there.

Issue #321, opened 2024-09-13
Add Logger.withMetadata() function to create a logger with multiple metadata values


gabbifish:
### Expected behavior
Let’s say I have an existing Logger.Metadata struct I want to add to a Logger as metadata. It doesn’t seem like there is a simple function like `Logger.withMetadata(metadata: Logger.Metadata)` I can use. This means I have to iterate through every kv pair in my Logger.Metadata and add it to the logger with `logger[key] = value`, which is verbose. 

### Actual behavior
N/A

### Steps to reproduce
N/A

### If possible, minimal yet complete reproducer code (or URL to code)
N/A

### SwiftLog version/commit hash

`1.6.1/9cb486020ebf03bfa5b5df985387a14a98744537`

### Swift & OS version (output of `swift --version && uname -a`)
```
swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0
```

PR #322, opened 2024-09-13
Add withMetadata() convenience function for adding multiple metadata k/v pairs to a Logger

gabbifish:
Adds Add withMetadata() convenience function for adding multiple metadata k/v pairs to a Logger.

### Motivation:

Addresses #321

### Modifications:

Note that in the spirit of the `Logger[metadataKey: key] = value` subscript mutating a logger, this PR also mutates the logger in-place.

### Result:

Provides users a convenience function to add or modify multiple Logger.Metadata values in a Logger in place.


ktoso:

ktoso (approved):
Pretty good convenience func, I like it. Thanks for the PR

gabbifish:

ktoso:

Sources/Logging/Logging.swift
```diff
@@ -166,6 +166,17 @@ extension Logger {
         }
     }
 
+    /// Convenience function for adding multiple metadata items to a logger.
+    ///
+    /// - note: Logging metadata behaves as a value that means a change to the logging metadata will only affect the
+    ///         very `Logger` it was changed on.
+    @inlinable
+    public mutating func withMetadata(metadata: Logger.Metadata) {
```
glbrntt, line 174: Can I propose we spell this slightly differently? 

`with` is typically used in names that provide scoped access to something (`withUnsafePointer`, `withCheckedContinuation`, `withTaskGroup` etc.) so doesn't feel quite right here (although I appreciate it is the spelling used for loggers in other languages). 

How about `addMetadata`? Could we also drop the `metadata:` label to avoid the repetition?

glbrntt (changes requested):
Thanks @gabbifish – this looks handy! I've proposed a slightly different name but otherwise this looks great.

Sources/Logging/Logging.swift
FranzBusch: Wondering if `merging` is the right term here since we merge the new values over the old values if keys are in both metadata "bags". Could we also add a doc comment for the parameter that we pass here please

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -357,6 +357,25 @@ class LoggingTest: XCTestCase {
                                                    "nested-list": ["l1str", ["l2str1", "l2str2"]]])
     }
 
+    func testWithMetadata() {
```
FranzBusch, line 360: Could we add three more test cases:
1. Where the logger has some existing metadata and we add new metadata without any overlapping keys
2. Same as above but with overlapping keys
3. Adding more than once with overlapping keys

Sources/Logging/Logging.swift
t089: I think `merging` should be mutable and return a new `Logger`.
`merge` would be the `mutable` counterpart.

glbrntt: > I think `merging` should **not** be mutable and return a new `Logger`. `merge` would be the `mutable` counterpart.

Agreed. `merge` is the correct spelling for this context. If we do this we should allow the caller to specify how to handle non-unique keys too. `Dictionary` has a spelling for this: https://developer.apple.com/documentation/swift/dictionary/merge(_:uniquingkeyswith:)-m2ub

t089: since `Metadata` is actually a `Dictionary`, this is already available as `logger.metadata.merge(_:uniquingKeysWith:)` 🤷‍♂️

gabbifish: I do think `mergeMetadata`, paired with the dropping a required `metadata` argument label, is the clearest option here--will proceed with that.

Tests/LoggingTests/LoggingTest.swift
gabbifish: Added to existing `testMergeMetadata` function.

FranzBusch (commented):
I know this might open up a larger question since @t089 rightly pointed out that `Metadata` really is just a type alias for `Dictionary<String: MetadataValue>` what if we expose the metadata that a logger has via a `var` so that users could just call `metadata.merge` via the standard method on dictionary.

@weissi @ktoso do you know if there is historical reasons why we don't expose the metadata on the logger? `LogHandler` has it as a `get/set` requirement on the protocol so it seems very easy to wire through safely.

gabbifish:
I would anticipate there are some concerns about exposing the logger handler's internal state so directly? Having a simple function call for accessing these functions provides a simple, easier-to-secure abstraction.

gabbifish:
One more thought: Exposing `metadata` directly means we do get a mix of APIs for interactions with metadata. Users now will have two options for setting metadata--`logger[metadataKey: "something"] = "else"` and `logger.metadata["something"] = "else"`. I suppose that would be a departure from the simple, single convenience subscript approach taken previously. This enables code that could use a mix of the two setters above (or their corresponding getters). Maybe that isn't a bad thing, but my personal take is that it could be confusing UX and promote less consistent logger usage.

weissi:
> I know this might open up a larger question since @t089 rightly pointed out that `Metadata` really is just a type alias for `Dictionary<String: MetadataValue>` what if we expose the metadata that a logger has via a `var` so that users could just call `metadata.merge` via the standard method on dictionary.
> 
> @weissi @ktoso do you know if there is historical reasons why we don't expose the metadata on the logger? `LogHandler` has it as a `get/set` requirement on the protocol so it seems very easy to wire through safely.

Yes, that's a performance reason. The `LogHandler` doesn't need to _store_ the metadata as a dictionary, it can use whatever data structure it sees fit. It needs to be able to _accept_ it as `Dictionary<...>` though.

weissi:
> One more thought: Exposing `metadata` directly means we do get a mix of APIs for interactions with metadata. Users now will have two options for setting metadata--`logger[metadataKey: "something"] = "else"` and `logger.metadata["something"] = "else"`. I suppose that would be a departure from the simple, single convenience subscript approach taken previously. This enables code that could use a mix of the two setters above (or their corresponding getters). Maybe that isn't a bad thing, but my personal take is that it could be confusing UX and promote less consistent logger usage patterns.
> 
> I guess this already exists today, though? With `logger.handler.metadata`? the suggestion above for `logger.metadata` would just add a third path.

Yes, this was also a reason we do _not_ offer `Logger.metadata`. Properties should also have _O(1)_ access and that's impossible to achieve because the `LogHandler` could be storing the metadata in a different data structure. Or more likely, it could have multiple sources of metadata that it never wants to merge.

weissi:
> > One more thought: Exposing `metadata` directly means we do get a mix of APIs for interactions with metadata. Users now will have two options for setting metadata--`logger[metadataKey: "something"] = "else"` and `logger.metadata["something"] = "else"`. I suppose that would be a departure from the simple, single convenience subscript approach taken previously. This enables code that could use a mix of the two setters above (or their corresponding getters). Maybe that isn't a bad thing, but my personal take is that it could be confusing UX and promote less consistent logger usage patterns.
> > I guess this already exists today, though? With `logger.handler.metadata`? the suggestion above for `logger.metadata` would just add a third path.
> 
> Yes, this was also a reason we do _not_ offer `Logger.metadata`. Properties should also have _O(1)_ access and that's impossible to achieve because the `LogHandler` could be storing the metadata in a different data structure. Or more likely, it could have multiple sources of metadata that it never wants to merge.

So yes, please keep _not_ exposing `Logger.metadata`, that wouldn't be good for performance or extensibility.

FranzBusch (approved):
Thanks for the context @weissi. This looks good to me. @glbrntt  can you give your thumbs up as well then we can merge it.

Sources/Logging/Logging.swift
```diff
@@ -166,6 +166,20 @@ extension Logger {
...
 
+    /// Convenience function for merging multiple metadata items into a Logger's existing metadata.
+    ///
+    /// - parameters:
+    ///    - metadata: The metadata to merge into a Logger's existing metadata.
+    ///
+    /// - note: Logging metadata behaves as a value that means a change to the logging metadata will only affect the
+    ///         very `Logger` it was changed on.
+    @inlinable
+    public mutating func mergeMetadata(_ metadata: Logger.Metadata) {
```
glbrntt, line 177: Can we add the `uniquingKeysWith` parameter as well [à la Dictionary](https://developer.apple.com/documentation/swift/dictionary/merge(_:uniquingkeyswith:)-m2ub)?

Sources/Logging/Logging.swift
```diff
@@ -166,6 +166,20 @@ extension Logger {
...
+        metadata.forEach { (key, value) in
+            self.handler[metadataKey: key] = value
+        }
```
glbrntt, line 180: I suspect in many (all?) handler implementations this will acquire and release a lock for each value we update. Can we call `merge` directly on the `handler.metadata` to avoid this? i.e. `self.handler.metadata.merge(metadata)`

PR #323, opened 2024-09-25
Fix logger for cocoapods

ArielDemarco:


ArielDemarco:
PR to wrong fork :) closing

PR #324, opened 2024-09-26
Fix Android build

lhoward:
Building with the Android SDK, stdout and stderr are not Nullable


hyp:
This is needed for Swift < 6, for Swift 6 + you need https://github.com/apple/swift-log/pull/325

lhoward:
Maybe I'll just wait for a Swift 6 Android toolchain then.

PR #328, opened 2024-10-29
[CI] Migrate to GHA

FranzBusch:
# Motivation

We want to modernize our CI.

# Modification

This PR migrates to GH actions and turns on all the right flags.

# Result

Modern CI.


Tests/LoggingTests/LocalLoggingTest.swift
```diff
@@ -114,15 +140,15 @@ private struct Struct1 {
...
+        c.logLevel = .info  // only effects from this point on
         c.logger.info("Struct2::doSomething")
         self.doSomethingElse(context: c)
         c.logger.debug("Struct2::doSomething::end")
     }
 
     private func doSomethingElse(context: Context) {
         var c = context
-        c.logLevel = .debug // only effects from this point on
+        c.logLevel = .debug  // only effects from this point on
```
ktoso, line 151: I wonder if there's a way to prevent those, it's somewhat silly to insert a double space... maybe there's a bug for it already?

ktoso: Filed an issue about it https://github.com/swiftlang/swift-format/issues/868

ktoso (approved):
I wish the reformat was a single commit, it's very noisy to review. But looks okey AFAICS

ktoso:
Ok I think we're green on GH Actions now. I changed all the required checks as well.

PR #334, opened 2024-11-08
Codable extension for Logger.MetadataValue

bwdmr:
This Codable extension for Logger.MetadataValue provides encoding and decoding functionality, mapping MetadataValue cases (string, stringConvertible, dictionary, array) into JSON-compatible formats using the associated type and value keys

* Container Encoding:
  * Each case of MetadataValue is encoded with an explicit type (ValueType) and corresponding value.
  * stringConvertible is encoded using .description since we need it in a String format.
* Container Decoding
  * When decoding, ValueType is first extracted to determine the case.
  * For stringConvertible, the String value is decoded and then assigned back to stringConvertible.
  * Dictionary and array cases are decoded as Logger.Metadata and [Logger.MetadataValue] respectively.

### DIFF
--- a/Sources/Logging/MetadataProvider.swift
+++ b/Sources/Logging/MetadataProvider.swift
@@ -99,3 +99,63 @@ extension Logger.MetadataProvider {
         }
     }
 }
+
+
+extension Logger.MetadataValue: Codable {
+    private enum CodingKeys: String, CodingKey {
+        case type
+        case value
+    }
+    
+    private enum ValueType: String, Codable {
+        case string
+        case stringConvertible
+        case dictionary
+        case array
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        
+        switch self {
+        case .string(let stringValue):
+            try container.encode(ValueType.string, forKey: .type)
+            try container.encode(stringValue, forKey: .value)
+            
+        case .stringConvertible(let customValue):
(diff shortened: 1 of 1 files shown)


ktoso:
I think that's fine but I'd like to request a test (can pull in foundation and use JSON), so that we check this is covered.

FranzBusch:
I am not yet convinced this is the right thing to do. Encoding enums is always a bit tricky and results in interesting JSON formats. Not saying we shouldn't do this but I would like to understand the motivation a bit more here.

bwdmr:
generally meant for metadatavalue to be compatible with codable.
it was meant for cases such as string and string convertible

for enums would imply a codable constraint. enforcing a custom implementation each time one would rely on it.
i dont know how excessive this requirement is.

might be too much to ask. what do you guys think

FranzBusch:
> generally meant for metadatavalue to be compatible with codable. it was meant for cases like string
> 
> for enums would imply a codable constraint. enforcing a custom implementation each time one would rely on it. i dont know how excessive this requirement is.
> 
> might be too much to ask. what do you guys think

Thanks for explaining your intention. I remain unconvinced that this is the right thing to do. Specifically because the way we encode this is very custom to how the Swift type is structured. I also don't see this as something that is blocking people from writing their own backends that have to emit metadata values as JSON. We have many log handler backends that do this already and they often hand roll the JSON emission for better performance.

bwdmr:
conformity with codable is up to optimization of the implemtation due to performance critical role.

PR #345, opened 2025-01-07
Add swift-log-systemd.git to the list of available backends (#345)

lhoward:
systemd-based log integration

### Motivation:

Using native backend for storing log data

### Modifications:

Add swift-log-systemd to the list of available backends

### Result:

Another one option to use the package with


glbrntt (approved):
Thanks!

PR #346, opened 2025-01-12
Add swift-log-systemd has moved (#345)

lhoward:
swift-log-systemd has been merged into the umbrella swift-systemd package


ktoso (approved):

ktoso:
Thanks!

PR #351, opened 2025-02-18
Add SumoLogHandler to the list of available backends

vlaminck:
Add SumoLogHandler to the list of available backends

### Motivation:

A LogHandler for SumoLogic is available

### Modifications:

Updated the table in the README

### Result:

The README will have another backend LogHandler


ktoso:
We no longer maintain a list of backends in the readme, instead please list your package in the Swift Package Index: https://swiftpackageindex.com/search?query=swift-log

PR #360, opened 2025-04-21
Add new library link for PostgresNIO backend

atacan:
Example backend implementation using PostgresNIO

### Motivation:

Sharing a new backend implementation.

### Modifications:

only README.md

### Result:

A new row in the library table.


ktoso:
We no longer maintain a list of backends in the readme, instead please list your package in the Swift Package Index: https://swiftpackageindex.com/search?query=swift-log

PR #367, opened 2025-07-20
Append `.git` for Importing Dependencies.

lamtrinhdev:
As `Importing Dependencies` Guide at https://www.swift.org/documentation/package-manager/ .I would like to append `.git` for the link. It is minor but I think we should consistent (append .git) with https://www.swift.org/documentation/package-manager/ .

### Modifications:
+ Readme.md

### Result:
Consistent `Importing Dependencies` with https://www.swift.org/documentation/package-manager/

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ import PackageDescription
 let package = Package(
     name: "YourApp",
     dependencies: [
-        .package(url: "https://github.com/apple/swift-log", from: "1.6.0")
+        .package(url: "https://github.com/apple/swift-log.git", from: "1.6.0")
     ],
     targets: [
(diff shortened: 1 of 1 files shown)


lamtrinhdev:
Dear @FranzBusch ,

I have a minor update for `README.md` . When you have time, could you help me to review it?

Thanks,
Lam

czechboy0:
Hi @lamtrinhdev,

can you explain why it needs to change? SwiftPM handles the shorter link just fine. If you'd like to use the long link in your projects, you can - but why should it be prescribed for all adopters of Swift Log when it makes no functional difference?

PR #370, opened 2025-08-18
Add new best practices for structured logging and accepting loggers

FranzBusch:


Sources/Logging/Docs.docc/BestPractices/002-StructuredLogging.md
```diff
@@ -0,0 +1,96 @@
...
+Use hierarchical dot-notation for related fields:
+
+```swift
+// ✅ Good: Hierarchical keys
+logger.debug(
+    "Database operation completed",
+    metadata: [
+        "db.operation": "SELECT",
+        "db.table": "users",
+        "db.duration": "\(duration)ms",
```
glbrntt: Including units the value seems like a bad idea, I can only imagine it makes it harder to query for durations as you first need to parse out the numerical value.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,108 @@
...
+        logger.debug("Business logic completed")
+        return HTTPResponse(data: dbResult)
+    }
+}
+```
+
+#### Alternative: Accept logger through initializer when appropriate
+
+```swift
+// ✅ Acceptable: Logger through initializer for long-lived components
```
glbrntt, line 64: This example isn't very compelling because there's no obvious reason that the logger couldn't be passed in to `processJob`.

Maybe something like a connection pool would be a better example? E.g. long running background task where you can still reach in with a method-specific logger.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,108 @@
...
+    private func processJobInternal(_ job: Job, logger: Logger) async {
+        logger.debug("Processing job")
+        // Process with contextual logger
+    }
+}
+```
+
+#### Avoid: Libraries creating their own loggers
+
+Libraries might create their own loggers; however, this leads to two problems.
```
glbrntt, line 82: Might be worth adding a caveat: libraries should default to creating a no-op logger if the user didn't pass one in, that makes it easier to ensure that the logger is always propagated through the library.

FranzBusch: I intentionally didn't add this since I think this is a topic of contention right now. I personally think this is a bad pattern since I have seen many cases where the default parameter led to folks not passing in their logger and when debugging they didn't get log statements. I personally much rather prefer having a non-optional non-defaulted parameter since it requires users attention. However, I think there is design space here by leveraging structured concurrency more to avoid all these manual passings of loggers altogether.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
FranzBusch: Even a connection pool can use a request specific logger and really should if possible. I agree the example isn't good due to the job processing nature. I will just make this a run method instead.

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
```diff
@@ -47,21 +46,21 @@ unless it is a one-time (for example during startup) warning, that cannot lead
...
-- **Performance**: Should not significantly undermine production performance
-- **Content**: High-level operation overview, connection events, major decisions
+- **Usage**: May be enabled in some production deployments.
+- **Performance**: Should not significantly undermine production performance.
+- **Content**: High-level operation overview, connection events, major decisions.
 
 ##### Info Level
 - **Usage**: Reserved for things that went wrong but can't be communicated
-through other means like throwing from a method
+through other means like throwing from a method.
```
heckj:
```suggestion
through other means, like throwing from a method.
```

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
```diff
@@ -47,21 +46,21 @@ unless it is a one-time (for example during startup) warning, that cannot lead
...
 - **Examples**: Connection retry attempts, fallback mechanisms, recoverable
-  failures
-- **Guideline**: Use sparingly - not for normal successful operations
+  failures.
+- **Guideline**: Use sparingly - not for normal successful operations.
```
heckj:
```suggestion
- **Guideline**: Use sparingly - Don't use for normal successful operations.
```

Sources/Logging/Docs.docc/BestPractices/002-StructuredLogging.md
```diff
@@ -0,0 +1,96 @@
...
+## Overview
+
+Structured logging uses metadata to separate human-readable messages from
+machine-readable data. This practice makes logs easier to search, filter, and
+analyze programmatically while maintaining readability.
+
+### Motivation
+
+Traditional string-based logging embeds all information in the message text,
+making it difficult for automated tools to extract specific information.
```
heckj:
```suggestion
making it more difficult for automated tools to parse and extract.
```

Sources/Logging/Docs.docc/BestPractices/002-StructuredLogging.md
```diff
@@ -0,0 +1,96 @@
...
+Structured logging separates concerns: messages provide human context while
```
heckj:
```suggestion
Structured logging separates these concerns; messages provide human readable context while
```

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
...
+propagation of contextual metadata attached to the logger instance. This is
+especially important for distributed systems where correlation IDs and tracing
+information must flow through the entire request processing pipeline.
+
+### Example
+
+#### Recommended: Accept logger through method parameters
+
+```swift
+// ✅ Good: Logger passed through method parameters
```
heckj:
```suggestion
// ✅ Good: Pass the logger through method parameters.
```

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
...
+struct RequestProcessor {
+    func processRequest(_ request: HTTPRequest, logger: Logger) async throws -> HTTPResponse {
+        // Add metadata to the logger that every log statement should contain
```
heckj:
```suggestion
        // Add structured metadata that every log statement should contain.
```

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
...
+        var logger = logger
+        logger[metadataKey: "request.method"] = "\(request.method)"
+        logger[metadataKey: "request.path"] = "\(request.path)"
+        logger[metadataKey: "request.id"] = "\(request.id)"
+
+        logger.debug("Processing request")
+        
+        // Pass logger down to maintain metadata context
```
heckj:
```suggestion
        // Pass the logger down to maintain metadata context.
```

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
...
+    private func validateRequest(_ request: HTTPRequest, logger: Logger) throws -> ValidatedRequest {
+        logger.debug("Validating request parameters")
+        // Validation logic with same logger context
+        return ValidatedRequest(request)
+    }
+    
+    private func executeBusinessLogic(_ data: ValidatedRequest, logger: Logger) async throws -> HTTPResponse {
+        logger.debug("Executing business logic")
+        
+        // Further propagate to other services
```
heckj:
```suggestion
        // Further propagate the logger to other services.
```

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
...
+        logger.debug("Update about long running work")
+        // Execute some more long running work
+    }
+}
+```
+
+#### Avoid: Libraries creating their own loggers
+
+Libraries might create their own loggers; however, this leads to two problems.
+First, the user of the library can't inject their own loggers which means he has
```
heckj:
```suggestion
First, the user of the library can't inject their own loggers which means they have
```

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
...
+        let validatedData = try validateRequest(request, logger: logger)
+        let result = try await executeBusinessLogic(validatedData, logger: logger)
+        
+        logger.debug("Request processed successfully")
+        return result
+    }
+    
+    private func validateRequest(_ request: HTTPRequest, logger: Logger) throws -> ValidatedRequest {
+        logger.debug("Validating request parameters")
+        // Validation logic with same logger context
```
heckj:
```suggestion
        // Include validation logic that uses the same logger context.
```

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
```diff
@@ -47,21 +46,21 @@ unless it is a one-time (for example during startup) warning, that cannot lead
...
 ##### Debug Level  
-- **Usage**: May be enabled in some production deployments
-- **Performance**: Should not significantly undermine production performance
-- **Content**: High-level operation overview, connection events, major decisions
+- **Usage**: May be enabled in some production deployments.
+- **Performance**: Should not significantly undermine production performance.
+- **Content**: High-level operation overview, connection events, major decisions.
 
 ##### Info Level
 - **Usage**: Reserved for things that went wrong but can't be communicated
```
toffaletti, line 60: I've not seen "info" level being used for things that went wrong. I would consider warn and error levels for things that went wrong. I use info for information like version, configuration, etc. that would help to diagnose "what software am I debugging" type questions.

FranzBusch: This is guidance in particular for libraries which should avoid logging at levels higher than info to avoid spamming log systems; however, they still sometimes need to inform users that something might have went wrong. In this case, we decided that `info` is the right level for that.

DominikPalo:
The “For libraries” paragraph is a bit confusing right now, since it contains a sentence "Each level serves different purposes:"  in the middle of the paragraph. I think this sentence should be moved to the end of that paragraph, just before explaining different loggings levels.

I'm attaching a git patch fixing the issue:
[swift-log-14-29-48.patch](https://github.com/user-attachments/files/21855667/swift-log-14-29-48.patch)

heckj (approved):

ktoso:
@DominikPalo 

> The “For libraries” paragraph in 001-ChoosingLogLevels.md is a bit confusing right now, since it contains a sentence 

Please make a separate PR about this, this isn't PR about that document; let's keep the discussions and fixes separate.

Sources/Logging/Docs.docc/BestPractices/002-StructuredLogging.md
```diff
@@ -0,0 +1,96 @@
...
+logger.info(
+    "HTTP response",
+    metadata: [
+        "http.method": "POST",
+        "http.status": "201",
+        "http.path": "/api/users",
+        "http.duration": "\(duration)"
+    ]
+)
+```
```
ktoso, line 96: This doc LGTM;

Sources/Logging/Docs.docc/BestPractices/002-StructuredLogging.md
```diff
@@ -0,0 +1,96 @@
...
+
+```swift
+// ❌ Not structured - hard to parse programmatically
+logger.info("Accepted connection \(id) from \(peer), total: \(count)")
+logger.error("Database query failed after \(retries) retries: \(error)")
+```
+
+### Metadata key conventions
+
+Use hierarchical dot-notation for related fields:
```
ktoso, line 72: Yeah dots is a good recommendation I think... if some specific backend needs something else they can do what prometheus does in metrics for labels (nameAndLabelSanitizer) 👍

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
...
+storing them as instance variables. This practice ensures metadata (such as
+correlation IDs) is properly propagated down the call stack, while giving
+applications control over logging configuration.
+
+### Motivation
+
+When libraries accept loggers as method parameters, they enable automatic
+propagation of contextual metadata attached to the logger instance. This is
+especially important for distributed systems where correlation IDs and tracing
+information must flow through the entire request processing pipeline.
```
ktoso: The best practices listed are OK, however it somewhat pretty weird to call out tracing as motivation and not mention the way tracing _actually_ integrates with logging -- which isn't this pattern (regardless of opinions on status quo). 

I'd just drop "distributed tracing" from the motivation entirely if the examples are not going to address it at all, causes some confusion as it's not showing off how currently tracing actually integrates with loggers today.

ktoso (approved):
LGTM, please make two PRs next time when adding unrelated documents though

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
FranzBusch: I dropped tracing and just kept the correlation IDs part.

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
```diff
@@ -39,29 +38,30 @@ its logging environment.
 
 #### For libraries
 
-Libraries should use **info level or lower** (info, debug, trace). Each level
-serves different purposes:
+Libraries should use **info level or lower** (info, debug, trace).
```
czechboy0:
```suggestion
Libraries should use **info level or less severe** (info, debug, trace).
```

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -0,0 +1,100 @@
+# 003: Accepting loggers in libraries
+
+Accept loggers through method parameters to ensure proper metadata propagation.
```
czechboy0, line 3: Would you consider splitting this one into a separate PR to allow us to discuss more, without blocking the other changes? I have more thoughts here.

ktoso (approved):
still lgtm :-)

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
```diff
@@ -39,29 +38,30 @@ its logging environment.
 
 #### For libraries
 
-Libraries should use **info level or lower** (info, debug, trace). Each level
-serves different purposes:
+Libraries should use **info level or less severe** (info, debug, trace).
 
 Libraries **should not** log information on **warning or more severe levels**,
```
kukushechkin, line 43: What is the suggested design around this — libraries throw or return errors and it is application responsibility to log errors?

Sources/Logging/Docs.docc/BestPractices/002-StructuredLogging.md
kukushechkin: Are those considered equivalent?

```
logger.debug(
    "Database operation completed",
    metadata: [
        "db.operation": "SELECT",
        "db.table": "users",
        "db.duration": "\(duration)",
        "db.rows": "\(rowCount)"
    ]
)
```

and 

```
logger.debug(
    "Database operation completed",
    metadata: [
        "db": [
            "operation": "SELECT",
            "table": "users",
            "duration": "\(duration)",
            "rows": "\(rowCount)"
        ]
    ]
)
```

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
FranzBusch: Good question. We revisited this point recently a bit. Before this best practice we actually recommended libraries to only log below `info` level. That meant that they only had `trace` and `debug` logs which were often not logged. So in general libraries should surface errors by throwing from methods; however, sometimes there is a background task happening which a library doesn't want to teardown just because of an intermediate error. With the updated guidance libraries can now log that `info` level.

Sources/Logging/Docs.docc/BestPractices/002-StructuredLogging.md
FranzBusch: No they are not. They result in different metadata being generated.

FranzBusch: The second one also requires another dictionary allocation

PR #375, opened 2025-10-07
Erase file pointer types for default log handler on Android 

### LINKED ISSUE #374: Support Android 23
`swift-log` does not currently compile for Android 23, but only Android 24+.

This is due to the following code:
https://github.com/apple/swift-log/blob/1be2c17b808c404e31436d114adf448a3eba5872/Sources/Logging/Logging.swift#L1206-L1210

On Android 23 a file pointer is actually `UnsafeMutablePointer<FILE>` and not `OpaquePointer`, unlike newer API versions.

We are working on adding support for [Android version availability checks](https://github.com/swiftlang/swift/pull/84574), but until that is officially merged and **released**, we cannot use `swift-log` or any dependencies which use it, which is quite unfortunate.

@ktoso and I discussed that removing the default file logger for Android could be a solution. "Technically" android is just best effort supported right now, so you could argue that this is fine. 

Also, Android developers would probably bring their own logger implementation, which uses `<android/log.h>`, instead of just using to stdout, since by default, the Android system sends stdout and stderr output to /dev/null.

madsodgaard:
Introduces a C shim to erase the pointer type of stdio types, that are different across Android versions.

With this addition `swift-log` and any packages that use it can be compiled on Android 23+

Resolves #374

### DIFF
--- a/Sources/CSwiftLogAndroidSupport/include/CSwiftLogAndroidSupport.h
+++ b/Sources/CSwiftLogAndroidSupport/include/CSwiftLogAndroidSupport.h
@@ -6,7 +6,6 @@
 // These C functions will be called from Swift. They provide a stable
 // interface, hiding the pointer type differences between Android versions.
 
-// Get opaque pointers to standard streams.
 void *shim_stdout(void);
 void *shim_stderr(void);
 
--- a/Sources/CSwiftLogAndroidSupport/include/CSwiftLogAndroidSupport.h
+++ b/Sources/CSwiftLogAndroidSupport/include/CSwiftLogAndroidSupport.h
@@ -1,0 +1,18 @@
+#ifndef SWIFT_LOG_ANDROID_SUPPORT_H
+#define SWIFT_LOG_ANDROID_SUPPORT_H
+
+#include <stdio.h>
+
+// These C functions will be called from Swift. They provide a stable
+// interface, hiding the pointer type differences between Android versions.
+
+// Get opaque pointers to standard streams.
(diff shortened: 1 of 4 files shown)


Sources/CSwiftLogAndroidSupport/include/CSwiftLogAndroidSupport.h
```diff
@@ -0,0 +1,17 @@
+#ifndef SWIFT_LOG_ANDROID_SUPPORT_H
+#define SWIFT_LOG_ANDROID_SUPPORT_H
+
+#include <stdio.h>
+
+// These C functions will be called from Swift. They provide a stable
+// interface, hiding the pointer type differences between Android versions.
+
+void *shim_stdout(void);
```
weissi: ideally, given them a little more unique prefixes. Maybe `swift_log_android` or smth. The alternative is to make them `static inline` functions and define them in the `.h` file which means they'll be inlined entirely and won't get a symbol at all.

weissi: that way we can keep the `.c` file empty and we're not paying overhead. Example: https://github.com/apple/swift-nio/blob/dee7095f8d8a37a4cd165109768552163aa76458/Sources/CNIOLLHTTP/include/CNIOLLHTTP.h#L22-L26

weissi: Also, we can do this for all platforms, no need to special-case Android

PR #382, opened 2025-10-22
Migrate from XCTest to Swift Testing

kukushechkin:
Migrating `swift-log` from XCTest to Swift Testing

### Motivation

Swift Testing is a modern alternative to XCTest. Swift Server Ecosystem packages should serve as reference implementations for modern Swift features and best practices.

### Modifications

- 1-to-1 migration of test suites, tests, and assertions to Swift Testing syntax with XCTest import removed
- XCTest references removed from documentation
- `.trace` log level added to relevant tests
- `GlobalLoggerTest` suite serialized to handle global `LoggingSystem` access (XCTest runs tests serially by default; Swift Testing runs in parallel, so serializing this suite avoids requiring `swift test --no-parallel` across all CI and development environments)

### Result

Successfully executes 52 tests with `swift test --disable-xctest`.

### DIFF
--- a/Tests/LoggingTests/GlobalLoggingTest.swift
+++ b/Tests/LoggingTests/GlobalLoggingTest.swift
@@ -124,6 +124,96 @@ struct GlobalLoggerTest {
     }
 }
 
+/// MetadataProvider tests relying on the global state
+extension GlobalLoggerTest {
+    @Test func loggingMergesOneOffMetadataWithProvidedMetadataFromExplicitlyPassed() throws {
+        let logging = TestLogging()
+
+        LoggingSystem.bootstrapInternal(
+            { logging.makeWithMetadataProvider(label: $0, metadataProvider: $1) },
+            metadataProvider: .init {
+                ["common": "initial"]
+            }
+        )
+
+        let logger = Logger(
+            label: #function,
+            metadataProvider: .init {
+                [
+                    "common": "provider",
+                    "provider": "42",
+                ]
+            }
+        )
+
+        logger.log(level: .info, "test", metadata: ["one-off": "42", "common": "one-off"])
+
+        logging.history.assertExist(
+            level: .info,
+            message: "test",
+            metadata: ["common": "one-off", "one-off": "42", "provider": "42"]
+        )
+    }
+
+    @Test func loggerWithoutFactoryOverrideDefaultsToUsingLoggingSystemMetadataProvider() {
+        let logging = TestLogging()
+        LoggingSystem.bootstrapInternal(
+            { logging.makeWithMetadataProvider(label: $0, metadataProvider: $1) },
+            metadataProvider: .init { ["provider": "42"] }
+        )
+
+        let logger = Logger(label: #function)
+
+        logger.log(level: .info, "test", metadata: ["one-off": "42"])
+
+        logging.history.assertExist(
+            level: .info,
+            message: "test",
+            metadata: ["provider": "42", "one-off": "42"]
+        )
+    }
+
+    @Test func loggerWithPredefinedLibraryMetadataProvider() {
+        let logging = TestLogging()
+        LoggingSystem.bootstrapInternal(
+            { logging.makeWithMetadataProvider(label: $0, metadataProvider: $1) },
+            metadataProvider: .exampleProvider
+        )
+
+        let logger = Logger(label: #function)
+
+        logger.log(level: .info, "test", metadata: ["one-off": "42"])
+
+        logging.history.assertExist(
+            level: .info,
+            message: "test",
+            metadata: ["example": "example-value", "one-off": "42"]
+        )
+    }
+
+    @Test func loggerWithFactoryOverrideDefaultsToUsingLoggingSystemMetadataProvider() {
+        let logging = TestLogging()
+        LoggingSystem.bootstrapInternal(
+            { logging.makeWithMetadataProvider(label: $0, metadataProvider: $1) },
+            metadataProvider: .init { ["provider": "42"] }
+        )
+
+        let logger = Logger(
+            label: #function,
+            factory: { label in
+                logging.makeWithMetadataProvider(label: label, metadataProvider: LoggingSystem.metadataProvider)
+            }
+        )
+        logger.log(level: .info, "test", metadata: ["one-off": "42"])
+
+        logging.history.assertExist(
+            level: .info,
+            message: "test",
+            metadata: ["provider": "42", "one-off": "42"]
+        )
+    }
+}
+
 private struct Struct1 {
     private let logger = Logger(label: "GlobalLoggerTest::Struct1")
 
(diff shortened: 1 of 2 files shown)


Tests/LoggingTests/GlobalLoggingTest.swift
```diff
@@ -181,3 +194,78 @@ private struct Struct3 {
...
+            level: .info,
+            message: "test",
+            metadata: ["common": "one-off", "one-off": "42", "provider": "42"]
+        )
+    }
+}
+
+/// StreamLogHandler tests relying on the global state
+extension GlobalLoggerTest {
+    @Test func compileInitializeStandardStreamLogHandlersWithMetadataProviders() {
```
czechboy0: Since this is a compile-time only test, you can move it back to the original file - it never actually executes the bootstrap calls, so it doesn't need to be serialized.

Tests/LoggingTests/LocalLoggingTest.swift
```diff
@@ -36,29 +36,33 @@ class LocalLoggerTest: XCTestCase {
         logging.history.assertExist(level: .error, message: "Struct3::doSomething", metadata: ["bar": "baz"])
         logging.history.assertExist(level: .error, message: "Struct3::doSomethingElse", metadata: ["bar": "baz"])
         logging.history.assertExist(level: .warning, message: "Struct3::doSomethingElseAsync", metadata: ["bar": "baz"])
-        logging.history.assertExist(level: .info, message: "TestLibrary::doSomething")
-        logging.history.assertExist(level: .info, message: "TestLibrary::doSomethingAsync")
```
czechboy0, line 40: Is this missing because we're not bootstrapping the global log handler anymore? If so, we might need to turn this one back into a global test and move it, to ensure we test collecting logs from libraries as well, even if we didn't explicitly provide the Logger to them.

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -723,14 +787,14 @@ class LoggingTest: XCTestCase {
         )
     }
 
-    func testLoggerWithoutFactoryOverrideDefaultsToUsingLoggingSystemMetadataProvider() {
+    @Test func loggerWithoutFactoryOverrideDefaultsToUsingLoggingSystemMetadataProvider() {
```
czechboy0: I think this one also needs to be a global test, suggested by the test name

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -741,14 +805,15 @@ class LoggingTest: XCTestCase {
         )
     }
 
-    func testLoggerWithPredefinedLibraryMetadataProvider() {
+    @Test func loggerWithPredefinedLibraryMetadataProvider() {
```
czechboy0: This one too should stay global

Tests/LoggingTests/LoggingTest.swift
```diff
@@ -759,17 +824,13 @@ class LoggingTest: XCTestCase {
         )
     }
 
-    func testLoggerWithFactoryOverrideDefaultsToUsingLoggingSystemMetadataProvider() {
+    @Test func loggerWithFactoryOverrideDefaultsToUsingLoggingSystemMetadataProvider() {
```
czechboy0: This one too should stay global

Tests/LoggingTests/TestLogger.swift
```diff
@@ -233,17 +233,17 @@ extension History {
         message: String,
         metadata: Logger.Metadata? = nil,
         source: String? = nil,
-        file: StaticString = #filePath,
+        file: String = #filePath,
```
czechboy0, line 236: Why the change?

CONTRIBUTING.md
```diff
@@ -62,35 +62,15 @@ We require that your commit messages match our template. The easiest way to do t
 
 ### Make sure Tests work on Linux
 
-SwiftLog uses XCTest to run tests on both macOS and Linux. While the macOS version of XCTest is able to use the Objective-C runtime to discover tests at execution time, the Linux version is not.
-For this reason, whenever you add new tests **you have to run a script** that generates the hooks needed to run those tests on Linux, or our CI will complain that the tests are not all present on Linux. To do this, merely execute `ruby ./scripts/generate_linux_tests.rb` at the root of the package and check the changes it made.
+SwiftLog uses Swift Testing to run tests on all supported platforms.
```
czechboy0, line 65: Thanks for updating this guide, you could copy some parts from https://github.com/apple/swift-openapi-runtime/blob/main/CONTRIBUTING.md#run-ci-checks-locally as well, which describe in more detail how to run these checks locally.

czechboy0 (changes requested):
Thanks @kukushechkin, added a few comments, but overall looks great 🙏

Tests/LoggingTests/LocalLoggingTest.swift
kukushechkin: `GlobalLoggerTest` also already has checks for `TestLibrary::` messages.

Tests/LoggingTests/TestLogger.swift
kukushechkin: To avoid creating a new `String` as `SourceLocation` is expecting `String`.

CONTRIBUTING.md
kukushechkin: This is pretty cool. Copied the guide and verified it works with `act workflow_call -j soundness --input shell_check_enabled=true`

Tests/LoggingTests/LoggingTest.swift
kukushechkin: Yes, indeed it is testing the global metadata override. Moved to the global serialised suite.

Tests/LoggingTests/LoggingTest.swift
kukushechkin: Moved as well.

Tests/LoggingTests/LoggingTest.swift
kukushechkin: Moved as well.

Tests/LoggingTests/LoggingTest.swift
kukushechkin: I don't quite catch the difference between  `loggerWithoutFactoryOverrideDefaultsToUsingLoggingSystemMetadataProvider` and `loggerWithPredefinedLibraryMetadataProvider`. Both test a `Logger` without a custom factory, just with different global metadata.

Tests/LoggingTests/LocalLoggingTest.swift
czechboy0: Fair enough, then you can leave it as-is.

Tests/LoggingTests/LoggingTest.swift
czechboy0: Yeah the only difference is the construction of the metadata, which might be what it's explicitly testing.

Issue #291, opened 2024-04-23
Adding Error to log functions


Gucky:
There was a slightly similar issue ([Simplify logging Error types #267](https://github.com/apple/swift-log/issues/267)) but I would like to give this another twist.

Basically all advanced logging systems accept a textual message and allow to add additional values to better describe the log event.
Any also many system allow to explicitly log an Error, not just as part of the message or metadata. Examples: Sentry, Datadog
These service do handle explicit errors in special ways like automated grouping, warning thresholds, incident management...
It's more complex to train these systems to analyze/filter/map the metadata of log events and perform the same routines as with explicit errors.

So I would like to propose to add an additional optional Error to the log functions that then can be send to log services.
Manual extraction of key values from an error, in it's simplest form `error.localizedDescription` or more complex, is not sufficient to leverage the potential of log services.

### Given
```swift
let error: Error? = ...
let logger = Logger(label: "com.example")
```

### Expected behavior
```swift
logger.error("message", error)
```
and a signature of the log(...) function of the LogHandler protocol
```swift
func log(
        level: Logging.Logger.Level,
        message: Logging.Logger.Message,
        error: Error?, // <<<<<<<<<< new line here
        metadata: Logging.Logger.Metadata?,
        source: String,
        file: String,
        function: String,
        line: UInt
    )
```

### Actual behavior
The given log functions `.log(...)` or explicit like `.error(...)` on a logger instance only accept a message and metadata where metadata is a String, or array/dictionary hierarchy where terminal elements are again String.

### SwiftLog version/commit hash
v1.5.4

raviraj-tribe:
Thank you! Looking forward for this!

one1color:
Would be nice :)

jefflewis:
I think this would greatly improve the ergonomics of this logging system. So many logging services have special affordances for `Error` that being able to `log.error` the actual error and have it end up in the log handler for proper processing seems like a critical need.

As-is, to do this, I have to wrap an error with a `CustomStringConvertible` conforming error wrapper, and attempt to extract the error from _this_ error. :/

samuelmurray:
[...]

PR #384, opened 2025-10-29
Add option to pass Errors to log functions

samuelmurray:
Add option to pass Errors to log functions

### Motivation:

Addresses #291

### Modifications:

Add new API to Logger, that allows the caller to pass an Error object to any log function. Extend LogHandler to allow an implementation to access the error, in order to format the log appropriately. Provide default implementations to preserve backwards compatibility.

### Result:

By providing default implementations, the change should be compatible between old and new code, both from the API and implementation side. I.e. a client can use the new API and it will still work with 'old' implementations, and vice versa.

### DIFF
--- a/Sources/Logging/LogHandler.swift
+++ b/Sources/Logging/LogHandler.swift
@@ -139,6 +139,20 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
     func log(
         level: Logger.Level,
         message: Logger.Message,
+        error: Error?,
+        metadata: Logger.Metadata?,
+        source: String,
+        file: String,
+        function: String,
+        line: UInt
+    )
+    
+    /// SwiftLog 1.6 compatibility method. Please do _not_ implement, implement
+    /// `log(level:message:error:metadata:source:file:function:line:)` instead.
+    @available(*, deprecated, renamed: "log(level:message:error:metadata:source:file:function:line:)")
+    func log(
+        level: Logger.Level,
+        message: Logger.Message,
         metadata: Logger.Metadata?,
         source: String,
         file: String,
@@ -147,8 +161,8 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
     )
 
     /// SwiftLog 1.0 compatibility method. Please do _not_ implement, implement
-    /// `log(level:message:metadata:source:file:function:line:)` instead.
-    @available(*, deprecated, renamed: "log(level:message:metadata:source:file:function:line:)")
+    /// `log(level:message:error:metadata:source:file:function:line:)` instead.
+    @available(*, deprecated, renamed: "log(level:message:error:metadata:source:file:function:line:)")
     func log(
         level: Logging.Logger.Level,
         message: Logging.Logger.Message,
@@ -196,6 +210,7 @@ extension LogHandler {
                     level: .warning,
                     message:
                         "Attempted to set metadataProvider on \(Self.self) that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.",
+                    error: nil,
                     metadata: nil,
                     source: "Logging",
                     file: #file,
@@ -213,6 +228,20 @@ extension LogHandler {
     public func log(
         level: Logger.Level,
         message: Logger.Message,
+        error: Error?,
+        metadata: Logger.Metadata?,
+        source: String,
+        file: String,
+        function: String,
+        line: UInt
+    ) {
+        self.log(level: level, message: message, metadata: metadata, source: source, file: file, function: function, line: line)
+    }
+    
+    @available(*, deprecated, renamed: "log(level:message:error:metadata:source:file:function:line:)")
+    public func log(
+        level: Logger.Level,
+        message: Logger.Message,
         metadata: Logger.Metadata?,
         source: String,
         file: String,
@@ -222,7 +251,7 @@ extension LogHandler {
         self.log(level: level, message: message, metadata: metadata, file: file, function: function, line: line)
     }
 
-    @available(*, deprecated, renamed: "log(level:message:metadata:source:file:function:line:)")
+    @available(*, deprecated, renamed: "log(level:message:error:metadata:source:file:function:line:)")
     public func log(
         level: Logging.Logger.Level,
         message: Logging.Logger.Message,
@@ -234,6 +263,7 @@ extension LogHandler {
         self.log(
             level: level,
             message: message,
+            error: nil,
             metadata: metadata,
             source: Logger.currentModule(filePath: file),
             file: file,
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -104,6 +104,49 @@ extension Logger {
     /// - parameters:
     ///    - level: The log level to log `message` at. For the available log levels, see `Logger.Level`.
     ///    - message: The message to be logged. `message` can be used with any string interpolation literal.
+    ///    - error: An `Error` related to the event.
+    ///    - metadata: One-off metadata to attach to this log message.
+    ///    - source: The source this log messages originates from. Defaults
+    ///              to the module emitting the log message.
+    ///    - file: The file this log message originates from (there's usually no need to pass it explicitly as it
+    ///            defaults to `#fileID`).
+    ///    - function: The function this log message originates from (there's usually no need to pass it explicitly as
+    ///                it defaults to `#function`).
+    ///    - line: The line this log message originates from (there's usually no need to pass it explicitly as it
+    ///            defaults to `#line`).
+    @inlinable
+    public func log(
+        level: Logger.Level,
+        _ message: @autoclosure () -> Logger.Message,
+        error: Error? = nil,
+        metadata: @autoclosure () -> Logger.Metadata? = nil,
+        source: @autoclosure () -> String? = nil,
+        file: String = #fileID,
+        function: String = #function,
+        line: UInt = #line
+    ) {
+        if self.logLevel <= level {
+            self.handler.log(
+                level: level,
+                message: message(),
+                error: error,
+                metadata: metadata(),
+                source: source() ?? Logger.currentModule(fileID: (file)),
+                file: file,
+                function: function,
+                line: line
+            )
+        }
+    }
+    
+    /// Log a message passing the log level as a parameter.
+    ///
+    /// If the `logLevel` passed to this method is more severe than the `Logger`'s ``logLevel``, it will be logged,
+    /// otherwise nothing will happen.
+    ///
+    /// - parameters:
+    ///    - level: The log level to log `message` at. For the available log levels, see `Logger.Level`.
+    ///    - message: The message to be logged. `message` can be used with any string interpolation literal.
     ///    - metadata: One-off metadata to attach to this log message.
     ///    - source: The source this log messages originates from. Defaults
     ///              to the module emitting the log message.
@@ -123,17 +166,7 @@ extension Logger {
         function: String = #function,
         line: UInt = #line
     ) {
-        if self.logLevel <= level {
-            self.handler.log(
-                level: level,
-                message: message(),
-                metadata: metadata(),
-                source: source() ?? Logger.currentModule(fileID: (file)),
-                file: file,
-                function: function,
-                line: line
-            )
-        }
+        self.log(level: level, message(), error: nil, metadata: metadata(), source: source(), file: file, function: function, line: line)
     }
 
     /// Log a message passing the log level as a parameter.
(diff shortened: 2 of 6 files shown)


Sources/Logging/LogHandler.swift
```diff
@@ -136,6 +136,20 @@ public protocol LogHandler: _SwiftLogSendableLogHandler {
...
+        metadata: Logger.Metadata?,
+        source: String,
+        file: String,
+        function: String,
+        line: UInt
+    )
+    
+    /// SwiftLog 1.6 compatibility method. Please do _not_ implement, implement
+    /// `log(level:message:error:metadata:source:file:function:line:)` instead.
+    @available(*, deprecated, renamed: "log(level:message:error:metadata:source:file:function:line:)")
```
0xTim, line 152: Why are we deprecating this and making error a standardised parameter in all log messages?

czechboy0:
> Errors may still be NSErrors with domain and error codes, and having access to that is helpful for efficient grouping

Right, but that is starting to sound like wanting to "see" through the abstraction layer, because you know what's happening on both sides of the LogHandler - you have a _specific_ LogHandler, one that manages errors the way you need, plus you also have your own code that emits errors directly into the Logger, instead of adding them as metadata.

Maybe the right question we should ask ourselves to help figure out if this should be a convenience on Logger, or actually something that propagates through the LogHandler: are errors _more_ than just metadata? Or are they just a common piece of metadata, same as an HTTP response code, HTTP headers, raw data, a "cause" error string, and so on? My current lean is the latter, which is why I just suggested a method on Logger, but I'm open to having my mind changed if we find how an error attached to a log line is fundamentally different to any other piece of metadata attached to a log line.

Sources/Logging/LogHandler.swift
samuelmurray: My reasoning was that a LogHandler should not have to implement two separate methods - one with errors and one without. But still the error parameter is optional, so it’s not expected to be included in all log post, just that a LogHandler should decide how to include the error in the resulting log post, if passed to the log method.
Of course this is dependent on the other discussion in this PR, whether or not we want the argument to be passed to the handler at all.

samuelmurray:
> We could change the definition of Metadata to also allow an Error. That will also allow to transport an error through the logger layer. To me that doesn't feel right. Like piggybacking essential information among noise as - again - for .error and .warning logs an error is the source of the event and not just metadata.

I agree that this doesn’t feel right. Also it would be hard to agree on a common key in which to pass the error. I suspect there are already consumers of the api that pass stringified errors under “error” or “err”, and so then the log handlers would still have to do pattern matching to see if a value is an Error. 

For me a big reasong for adding this is to allow LogHandlers to do structured logging, e.g. if you want a handler to output ECS-formatted logs. Then you want to pass the error message to “error.message”, and perhaps the type name to “error.type”. You might also want to do pattern matching to extract an error code (from an NSError) to “error.code” and so on. Simply passing the stringified error to the handler prevents all of this, or forces the caller of the log API to do all the extraction (passing the message, code, and type to separate metadata fields) which then couples the code that generates the logs with the desired log format.
The same argument could be said about other metadata as well, but I guess to me, as @Gucky is arguing, an Error is not just “any other metadata” but rather could be related to the cause of the log post in the first place.

czechboy0:
Can you describe what a LogHandler would do with an `Error`, other than turn it into a string, without knowing what concrete types of error your code is sending to the Logger? Maybe I'm missing something here, but Error alone gives you very little when it comes to programmatic handling. Are you saying every LogHandler should try to cast every incoming Error into several common concrete error types, like DecodingError, etc? Or how can you go from an Error back to the structured error you describe above, without coupling it with the emission code?

samuelmurray:
> Overall, I am open to changing both `LogHandler` and `Logger` to accept `error`s since this often carry a lot of additional information as pointed out here. However, I would like to see this change be motivated with a formal lightweight proposal similar to how we did it with other changes. In particular, this should call out some of the intended usage here and it should also address the source stability concerns of introducing a new method on `Logger` and the `LogHandler` protocol.

This sounds reasonable. @Gucky would you be able to help produce such a proposal?

0xTim:
> Imagine the following example:
> 
> ```swift
> enum NetworkError: Error {
>   case hostNotFound
>   case httpCode(Int)
> }
> ```

I think the point that was @czechboy0 was making was that if you have an error like that in your application you also need the log handler you're using to know about that error as well so it knows how to handle the error. At which point, why not do that logic in the application anyway (not saying I completely disagree with the premise of the PR, but we are going to have to make sure that log handlers decide what to do with an error)

samuelmurray:
> I think the point that was @czechboy0 was making was that if you have an error like that in your application you also need the log handler you're using to know about that error as well so it knows how to handle the error.

I disagree with this. The error can come from a third party library that your application uses, and so you might not know what error type you caught. My example was just to show that there is more information that the LogHandler might want to extract, other than the stringified version of the error. (Of course, the application code could also extract this, and pass as metadata, but the idea of this PR is to shift the responsibility to the LogHandler on what to extract and not, and how to format the final log post, and provide a unified API for how we log Errors.)

czechboy0:
What I think would be very reasonable is an extension like this, in Swift Log:

```swift
extension Logger {
    public func log(
        level: Logger.Level,
        error: any Error,
        file: String = #fileID,
        function: String = #function,
        line: UInt = #line
    ) {
        self.log(
            level: level, 
            message(), 
            metadata: { 
                var metadata = metadata()
                metadata["error.type"] = "\(type(of: error))"
                metadata["error.message"] = "\(error)"
                return metadata
            }(),
            source: source(), 
            file: file, 
            function: function, 
            line: line
        )
    }
}
```

And in your application code, you could implement additional similar extensions that work with the errors your app is throwing.

I'm just not sure what the LogHandler can do that a Logger extension _can't_ do. And the upside of the above is that it'll work with every LogHandler, not just those that explicitly opt into the potential new override. Keep in mind that there are many LogHandlers in the ecosystem already, and ideally they all shouldn't have to be updated to take advantage of more ergonomic error logging.

samuelmurray:
> I'm just not sure what the LogHandler can do that a Logger extension _can't_ do. And the upside of the above is that it'll work with every LogHandler, not just those that explicitly opt into the potential new override. Keep in mind that there are many LogHandlers in the ecosystem already, and ideally they all shouldn't have to be updated to take advantage of more ergonomic error logging.

I agree that a sensible default such as that would be nice, so we don’t risk having people attaching errors (or worse, change existing code from manually stringifying errors to using the new extensions) that are completely ignored by existing LogHandlers. However, why not put the default implementation at the LogHandler level? I.e. change the proposed new default implementation ([LogHandler.swift#227](https://github.com/apple/swift-log/pull/384/files#diff-e992e165711bc3ac167a81e9ea45c69d65a0eea7127dad70bcc4ba4c54483d9bR227-R239)) to

```swift
extension LogHandler
    @available(*, deprecated, message: “Implement this method if you want a custom serialization of errors")
    public func log(
        level: Logger.Level,
        message: Logger.Message,
        error: Error?,
        metadata: Logger.Metadata?,
        source: String,
        file: String,
        function: String,
        line: UInt
    ) {
        self.log(
          level: level, 
          message: message, 
          metadata: { 
                var metadata = metadata()
                metadata["error.type"] = "\(_typeName(type(of: error), qualified: true))"
                metadata["error.message"] = "\(error)"
                return metadata
          }(),
          source: source, 
          file: file, 
          function: function, 
          line: line
        )
    }
}
```

This way, we get a decent default, but still allow LogHandlers to do something else if they want to. To me, it’s not about what _can_ be done at each layer (application, Logger-API, LogHandler), but rather which layer gets to decide/influence what. The way I see it, the application is responsible for deciding *what* to log, and the LogHandler decides the output format, and the sink. “error.message” is what ECS expects, but what about a LogHandler that outputs logs in a different format? Why not allow it to choose its own keys?

samuelmurray:
We can also compare with `apple/swift-distributed-tracing`, where the protocol `Span` has a function [`recordError`](https://github.com/apple/swift-distributed-tracing/blob/main/Sources/Tracing/SpanProtocol.swift#L83-L87) which accepts an `Error`. This lets an implementation, e.g. [`swift-otel`](https://github.com/swift-otel/swift-otel/blob/main/Sources/OTel/OTelCore/Tracing/OTelSpan.swift#L302-L303) decide how to serialize the error, i.e. in this case choose what keys to use (note that they use the OTel keys `"exception.*"`, by which I mean to say that there might not be a “one key fits all"). It also let’s a library that uses distributed tracing, like [`grpc-swift-extras`](https://github.com/grpc/grpc-swift-extras/blob/main/Sources/GRPCOTelTracingInterceptors/Tracing/ServerOTelTracingInterceptor.swift#L211), pass an error to `recordError` without specifying the format, and so we are guaranteed that we get consistent keys and formats across an application that uses OTel and gRPC, independent of if the error was recorded from my application code, or from gRPC.

Contrast this with logging in [`gRPC` 1.X](https://github.com/grpc/grpc-swift/blob/release/1.x/Sources/GRPC/LoggingServerErrorDelegate.swift#L26) where they log an `Error`, but have to explicitly stringify it, and set a key (they choose `error`). If my application uses gRPC and swift-log (and any suitable log implementation), sure I can manually format all errors _in my application code_, but they will differ from those logged by `swift-grpc`.

What I hope to enable is for logging to eventually be more like the former.

FranzBusch:
> I'll point out that there can be valid LogHandler implementations today that will not be able to implement this new requirement. For example, a LogHandler that transparently forwards all calls over the network to another collector. Until now, everything passed through was serializable data. If we add a custom user error type to LogHandler, and the upstream LogHandler inspects specific known error types, these network-based log handlers literally have no way to adapt.
> 
> That's why I'm against ever including user types in the protocol's abstraction. We go from plain ol' data abstraction to stateful user types, which breaks the ability to mix any emission code with any log handler.
> 
> For adopters who own both the emission and collection side, they should skip the abstraction and pass their data directly. But we shouldn't break the concept of the abstraction just to make it easier to use for adopters who don't need the abstraction in the first place, since they know and control both sides.

I don't think that's true nor does that proposal change that. `LogHandler`s today are already dealing with types that they need to decide themselves how to serialize i.e. all the metadata needs to be decided by the individual log handler how to serialize. 
Having said that this additional error parameter is something that `LogHandler`'s can handle but our default extension implementation will just extract as much information generically as possible and attach it as metadata. This way `LogHandler'`s that either don't want to or can't handle `Error`s more specifically will continue to work.

Sources/Logging/Logging.swift
```diff
@@ -135,6 +138,36 @@ extension Logger {
...
+    public func log(
+        level: Logger.Level,
+        _ message: @autoclosure () -> Logger.Message,
+        metadata: @autoclosure () -> Logger.Metadata? = nil,
+        source: @autoclosure () -> String? = nil,
+        file: String = #fileID,
+        function: String = #function,
+        line: UInt = #line
+    ) {
+        self.log(level: level, message(), error: nil, metadata: metadata(), source: source(), file: file, function: function, line: line)
```
aluco100, line 169: Nit: I think here could be a multiline  method call

FranzBusch:
> This part is interesting. So let's assume we keep user types from leaking down to the `LogHandler`s by catching and flattening errors on the `Logger` level. We need to pick some reasonable labels for this metadata, not just random strings like `error.type` and `error.message`, let's introduce a `enum MetadataKey { case errorType, errorMessage, custom(String) }` (this can be backward compatible for `LogHandler`s supporting just String keys with `typealias StructuredMetadata = [MetadataKey: MetadataValue]`). Now this becomes part of the interface between the `Logger` and `LogHandler`s and they can apply any kind of transformation to the keys. I can even imagine some optional `LogHandler` method for providing these custom labels mapping for `Logger` to apply. This is as close as we can get to a structured representation for unknown in advance errors.
> 
> But this is just 2 fields, what if we want to handle more complicated (but known in advance!) error types? In this case we can add a new `protocol LoggingMetadataError { func asMetadata() -> StructuredMetadata }` so `Logger` can handle those types. Conforming errors to this protocol is responsibility of the application. I can't imagine a situation when a particular `LogHandler` knows and wants to handle specific errors from the application layer, but if this is an actual use-case, I want to learn about it.

In my opinion, this is overcomplicating the matter a lot. If we decide that we want to support logging error instances then we must choose a default implementation based on the current `LogHandler` protocol. I personally think the proposed implementation that just creates two metadata values with default keys is fine. We don't need to offer customization for this in the default implementation. Log handlers that want to handle errors specifically can implement the new protocol requirement. Users can also use a custom log handler themselves that just changes the implementation of the new error method to change the default keys that are being used.

PR #386, opened 2025-11-06
Logging instructions and allocations benchmarks

kukushechkin:
Add simple benchmarks measuring instructions count and allocations when logging with various log levels.

### Motivation:

Establish the baseline and have automated checks before approaching performance improvements.

### Modifications:

- Dependency on https://github.com/ordo-one/package-benchmark added
- A new `Benchmarks` sub-package is added
- CI workflows now run benchmarks

### Result:

- CI workflows are running benchmarks on Linux and macOS
- Thresholds are verified only on macOS

An example of a benchmark failing with a **better** results than expected: https://github.com/apple/swift-log/actions/runs/19540944353/job/55950355442

An example of a benchmark failing with a **worse** results than expected: https://github.com/apple/swift-log/actions/runs/19569105729/job/56038101552

### DIFF
--- a/Benchmarks/Benchmarks/SwiftLogBenchmarks/NoOpLogHandler.swift
+++ b/Benchmarks/Benchmarks/SwiftLogBenchmarks/NoOpLogHandler.swift
@@ -14,7 +14,7 @@
 
 import Logging
 
-struct EmptyLogHandler: LogHandler {
+struct NoOpLogHandler: LogHandler {
     let label: String
     public var metadataProvider: Logger.MetadataProvider?
 
--- a/Benchmarks/Benchmarks/SwiftLogBenchmarks/SwiftLogBenchmarks.swift
+++ b/Benchmarks/Benchmarks/SwiftLogBenchmarks/SwiftLogBenchmarks.swift
@@ -26,8 +26,8 @@ let benchmarks: @Sendable () -> Void = {
         for logLevelUsed in logLevelParameterization {
             var logger = Logger(label: "BenchmarkRunner_\(logLevel)_\(logLevelUsed)")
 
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -30,6 +30,16 @@ jobs:
...
 
+  benchmarks:
+    name: Benchmarks
+    uses: apple/swift-nio/.github/workflows/benchmarks.yml@main
+    with:
+      benchmark_package_path: "Benchmarks"
+      macos_xcode_16_3_enabled: true
+      macos_xcode_16_4_enabled: true
+      macos_xcode_26_0_enabled: true
+      macos_xcode_26_1_enabled: true
```
czechboy0, line 39: Let's just run 16.4 and 26.1 here to match https://github.com/swiftlang/project-operations/blob/main/github-policies.md#principles

Benchmarks/Benchmarks/SwiftLogBenchmarks/EmptyLogger.swift
```diff
@@ -0,0 +1,78 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Logging API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Logging
+
+struct EmptyLogHandler: LogHandler {
```
FranzBusch: NIT: We usually call this NoOpLogHandler

kukushechkin: Updated

PR #389, opened 2025-11-27
[SLG-0002] Implementation: Compile-time log level elimination using traits

kukushechkin:
Add compile-time traits for eliminating log levels from binaries

###  Motivation:

Applications often know in advance which log levels will never be needed in production. For example, a production service might only need warning and above, while trace and debug levels are only useful during development. This PR introduces
  compile-time traits that completely eliminate unwanted log levels from binaries, achieving zero runtime overhead for disabled log statements.

### Modifications:

- Added seven `MaxLogLevel*` traits
- Extended benchmarks with the `MaxLogLevelWarning` suite
- Added "Disable log levels during compilation" section to the best practices guide

### Result:

Users can eliminate unwanted log levels at compile time by specifying traits in their `Package.swift` dependency declaration.

### DIFF
--- a/.github/workflows/pull_request_label.yml
+++ b/.github/workflows/pull_request_label.yml
@@ -14,7 +14,7 @@ jobs:
     timeout-minutes: 1
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           persist-credentials: false
       - name: Check for Semantic Version label


Package.swift
```diff
@@ -20,6 +20,16 @@ let package = Package(
...
         .library(name: "Logging", targets: ["Logging"])
     ],
+    traits: [
+        .trait(name: "DisableTraceLogs"),
+        .trait(name: "DisableDebugLogs"),
+        .trait(name: "DisableInfoLogs"),
+        .trait(name: "DisableNoticeLogs"),
+        .trait(name: "DisableWarningLogs"),
+        .trait(name: "DisableErrorLogs"),
+        .trait(name: "DisableCriticalLogs"),
```
FranzBusch: We should provide a description for each trait

Package.swift
```diff
@@ -20,6 +20,43 @@ let package = Package(
...
+    // swift-log provides compile-time traits to completely eliminate specific log levels from your binary.
+    // When a log level is disabled via a trait, both the level-specific method (e.g., `logger.debug()`)
+    // and calls to `logger.log(level: .debug)` become no-ops at compile time, with zero runtime overhead.
+    //
+    // Usage:
+    //   .package(url: "...", traits: ["DisableDebugLogs"])
+    //
+    // Performance impact:
+    //   - Level-specific methods (`.debug()`, `.info()`, etc.): Entire method body compiled out
+    //   - Generic `.log(level:)`: Uses O(1) switch statement; disabled levels jump to no-op default case
```
ktoso: nit: Not sure if documenting the switch statement's performance characteristics in O-notation far away here in the traits is the most useful, what we care about is that the log statements become a noop, which we said up there already

kukushechkin: I would still add this as this is a change in behaviour with traits comparing to no traits. If someone heavily using `logger.log()` for high levels of logs and do not compile them out, their code will become less performant.

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
```diff
@@ -128,11 +128,84 @@ logger.info("Request failed")
...
+logger.info("Database query executed")
 logger.info("Response sent")
 
 // ✅ Good: Use appropriate levels instead
 logger.debug("Processing request", metadata: ["path": "\(path)"])
 logger.trace("Query", metadata: ["sql": "\(query)"])
 logger.debug("Request completed", metadata: ["status": "\(status)"])
 ```
+
+### Disable log levels during compilation
```
FranzBusch: I think this should become its own article. This best practice is focused on documented what log levels library authors should choose. I would recommend to create a new article (not a best practice) that document these new compile time traits.

Package.swift
FranzBusch: I am not sure if the docs here are in general helpful or not. Very few people look at the `Package.swift`. Could we instead add a description to the below traits e.g. `trait(name: "", description: "Link to new doc article")`?

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
```diff
@@ -128,11 +128,84 @@ logger.info("Request failed")
...
+your binary reducing the runtime overhead.
+
+#### Motivation
+
+When deploying applications to production, you often know in advance which log
+levels will never be needed. For example, a production service might only need
+warning and above, while trace and debug levels are only useful during
+development. By using traits, you can completely remove these unnecessary log
+levels at compile time, achieving zero runtime overhead.
+
```
FranzBusch: We should mention here that these traits really should only be set by applications since they apply globally across the dependency tree. Libraries should not pass these traits.

Sources/Logging/Logging.swift
```diff
@@ -123,6 +125,66 @@ extension Logger {
         file: String = #fileID,
         function: String = #function,
         line: UInt = #line
+    ) {
+        #if DisableTraceLogs || DisableDebugLogs || DisableInfoLogs || DisableNoticeLogs || DisableWarningLogs || DisableErrorLogs || DisableCriticalLogs
```
FranzBusch: This is no longer needed right? At least I wasn't able to spot where `DisableTraceLogs` would be set.

kukushechkin: Yes

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
kukushechkin: Created a separate article.

Sources/Logging/Docs.docc/BestPractices/001-ChoosingLogLevels.md
kukushechkin: Added

PR #390, opened 2025-12-01
Add an InMemoryLogHandler

hamzahrmalik:
Add an InMemoryLogHandler

### Motivation:

Library maintainers should be able to test that their libraries log what they expect

### Modifications:

Create a new InMemoryLogging product, which contains an InMemoryLogHandler

This log handler can be used to make a logger, pass the logger into some function, and then assert that logs were emitted.

See the tests for example usage

### Future:

For now i have left this very bare-bones. We could in future add convenience functions. For example
- a function to assert that a particular log message was logged
- a static function to create a logger with this handler

### DIFF
--- a/Sources/InMemoryLogging/InMemoryLogHandler.swift
+++ b/Sources/InMemoryLogging/InMemoryLogHandler.swift
@@ -15,9 +15,9 @@
 import Logging
 import Synchronization
 
-/// A custom log handler which doesn't actually emit logs, but just collects them into memory.
-/// You can then retrieve a list of what was logged and run assertions on it.
-/// This handler is intended to be used in tests.
+/// A custom log handler which just collects logs into memory.
+/// You can then retrieve an array of those log entries.
+/// Example use cases include testing and buffering.
 ///
 /// # Usage
 /// ```swift
@@ -28,9 +28,11 @@ import Synchronization
 ///       logHandler
 ///     }
 /// )
-/// // Use logger to emit some logs, then:
+/// // Use logger to emit some logs
+/// someFunction(logger: logger)
+///
+/// // Retrieve what was logged
 /// let logEntries = logger.entries
-/// #expect(logEntries.contains(...))
 /// ```
 ///
 @available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, *)


Sources/InMemoryLogging/InMemoryLogHandler.swift
```diff
@@ -0,0 +1,131 @@
...
+        private let logs: Mutex<[Entry]> = .init([])
+
+        fileprivate func append(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata) {
+            self.logs.withLock {
+                $0.append(
+                    Entry(
+                        level: level,
+                        message: message,
+                        metadata: metadata
+                    )
```
ktoso, line 71: for completeness, wouldn't we want to also store the source/line? Maybe bad idea to assert on them but perhaps can be useful in debugging in some weird situations, not up to us to judge that I think 🤔

hamzahrmalik: We could do. But then equating Entry's becomes hard, unelss you know what source to expect...

ktoso: We'd skip it from equatable and hashable impls -- lets see if anyone cares about it during review; in any case this is LGTM as well as is

FranzBusch: I'm happy to not store file/line for now in this handler.

Sources/InMemoryLogging/InMemoryLogHandler.swift
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    /// All logs that have been collected.
+    public var entries: [Entry] {
+        self.logStore.entries
+    }
+
+    /// Clear all entries.
+    public func clear() {
```
FranzBusch, line 130:
```suggestion
    public mutating func clear() {
```

hamzahrmalik: this works without being mutating, because the underlying store is held in a Mutex

What's the benefit of adding mutating here?

FranzBusch: I know. I was just thinking if this mutating would future proof this more in the case where we want to change the underlying implementation.

proposals/0002-in-memory-handler.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Motivation
+
+Library maintainers should be able to test that their libraries log what they expect
+
+### Proposed solution
+
+Create a new InMemoryLogging product, which contains an InMemoryLogHandler
+
+This log handler can be used to make a logger, pass the logger into some function, and then assert that logs were emitted.
```
kukushechkin: Could you add a usage example right here in the proposal.

proposals/0002-in-memory-handler.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Detailed design
+
+Implementation is in https://github.com/apple/swift-log/pull/390
```
kukushechkin: For the detailed design, could you add InMemoryLogHandler public API. This will make it clear from the proposal what is the new add API and how to use it.

Sources/InMemoryLogging/InMemoryLogHandler.swift
hamzahrmalik: my only concern would be that it implies value semantics where there are not

E.g.
```swift
let logHandler = InMemoryLogHandler()
var copy = logHandler
copy.clear()
```

This clears both logHandler and copy

Sources/InMemoryLogging/InMemoryLogHandler.swift
```diff
@@ -0,0 +1,131 @@
...
+// See CONTRIBUTORS.txt for the list of Swift Logging API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Logging
+import Synchronization
+
+/// A custom log handler which doesn't actually emit logs, but just collects them into memory.
```
kukushechkin: Can it be used in the `Logger` tests instead of `TestLogHandler`?

hamzahrmalik: currenlty the TestLogHandler is much more sophisticated and has more functionality than this InMemoryLogHandler

It would be nice to re-use this there, and i will look into that as a followup

proposals/0002-in-memory-handler.md
kukushechkin, line 1: If we're not doing the proposal, no need to add the proposal .md either.

Sources/InMemoryLogging/InMemoryLogHandler.swift
FranzBusch:
```suggestion
/// A custom log handler which collects them into memory.
```

Sources/InMemoryLogging/InMemoryLogHandler.swift
```diff
@@ -0,0 +1,131 @@
...
+/// You can then retrieve a list of what was logged and run assertions on it.
+/// This handler is intended to be used in tests.
```
FranzBusch: I think this log handler has more applications than just tests. Can we make this doc comment a bit more general purpose?

Sources/InMemoryLogging/InMemoryLogHandler.swift
```diff
@@ -0,0 +1,131 @@
...
+/// ```
+///
+@available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, *)
+public struct InMemoryLogHandler: LogHandler {
+    public var metadata: Logger.Metadata = [:]
+    public var metadataProvider: Logger.MetadataProvider?
+    public var logLevel: Logger.Level = .info
+    private let logStore: LogStore
+
+    /// A single item which was logged.
```
FranzBusch:
```suggestion
    /// A struct representing a log entry.
```

PR #393, opened 2025-12-03
Fix Android build error with `@preconcurrency import Android`

MaxDesiatov:
### Motivation:

`swift-log` currently can't be built on Android due to this error:

```
/home/runner/work/WasmKit/WasmKit/.build/checkouts/swift-log/Sources/Logging/Logging.swift:1380:36: error: reference to var 'stdout' is not concurrency-safe because it involves shared mutable state
1378 |         let systemStdout = Glibc.stdout!
1379 |         #elseif canImport(Android)
1380 |         let systemStdout = Android.stdout
     |                                    `- error: reference to var 'stdout' is not concurrency-safe because it involves shared mutable state
```

### Modifications:

Marked `import Android` with `@preconcurrency` attribute.

### Result:

The package builds for Android now.

### DIFF
--- a/Sources/Logging/Docs.docc/Proposals/SLG-0001-metadata-providers.md
+++ b/Sources/Logging/Docs.docc/Proposals/SLG-0001-metadata-providers.md
@@ -1,10 +1,10 @@
 # Metadata Providers
 
-Authors: [Moritz Lang](https://github.com/slashmo), [Konrad 'ktoso' Malawski](https://github.com/ktoso)
-
-## Introduction
-
 While global metadata attributes may be manually set on a `LogHandler` level, there's currently no way of reliably providing contextual, automatically propagated, metadata when logging with swift-log.
+
+## Authors
+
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -47,8 +47,8 @@ jobs:
     name: Cxx interop
     uses: apple/swift-nio/.github/workflows/cxx_interop.yml@main
 
-  static-sdk:
-    name: Static SDK
+  static-linux:
+    name: Static Linux Swift SDK
```
czechboy0, line 51: Can you also rename it in main.yml too?

MaxDesiatov: That's already done in the other PR, would it matter which PR contains that change? https://github.com/apple/swift-log/pull/371/files

czechboy0: Didn't realize, yeah totally fine

MaxDesiatov:
Android build failed due to interrupted download, needs a failed CI jobs restart

ktoso (approved):
LGTM, thanks!

PR #394, opened 2025-12-04
Make InMemoryLogging available on all platforms

hamzahrmalik:
Make InMemoryLogging available on all platforms

### Motivation:

Users might want to adopt the in memory logger on lower platform versions.

### Modifications:

Remove availability annotations by using swift-logs own Lock instead of Mutex

### Result:

In memory logger can be used on all platforms that swift-log supports

### DIFF
--- a/Sources/InMemoryLogging/InMemoryLogHandler.swift
+++ b/Sources/InMemoryLogging/InMemoryLogHandler.swift
@@ -13,7 +13,6 @@
 //===----------------------------------------------------------------------===//
 
 import Logging
-import Synchronization
 
 /// A custom log handler which just collects logs into memory.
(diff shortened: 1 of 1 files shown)


Sources/InMemoryLogging/InMemoryLogHandler.swift
```diff
@@ -35,7 +35,6 @@ import Synchronization
 /// let logEntries = logger.entries
 /// ```
 ///
-@available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, *)
```
glbrntt, line 38: You can remove the `Synchronization` import too

PR #395, opened 2025-12-06
Conform `Logger.Level` to `CustomStringConvertible` and `LosslessStringConvertible`

fpseverino:
Conform `Logger.Level` to `CustomStringConvertible` and `LosslessStringConvertible`.

### Motivation:

There are libraries, such as Vapor, that add these conformances privately and `@retroactive`ly. This would avoid having to implement these conformances manually in each library and would unify and standardise the implementation.

### Modifications:

Make `Logger.Level` conform to `CustomStringConvertible` and `LosslessStringConvertible`, adding a public `description` computed property that returns the `rawValue` and a public initializer that takes a `description` string, lowercases it, and passes it to the `rawValue` initializer.

### Result:

`Logger.Level` conforms to `CustomStringConvertible` and `LosslessStringConvertible`.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -1026,6 +1026,19 @@ extension Logger.Level: Comparable {
     }
 }
 
+extension Logger.Level: CustomStringConvertible, LosslessStringConvertible {
+    /// A textual representation of the log level.
+    public var description: String {
+        self.rawValue
+    }
+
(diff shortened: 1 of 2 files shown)


sebsto:
This should have been a major release as it changes the public API of the library. 

Thank you @adam-fowler for having fixed the Lambda runtime.

PR #398, opened 2026-01-10
feat: Update lock implementation to more closely align with swift-nio implementation

scottmarchant:
### Motivation:

The swift-nio [lock implementation](https://github.com/apple/swift-nio/blob/main/Sources/NIOConcurrencyHelpers/lock.swift) differs from this older copy-paste of the implementation. This PR brings over the latest changes while preserving any deviations.

Besides alignment, this change improves the wasi pthread support, which was previously elided for wasm SDK's that support pthreads.

### Modifications:

- Copied `Lock` implementation over from [swift-nio](https://github.com/apple/swift-nio/blob/main/Sources/NIOConcurrencyHelpers/lock.swift)
- Extended the conditional patterns used in `Lock` to `ReadWriteLock

Deviations from swift-nio:

- Preserved Android condtional, which is missing from swift-nio
- Preserved `package` instead of `public` ACL specifiers
- Extended the conditional patterns used in `Lock` to `ReadWriteLock

### Result:

- Existing tests and builds still pass
- Mutex protection is no longer eliminated for wasm pthread targets.

### Testing done:

- Verified locally that wasm compilation still works using the following commands:

```
swift build --swift-sdk swift-6.3-DEVELOPMENT-SNAPSHOT-2026-01-06-a_wasm --target Logging
swift build --swift-sdk swift-6.3-DEVELOPMENT-SNAPSHOT-2026-01-06-a_wasm --target InMemoryLogging
```

- Verified unit tests all pass locally when running packages tests in Xcode
- Verified all expected [CI checks pass](https://github.com/PassiveLogic/swift-log/actions/runs/20870268506/job/59970396829?pr=1) in separate temporary PR: 

### Related work

This work is driven by a [larger effort](https://github.com/PassiveLogic/swift-web-examples/issues/1) by [PassiveLogic](https://passivelogic.com) to improve Wasm support across many swift repos.

This specific PR was requested as foundational work to support https://github.com/apple/swift-metrics/pull/190.

### DIFF
--- a/Sources/Logging/Locks.swift
+++ b/Sources/Logging/Locks.swift
@@ -26,20 +26,26 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if canImport(WASILibc)
-// No locking on WASILibc
-#elseif canImport(Darwin)
+#if canImport(Darwin)
 import Darwin
 #elseif os(Windows)
+import ucrt
 import WinSDK
 #elseif canImport(Glibc)
-import Glibc
+@preconcurrency import Glibc
 #elseif canImport(Android)
-import Android
+@preconcurrency import Android
 #elseif canImport(Musl)
-import Musl
+@preconcurrency import Musl
+#elseif canImport(Bionic)
+@preconcurrency import Bionic
+#elseif canImport(WASILibc)
+@preconcurrency import WASILibc
+#if canImport(wasi_pthread)
+import wasi_pthread
+#endif
 #else
-#error("Unsupported runtime")
+#error("The concurrency lock module was unable to identify your C library.")
 #endif
 
 /// A threading lock based on `libpthread` instead of `libdispatch`.
@@ -48,11 +54,12 @@ import Musl
 /// of lock is safe to use with `libpthread`-based threading models, such as the
 /// one used by NIO. On Windows, the lock is based on the substantially similar
 /// `SRWLOCK` type.
-package final class Lock: @unchecked Sendable {
-    #if canImport(WASILibc)
-    // WASILibc is single threaded, provides no locks
-    #elseif os(Windows)
+package final class Lock {
+    #if os(Windows)
     fileprivate let mutex: UnsafeMutablePointer<SRWLOCK> =
+        UnsafeMutablePointer.allocate(capacity: 1)
+    #elseif os(OpenBSD)
+    fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t?> =
         UnsafeMutablePointer.allocate(capacity: 1)
     #else
     fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t> =
@@ -61,14 +68,18 @@ package final class Lock: @unchecked Sendable {
 
     /// Create a new lock.
     package init() {
-        #if canImport(WASILibc)
-        // WASILibc is single threaded, provides no locks
-        #elseif os(Windows)
+        #if os(Windows)
         InitializeSRWLock(self.mutex)
-        #else
+        #elseif os(OpenBSD)
+        var attr = pthread_mutexattr_t(bitPattern: 0)
+        let err = pthread_mutex_init(self.mutex, &attr)
+        precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
+        #elseif (compiler(<6.1) && !os(WASI)) || (compiler(>=6.1) && _runtime(_multithreaded))
         var attr = pthread_mutexattr_t()
         pthread_mutexattr_init(&attr)
-        pthread_mutexattr_settype(&attr, .init(PTHREAD_MUTEX_ERRORCHECK))
+        debugOnly {
+            pthread_mutexattr_settype(&attr, .init(PTHREAD_MUTEX_ERRORCHECK))
+        }
 
         let err = pthread_mutex_init(self.mutex, &attr)
         precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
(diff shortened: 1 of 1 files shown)


Sources/Logging/Locks.swift
```diff
@@ -26,20 +26,26 @@
...
+@preconcurrency import Android
 #elseif canImport(Musl)
-import Musl
+@preconcurrency import Musl
+#elseif canImport(Bionic)
+@preconcurrency import Bionic
+#elseif canImport(WASILibc)
+@preconcurrency import WASILibc
+#if canImport(wasi_pthread)
+import wasi_pthread
```
scottmarchant, line 45: @MaxDesiatov Is there a special command I should use to test wasi_pthread compilation? I'm familiar with the commands for the swiftwasm.org pthreads sdks. But I thought I saw a special `-Xcc` parameter one can use to enable pthreads with the official swift.org Swift for WebAssembly SDK's. Can't seem to find that command, though.

I've verified that `swift build --swift-sdk wasm32-unknown-wasip1 --target Logging` works with these changes.

MaxDesiatov: Great, this should work. Thanks!

scottmarchant:
> Note, leaving this PR in draft at the moment while I finish checking on a few things. I expect to take it out of draft next week.

Ok, I believe everything is in order. @MaxDesiatov @czechboy0 @kateinoigakukun please feel free to review.

kateinoigakukun (approved):
Thanks! The Wasm part of the change seems good to me!

MaxDesiatov (approved):

Sources/Logging/Locks.swift
```diff
@@ -48,56 +54,56 @@ import Musl
...
+        #if os(Windows)
         InitializeSRWLock(self.mutex)
-        #else
+        #elseif os(OpenBSD)
+        var attr = pthread_mutexattr_t(bitPattern: 0)
+        let err = pthread_mutex_init(self.mutex, &attr)
+        precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
+        #elseif (compiler(<6.1) && !os(WASI)) || (compiler(>=6.1) && _runtime(_multithreaded))
         var attr = pthread_mutexattr_t()
         pthread_mutexattr_init(&attr)
```
czechboy0, line 79: Aren't we missing a pthread_mutexattr_destroy somewhere here?

Sources/Logging/Locks.swift
```diff
@@ -48,56 +54,56 @@ import Musl
...
-    #elseif os(Windows)
+package final class Lock {
+    #if os(Windows)
     fileprivate let mutex: UnsafeMutablePointer<SRWLOCK> =
         UnsafeMutablePointer.allocate(capacity: 1)
+    #elseif os(OpenBSD)
+    fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t?> =
+        UnsafeMutablePointer.allocate(capacity: 1)
     #else
     fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t> =
```
czechboy0, line 65: Would we use the same conditional `(compiler(<6.1) && !os(WASI)) || (compiler(>=6.1) && _runtime(_multithreaded))` to completely avoid the allocation/dealloc later?

Sources/Logging/Locks.swift
```diff
@@ -26,20 +26,26 @@
...
-// No locking on WASILibc
-#elseif canImport(Darwin)
+#if canImport(Darwin)
 import Darwin
 #elseif os(Windows)
+import ucrt
 import WinSDK
 #elseif canImport(Glibc)
-import Glibc
+@preconcurrency import Glibc
```
czechboy0, line 35: Why did we have to add `@preconcurrency` here now, was it emitting warnings without it?

Sources/Logging/Locks.swift
scottmarchant: @czechboy0 I think that happens in the deinit, on [line 96](https://github.com/apple/swift-log/pull/398/changes#diff-7b1bd45403dd1a7418287bd60682bf05ba2e4bf75dca1c0ca55a86d0477c9af5R96)

Sources/Logging/Locks.swift
scottmarchant: Yeah, it seems like we should use a `#elseif (compiler(<6.1) && !os(WASI)) || (compiler(>=6.1) && _runtime(_multithreaded))` condition here, and completely elide the `let mutex` out altogether in cases where it isn't used in the `init` method.

But we'd want to do that change in nio first, as this PR is just a copy-paste of the solution already implemented in nio.

Since the purpose of this PR is to align with the nio implementation, seems like we should either merge this as-is since it currently matches nio, or else merge a change to nio first and then downstream it here. Happy to go either route, but I'd prefer to merge this variation now, and run the smaller update through nio later. That way wasm support will be improved sooner.

Sources/Logging/Locks.swift
scottmarchant: The `@preconcurrency` is present in [the swift-nio implementation](https://github.com/apple/swift-nio/blob/main/Sources/NIOConcurrencyHelpers/lock.swift#L21). I didn't notice any warnings that this resolved, but since the goal here is to match the nio implementation as close as possible, figured it was best to update this.

Sources/Logging/Locks.swift
czechboy0: I don't think `pthread_mutex_destroy` calls `pthread_mutexattr_destroy`, I think the attributes are passed (in Swift terms) as borrowing, not consuming. So I think this might be leaking still today.

But do feel free to fix that in SwiftNIO first and then backport here, I agree it's good to keep them in sync.

czechboy0:
@scottmarchant I can't seem to be able to update the PR with the latest changes, can you do it manually please?

Sources/Logging/Locks.swift
scottmarchant: Oh, I completely missed noticing the `attr` part of `pthread_mutexattr_destroy` in your comment @czechboy0. Great find! Yes, I'll run an update through nio and another wave including that change as well.

scottmarchant:
> Approving in this form to make incremental progress, but please make the suggested improvements in the NIO version, especially the potential leak of the attributes.

@czechboy0 This is rebased on the tip of main now.

czechboy0:
My question is more about the explicit unsafe pointer allocate/deallocate calls stored in the Lock struct. I don't assume that'd get optimized away?

PR #401, opened 2026-01-16
Enable upcoming compiler features

kukushechkin:
Enable several upcoming compiler features to make the codebase more explicit in intention:

```
    // https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md
    // Require `any` for existential types.
    settings.append(.enableUpcomingFeature("ExistentialAny"))

    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
    settings.append(.enableUpcomingFeature("MemberImportVisibility"))

    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md
    settings.append(.enableUpcomingFeature("InternalImportsByDefault"))

    // https://docs.swift.org/compiler/documentation/diagnostics/nonisolated-nonsending-by-default/
    settings.append(.enableUpcomingFeature("NonisolatedNonsendingByDefault"))

    // Ensure all public types are explicitly annotated as Sendable or not Sendable.
    settings.append(.unsafeFlags(["-Xfrontend", "-require-explicit-sendable"]))
```

### Motivation:

Sync codebase with the modern practice, get ready for the upcoming features.

### Modifications:

- Compiler flags added for all the existing targets.
- Imports and access modifiers adjusted.
- `LoggingSystem` marked as `Sendable`

### Result:

Codebase is updated with the modern practices.

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -712,7 +712,7 @@ extension Logger {
 ///
 /// The default (``StreamLogHandler``) is intended to be a convenience.
 /// For production applications, implement the ``LogHandler`` protocol directly, or use a community-maintained backend.
-public enum LoggingSystem {
+public enum LoggingSystem: Sendable {
     private static let _factory = FactoryBox(
         { label, _ in StreamLogHandler.standardError(label: label) },
         violationErrorMesage: "logging system can only be initialized once per process."
--- a/Package@swift-6.0.swift
+++ b/Package@swift-6.0.swift
@@ -41,24 +41,24 @@ let package = Package(
     ]
 )
 
-for target in package.targets {
+for target in package.targets
+where [.executable, .test, .regular].contains(
+    target.type
+) {
     var settings = target.swiftSettings ?? []
-    settings.append(.enableExperimentalFeature("StrictConcurrency=complete"))
+
+    // https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md
+    // Require `any` for existential types.
+    settings.append(.enableUpcomingFeature("ExistentialAny"))
+
+    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
+    settings.append(.enableUpcomingFeature("MemberImportVisibility"))
+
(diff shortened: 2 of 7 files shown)


Package.swift
```diff
@@ -59,24 +59,27 @@ let package = Package(
...
+    settings.append(.enableUpcomingFeature("MemberImportVisibility"))
+
+    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md
+    settings.append(.enableUpcomingFeature("InternalImportsByDefault"))
+
+    // https://docs.swift.org/compiler/documentation/diagnostics/nonisolated-nonsending-by-default/
+    settings.append(.enableUpcomingFeature("NonisolatedNonsendingByDefault"))
+
+    // Ensure all public types are explicitly annotated as Sendable or not Sendable.
+    settings.append(.unsafeFlags(["-Xfrontend", "-require-explicit-sendable"]))
```
FranzBusch, line 82: We can't do this since unsafe flags are not allowed in SemVer stable packages

Sources/Logging/Logging.swift
```diff
@@ -734,7 +734,6 @@ public enum LoggingSystem {
     ///
     /// - parameters:
     ///     - factory: A closure that provides a ``Logger`` label identifier and produces an instance of the ``LogHandler``.
-    @preconcurrency
```
FranzBusch, line 737: This is an API break for anything that was passing a non-Sendable closure here. The `@preconcurrency` annotation downgrades the error to a warning. So we can't remove this or any other `@preconcurrency` annotation

Package.swift
ktoso: That's not true anymore since 6.2 https://github.com/swiftlang/swift-package-manager/pull/8951 / https://forums.swift.org/t/pitch-disable-checks-for-unsafe-flags-in-swiftpm/80698/27 so if we wanted to we could do so in Package+6.2

kukushechkin: I've added `Package@swift-6.1.swift` without the `.unsafeFlags` and the `Package.swift` now has `.unsafeFlags`.

czechboy0 (approved):

PR #405, opened 2026-01-25
[SLG-0003]: Standardized Error Metadata via Logger Convenience

samuelmurray:
Add standardized way of attaching data from Error instances to log posts.

### Motivation:

Addresses https://github.com/apple/swift-log/issues/291

### Modifications:

The "SLG-0003" proposal doc added.

### Result:

The proposal is ready for review.

### DIFF
--- a/Package@swift-6.0.swift
+++ b/Package@swift-6.0.swift
@@ -41,24 +41,24 @@ let package = Package(
     ]
 )
 
-for target in package.targets {
+for target in package.targets
+where [.executable, .test, .regular].contains(
+    target.type
+) {
     var settings = target.swiftSettings ?? []
-    settings.append(.enableExperimentalFeature("StrictConcurrency=complete"))
+
+    // https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md
+    // Require `any` for existential types.
+    settings.append(.enableUpcomingFeature("ExistentialAny"))
+
+    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
+    settings.append(.enableUpcomingFeature("MemberImportVisibility"))
+
+    // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md
+    settings.append(.enableUpcomingFeature("InternalImportsByDefault"))
+
+    // https://docs.swift.org/compiler/documentation/diagnostics/nonisolated-nonsending-by-default/
+    settings.append(.enableUpcomingFeature("NonisolatedNonsendingByDefault"))
+
     target.swiftSettings = settings
 }
-
-// ---    STANDARD CROSS-REPO SETTINGS DO NOT EDIT   --- //
-for target in package.targets {
-    switch target.type {
-    case .regular, .test, .executable:
-        var settings = target.swiftSettings ?? []
-        // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
-        settings.append(.enableUpcomingFeature("MemberImportVisibility"))
-        target.swiftSettings = settings
-    case .macro, .plugin, .system, .binary:
-        ()  // not applicable
-    @unknown default:
-        ()  // we don't know what to do here, do nothing
-    }
-}
-// --- END: STANDARD CROSS-REPO SETTINGS DO NOT EDIT --- //
--- a/Package@swift-6.1.swift
+++ b/Package@swift-6.1.swift
@@ -78,8 +78,5 @@ where [.executable, .test, .regular].contains(
     // https://docs.swift.org/compiler/documentation/diagnostics/nonisolated-nonsending-by-default/
     settings.append(.enableUpcomingFeature("NonisolatedNonsendingByDefault"))
 
-    // Ensure all public types are explicitly annotated as Sendable or not Sendable.
-    settings.append(.unsafeFlags(["-Xfrontend", "-require-explicit-sendable"]))
-
     target.swiftSettings = settings
 }
--- a/Sources/InMemoryLogging/InMemoryLogHandler.swift
+++ b/Sources/InMemoryLogging/InMemoryLogHandler.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Logging
+public import Logging
 
 /// A custom log handler which just collects logs into memory.
 /// You can then retrieve an array of those log entries.
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -712,7 +712,7 @@ extension Logger {
 ///
 /// The default (``StreamLogHandler``) is intended to be a convenience.
 /// For production applications, implement the ``LogHandler`` protocol directly, or use a community-maintained backend.
-public enum LoggingSystem {
+public enum LoggingSystem: Sendable {
     private static let _factory = FactoryBox(
         { label, _ in StreamLogHandler.standardError(label: label) },
         violationErrorMesage: "logging system can only be initialized once per process."
--- a/Tests/LoggingTests/LoggingTest.swift
+++ b/Tests/LoggingTests/LoggingTest.swift
@@ -443,13 +443,13 @@ struct LoggingTest {
         private var makeValue: (() -> String)?
         private var _value: String?
 
-        public init(_ makeValue: @escaping () -> String) {
+        init(_ makeValue: @escaping () -> String) {
             self.makeValue = makeValue
         }
 
         /// This allows caching a value in case it is accessed via an by name subscript,
         // rather than as part of rendering all metadata that a LoggingContext was carrying
-        public var value: String {
+        var value: String {
             if let f = self.makeValue {
                 self._value = f()
                 self.makeValue = nil
@@ -459,7 +459,7 @@ struct LoggingTest {
             return self._value!
         }
 
-        public var description: String {
+        var description: String {
             "\(self.value)"
         }
     }
@@ -1292,7 +1292,7 @@ struct LoggingTest {
         LoggingSystem.bootstrap { (label: String) in StreamLogHandler.standardOutput(label: label) }
         LoggingSystem.bootstrap { (label: String) in StreamLogHandler.standardError(label: label) }
 
-        // with metadata handler, explicitly, public api
+        // with metadata handler, explicitly, api
         LoggingSystem.bootstrap(
             { label, metadataProvider in
                 StreamLogHandler.standardOutput(label: label, metadataProvider: metadataProvider)
@@ -1337,7 +1337,7 @@ struct LoggingTest {
 }
 
 extension Logger {
-    public func error(
+    func error(
         error: any Error,
         metadata: @autoclosure () -> Logger.Metadata? = nil,
         file: String = #fileID,
--- a/Tests/LoggingTests/MetadataProviderTest.swift
+++ b/Tests/LoggingTests/MetadataProviderTest.swift
@@ -74,13 +74,13 @@ extension Logger.MetadataProvider {
     }
 }
 
-public struct LogHandlerThatDidNotImplementMetadataProviders: LogHandler {
+struct LogHandlerThatDidNotImplementMetadataProviders: LogHandler {
     let testLogging: TestLogging
     init(testLogging: TestLogging) {
         self.testLogging = testLogging
     }
 
-    public subscript(metadataKey _: String) -> Logging.Logger.Metadata.Value? {
+    subscript(metadataKey _: String) -> Logging.Logger.Metadata.Value? {
         get {
             nil
         }
@@ -89,11 +89,11 @@ public struct LogHandlerThatDidNotImplementMetadataProviders: LogHandler {
         }
     }
 
-    public var metadata: Logging.Logger.Metadata = [:]
+    var metadata: Logging.Logger.Metadata = [:]
 
-    public var logLevel: Logging.Logger.Level = .trace
+    var logLevel: Logging.Logger.Level = .trace
 
-    public func log(
+    func log(
         level: Logger.Level,
         message: Logger.Message,
         metadata: Logger.Metadata?,
@@ -114,13 +114,13 @@ public struct LogHandlerThatDidNotImplementMetadataProviders: LogHandler {
     }
 }
 
-public struct LogHandlerThatDidImplementMetadataProviders: LogHandler {
+struct LogHandlerThatDidImplementMetadataProviders: LogHandler {
     let testLogging: TestLogging
     init(testLogging: TestLogging) {
         self.testLogging = testLogging
     }
 
-    public subscript(metadataKey _: String) -> Logging.Logger.Metadata.Value? {
+    subscript(metadataKey _: String) -> Logging.Logger.Metadata.Value? {
         get {
             nil
         }
@@ -129,13 +129,13 @@ public struct LogHandlerThatDidImplementMetadataProviders: LogHandler {
         }
     }
 
-    public var metadata: Logging.Logger.Metadata = [:]
+    var metadata: Logging.Logger.Metadata = [:]
 
-    public var logLevel: Logging.Logger.Level = .trace
+    var logLevel: Logging.Logger.Level = .trace
 
-    public var metadataProvider: Logger.MetadataProvider?
+    var metadataProvider: Logger.MetadataProvider?
 
-    public func log(
+    func log(
         level: Logger.Level,
         message: Logger.Message,
         metadata: Logger.Metadata?,
--- a/Tests/LoggingTests/TestLogger.swift
+++ b/Tests/LoggingTests/TestLogger.swift
@@ -54,7 +54,7 @@ internal struct TestLogHandler: LogHandler {
     private var logger: Logger  // the actual logger
 
     let label: String
-    public var metadataProvider: Logger.MetadataProvider?
+    var metadataProvider: Logger.MetadataProvider?
 
     init(label: String, config: Config, recorder: Recorder, metadataProvider: Logger.MetadataProvider?) {
         self.label = label
@@ -125,7 +125,7 @@ internal struct TestLogHandler: LogHandler {
         }
     }
 
-    public var metadata: Logger.Metadata {
+    var metadata: Logger.Metadata {
         get {
             self._metadata
         }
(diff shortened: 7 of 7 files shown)


Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+        file: String,
+        function: String,
+        line: UInt
+    ) {
+        // process log post and serialize error however you like
+    }
+}
+```
+
+This was rejected due to breaking the existing abstraction layer between `Logger` and `LogHandler`, while also being questionable how much value it would add.
```
AllDmeat, line 407: It would be great if LogHandler could accept Error objects directly. That would let us forward non-fatal errors to Crashlytics (or similar services) and improve monitoring in apps.

samuelmurray: I agree, however that's out of scope for this proposal. That was my original idea in https://github.com/apple/swift-log/pull/384, but as per suggested in https://github.com/apple/swift-log/pull/384#issuecomment-3632258826 we split it in two phases to help moving forward. That is, if/when this is implemented, I might add a follow-up proposal, though from the discussion in my previous PR it’s not obvious if that’s a direction `swift-log` wants to go in.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+The above is a reasonable way to log a generic error, but it is unlikely that everyone will use the same metadata keys. It is also quite a lot of code, for something that feels like it should be easy to do - i.e. attach an error to a log.
+
+Finally, even if you log errors in a consistent way in your code, you do not control logs generated in other modules/packages that you call. In practice, this makes it so that different parts of your application could log error information under different keys, which, again, makes it harder for tooling.
+
+### Proposed solution
+
+The solution we propose is to add convenience methods in `Logger` that allows the caller to do the following:
+
+```swift
+logger.warning("Something went wrong", error: error)
```
kukushechkin: Please, provide an example with the `do-catch` context, like in the "Motivation" section.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+Under the hood, this would call 
+
+```swift
+logger.warning("Something went wrong", metadata: ["error.message": "\(error)", "error.type": "\(String(reflecting: type(of: error)))"])
```
kukushechkin: I would avoid spelling out an incorrect implementation, but rather explain the real mechanism of merging error-related metadata with the explicit metadata provided in the call.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+                return metadata
+            }(),
+            source: source(), 
+            file: file, 
+            function: function, 
+            line: line
+        )
+    }
+
+public func critical(
```
kukushechkin: Please, also add all the new proposed methods to the list. The implementations are not that important, as the proposal is about the public API design.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+        metadata: @autoclosure () -> Logger.Metadata? = nil,
+        source: @autoclosure () -> String? = nil,
+        file: String = #fileID,
+        function: String = #function,
+        line: UInt = #line
+    ) {
+        self.log(
+            level: level, 
+            message(), 
+            metadata: {
```
kukushechkin: Similar to the situation above — I would avoid an explicit implementation in the API design section. The implementation will have it's own PR :)

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+            metadata: metadata(),
+            source: source(),
+            file: file,
+            function: function,
+            line: line
+        )
+    }
+```
+
+We choose the keys `"error.message"` and `"error.type"` since they correspond to what [ECS](https://www.elastic.co/docs/reference/ecs/ecs-error) and [OpenTelemetry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/error/#error-type) use.
```
kukushechkin, line 332: I would at discuss and at least add to the alternatives introduction of something like `enum WellKnownSwiftLogMetadataKey: String` (just with a better name) to have it as an explicit part of the interface between `Logger` and `LogHandler`. This would make `LogHandler` life easier if they want to do something special with this metadata, like rename the keys. This would support the "keys ... should be considered stable and part of the `swift-log` API" of the proposal.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+Add standardized way of attaching data from `Error` instances to log posts.
+
+## Overview
+
+- Proposal: SLG-0003
+- Author(s): [Samuel Murray](https://github.com/samuelmurray)
+- Status: **Awaiting Review**
+- Issue: [apple/swift-log#291](https://github.com/apple/swift-log/issues/291)
+- Implementation:
+    - TBD
```
kukushechkin: Would you mind providing also a separate branch and a Draft PR as the proof of concept? It does not need to be fully polished at this stage, but at least be functional, proving the design does what it is supposed to accomplish.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+* It is obvious for the caller _how_ to attach an `Error` to a log (and the caller code also looks cleaner)
+* We get a consistent format for logged errors
+
+Of course, it might take time for this feature to be adopted, but the proposed solution is fully backwards compatible, meaning there is no added downside of having some parts of your code calling the new convenience methods, and other parts manually attaching the error.
+
+### Detailed design
+
+The proposed solution is to add a new overload for `Logger.log`, that accepts an `Error` as its third argument and serializes it, adding the type and message to the `metadata` parameter.
+
+Additionally, we add a new overload for each log method (`trace`, ..., `critical`) that accepts an `Error` as its second argument, that in turn calls the new `log` method.
```
kukushechkin: I would switch places of `Logger.log` overload and overloads for the tracing shortcut methods — log-level-specific methods are the primary public API users use and being used throughout this doc.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+
+```swift
+public func log(
```
kukushechkin: We usually put all the public API changes in the Detailed Design section, as they would be in the code with the accompanying doc strings.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+
+We propose that the metadata keys used (`"error.message"` and `"error.type"`) should be considered stable and part of the `swift-log` API. This makes it so that `LogHandler`s can reliably access this information. This is useful if a `LogHandler` wants to output the logs in a format which uses other keys for error information.
+
+### API stability
+
+The proposed additions are fully backwards compatible, meaning that all existing code will continue to function as is. Additionally, a user of `Logger` can start using the new convenience methods, without any modifications of the `LogHandler`, since a `LogHandler` is already expected to handle `Metadata`. 
+
+A `LogHandler` implementation is free to "ignore" the new metadata keys, and treat them as any other metadata. Alternatively, it can use the keys to e.g. re-map the error data to some other keys, such as `"err.message"` and `"err.type"`. However, it should be noted that only errors passed via the new convenience methods (or manually serialized using the same keys) will be captured this way. That is, the following is still valid:
```
kukushechkin: "ignore" might give an impression `LogHandler` dropping these keys or need some action to actively "ignore" the keys. I suggest formulation this in a way that adding specific metadata keys is transparent for Log Handlers.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+There are two clear gains from this: 
+
+* It is obvious for the caller _how_ to attach an `Error` to a log (and the caller code also looks cleaner)
+* We get a consistent format for logged errors
+
+Of course, it might take time for this feature to be adopted, but the proposed solution is fully backwards compatible, meaning there is no added downside of having some parts of your code calling the new convenience methods, and other parts manually attaching the error.
+
+### Detailed design
+
+The proposed solution is to add a new overload for `Logger.log`, that accepts an `Error` as its third argument and serializes it, adding the type and message to the `metadata` parameter.
```
czechboy0: I think we should also add to this proposal a way to set the error metadata on a Logger:

```swift
extension Logger {
    public mutating func recordError(_ error: any Error) {
        self[metadataKey: "error.type"] = "\(String(reflecting: type(of: error)))"
        self[metadataKey: "error.message"] = "\(error)"
    }
}
```

This means the user can do any of the following:

```swift
let error: any Error = ...
let logger: Logger = ...

// Option 1:
logger.info("Stuff went wrong, error: error)

// Option 2:
logger.recordError(error)
logger.info("Stuff went wrong")
```

We generally want to allow adopters to attach metadata on the logger as well, to make it easier to pass loggers with appropriate metadata to be passed into nested functions.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,186 @@
...
+
+The proposed additions are fully backwards compatible, meaning that all existing code will continue to function as is. Additionally, a user of `Logger` can start using the new convenience methods, without any modifications of the `LogHandler`, since a `LogHandler` is already expected to handle `Metadata`. 
+
+A `LogHandler` implementation is free to "ignore" the new metadata keys, and treat them as any other metadata. Alternatively, it can use the keys to e.g. re-map the error data to some other keys, such as `"err.message"` and `"err.type"`. However, it should be noted that only errors passed via the new convenience methods (or manually serialized using the same keys) will be captured this way. That is, the following is still valid:
+
+```swift
+logger.warning("Valid log", metadata: ["error": "\(error)"])
+```
+
+There is also no guarantee that the keys are not used manually with other data, i.e. the following is still valid (though unlikely to be an issue in reality):
```
kukushechkin: As this is the "API Stability" section, it should focus on the fact that if the keys can still be used explicitly. The interesting question will be for the implementation — how are they combined if both provided in the metadata and extracted from an error.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: I’m not sure what this comment suggests - these are in `Detailed design`. Did you mean to say `Proposed solution`?

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: Yeah I agree, ignore was not a good word to use

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: I’ll add this to “Alternatives considered"

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: Sure!

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: I restructured it slightly, please let me know what you think.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
kukushechkin: Than you for adding. Could you also add a rejection reason for it?

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
kukushechkin: Oh, sorry for being unclear. I suggest adding the new methods **and** their docc comments, as this is also part of the "public interface" of the package.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,236 @@
...
+logger.warning("Valid log", metadata: ["error.message": ["msg1", "msg2"]])
+```
+
+This suggests that a `LogHandler` that wants to extract error information from the metadata must ensure that the expected types are passed. For example, it could opt to re-map any error metadata that contain only strings, and leave any other metadata using those keys untouched.
+
+### API stability
+
+The proposed additions are fully backwards compatible, meaning that all existing code will continue to function as is. Additionally, a user of `Logger` can start using the new convenience methods, without any modifications of the `LogHandler`, since a `LogHandler` is already expected to handle `Metadata`. 
+
+If an `Error` is passed using the new convenience methods, and the same keys (`"error.message"`, `"error.type"`) are added as metadata on the logger (or on a metadata provider), it is up to the `LogHandler` how to merge these metadata. This is not a new issue, as the same may already happen today, if the caller manually serializes the error using those keys. Really, the only potential issue is if a caller uses the new methods to pass an `Error`, while also passing metadata with the same keys.
```
kukushechkin:
```suggestion
A potential metadata collision exists when an `Error` is passed via the new methods while the same keys (`"error.message"`, `"error.type"`) are present in metadata:
```

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,236 @@
...
+
+```swift
+logger.warning("Something went wrong", error: error, metadata: ["error.message": "some other info"])
+```
+
+We suggest that in this case, the `error` takes precedence, and the `"error.message"` from `metadata` is discarded. We don't expect this to be an issue in reality.
```
kukushechkin:
```suggestion
We recommend that the `error` parameter takes precedence in such cases, with conflicting metadata values discarded. This aligns with the principle that explicit `error:` parameters represent primary data, while the same keys in metadata are likely incidental. In practice, we don't expect callers to deliberately pass conflicting error information.
```

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: Oh sorry, I didn’t realize that “Alternatives considered” was only meant for rejected topics. I like the idea, so let me move it to a more proper place

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: I see, added it now! Takes quite a bit of space with all the different methods 😅

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: Should I add it already at this stage, or would that be part of the review?

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -0,0 +1,379 @@
...
+A draft implementation can be found in [apple/swift-log#406](https://github.com/apple/swift-log/pull/406).
+
+We choose the keys `"error.message"` and `"error.type"` since they correspond to what [ECS](https://www.elastic.co/docs/reference/ecs/ecs-error) and [OpenTelemetry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/error/#error-type) use.
+
+We propose that the metadata keys used (`"error.message"` and `"error.type"`) should be considered stable and part of the `swift-log` API. This makes it so that `LogHandler`s can reliably access this information. This is useful if a `LogHandler` wants to output the logs in a format which uses other keys for error information (e.g. re-map the error to `"err.message"` and `"err.type"`). To facilitate this, we suggest adding an enum with the metadata key names.
+
+```swift
+extension Logger {
+    /// Metadata keys with specific use-cases
+    public enum WellKnownMetadataKey: String {
```
kukushechkin: There is a good chance this enum will be extended with more keys in the future, so it should be up for backward compatible evolution. One option is [`@nonexhaustive`](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0487-extensible-enums.md), but this limits the supported Swift versions. Or a `struct` with `static` properties similar to [this](https://github.com/apple/swift-distributed-tracing/blob/main/Sources/InMemoryTracing/InMemoryTracer.swift#L197). I would lean towards the second option as `LogHandler` would not need to match over all the special keys supported by the `Logger`, but would rather reference a few interesting ones.

What do you think @czechboy0?

czechboy0: I agree that this shouldn't be an enum. I don't think we should use `@nonexhaustive` since it requires Swift 6.2.3, but Swift Log needs to support older versions. So a struct with static vars is probably best.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
kukushechkin: I think this can be added to the proposal, whenever there is more than one log statement (e.g. an `.error` and then a `.debug` with more data) people will use it. Could probably be `@inlinable`.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: Good point, changed to a struct.

Sources/Logging/Docs.docc/Proposals/SLG-0003.md
samuelmurray: Added it to the proposal

kukushechkin:
@samuelmurray thank you for addressing the comments! The proposal looks good for the public review now. Let's wait until the Monday for any more suggestion, and if none arise, I will merge this and start the formal review at https://forums.swift.org.

PR #406, opened 2026-01-28
POC: Implementation of SLG-0003

samuelmurray:
Draft implementation of SLG-0003 https://github.com/apple/swift-log/pull/405

### DIFF
--- a/Sources/Logging/Logging.swift
+++ b/Sources/Logging/Logging.swift
@@ -232,8 +232,8 @@ extension Logger {
             message(),
             metadata: {
                 var metadata = metadata() ?? Metadata()
-                metadata[WellKnownMetadataKey.errorMessage.rawValue] = "\(error)"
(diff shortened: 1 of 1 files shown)


Sources/Logging/Logging.swift
```diff
@@ -257,6 +310,45 @@ extension Logger {
...
+        self.log(
+            level: .trace,
+            message(),
+            error: error,
+            metadata: metadata(),
+            source: source(),
+            file: file,
+            function: function,
+            line: line
+        )
```
samuelmurray, line 349: It’s a bit unfortunate that we don’t call the `trace` method with `#if !MaxLogLevelDebug && !MaxLogLevelInfo && !MaxLogLevelNotice && !MaxLogLevelWarning && !MaxLogLevelError && !MaxLogLevelCritical && !MaxLogLevelNone`, but since we add a `log` overload as well, I thought it would make sense to call that, and so only do the metadata merging at one place.

samuelmurray:
Replaced with https://github.com/apple/swift-log/pull/425

PR #387, opened 2025-11-06
Add support for FreeBSD and OpenBSD

kkebo:
I've added support for FreeBSD. I've only tested this PR on FreeBSD, but I believe it should also work on OpenBSD.

### Motivation:

closes #372 

Also, [Swift on FreeBSD preview toolchain](https://forums.swift.org/t/swift-on-freebsd-preview/83064) is now available.

### Modifications:

- changed pthread-related code for *BSD
- changed `Glibc.stdout` and `Glibc.stderr` for *BSD because they are non-optional on *BSD

### Result:

swift-log and packages depending on it will be usable on FreeBSD.

### Checks:

I tested it with `swift build` and `swift test`. All tests have passed.

```console
$ swift test
...
Test Suite 'All tests' passed at 2025-11-07 04:29:34.003                                                                                         
         Executed 0 tests, with 0 failures (0 unexpected) in 0.0 (0.0) seconds
...
✔ Test run with 52 tests in 7 suites passed after 0.309 seconds.
```

Environment:

```console
$ uname -a
FreeBSD Coconut-rhinoceros-beetle 15.0-BETA4 FreeBSD 15.0-BETA4 releng/15.0-n280841-a7707f2a3bf4 GENERIC amd64
$ swift --version
Swift version 6.3-dev (LLVM bff1370bd79c983, Swift 57cf4ce563f700b)
Target: x86_64-unknown-freebsd14.3
Build config: +assertions
```

Swift toolchain is from https://forums.swift.org/t/swift-on-freebsd-preview/83064.

PR #408, opened 2026-02-04
chore: update Lock implementation with latest changes from swift-nio

scottmarchant:
Ports changes to `Lock` from swift-nio to swift-log.

### Motivation:

The copy-paste of `Lock` in swift-log should be updated with changes from swift-nio.

### Modifications:

- Rolls back changed [made previously for FreeBSD](https://github.com/apple/swift-log/pull/387/changes#diff-7b1bd45403dd1a7418287bd60682bf05ba2e4bf75dca1c0ca55a86d0477c9af5L73) in favor of a consolidated implementation in swift-nio that will be in place once https://github.com/apple/swift-nio/pull/3494 merges.  @kkebo Please confirm this works. After both of our PR's merge, swift-log and swift-nio will have the same implementation as far as FreeBSD is concerned. 
- Brought the following recent changes from swift-nio over to swift-log:

  - https://github.com/apple/swift-nio/pull/3482
  - https://github.com/apple/swift-nio/pull/3480
  - https://github.com/apple/swift-nio/pull/3483

- Removed stale `// SRWLOCK does not need to be free'd` comment.

### Result:

`Lock` implementation will have the same implementation as swift-nio (assuming https://github.com/apple/swift-nio/pull/3494 merges).

### DIFF
--- a/Tests/LoggingTests/LockTest.swift
+++ b/Tests/LoggingTests/LockTest.swift
@@ -1,0 +1,74 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Logging API open source project
+//
+// Copyright (c) 2018-2026 Apple Inc. and the Swift Logging API project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Logging API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Testing
+
+@testable import Logging
+
+#if canImport(Dispatch)
+import Dispatch
+private let canImportDispatch = true
+#else
+private let canImportDispatch = false
+#endif
+
+@Suite("Lock Test Suite")
+struct LockTestSuite {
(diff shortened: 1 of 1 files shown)


Sources/Logging/Locks.swift
```diff
@@ -70,34 +70,38 @@ package final class Lock {
     package init() {
         #if os(Windows)
         InitializeSRWLock(self.mutex)
-        #elseif (compiler(<6.1) && !os(WASI)) || (compiler(>=6.1) && _runtime(_multithreaded))
-        #if os(FreeBSD) || os(OpenBSD)
+        #elseif os(FreeBSD) || os(OpenBSD)
```
czechboy0: I think we're still missing a call to pthread_mutexattr_destroy in the FreeBSD/OpenBSD branch.

kkebo: I have a question. Why is swift-log using swift-nio's deprecated `Lock`, not the current `NIOLock`? `NIOLock` doesn't have such an issue.

kkebo: Anyway, I've added the fix for it to apple/swift-nio#3494. Thank you.

https://github.com/apple/swift-nio/pull/3494/commits/bcc8d0593dd3847516e6a0d0fe707bc67dc87f4a

scottmarchant: @kkebo Thanks, I updated this PR to match the lock implementation in your PR.

scottmarchant: > Why is swift-log using swift-nio's deprecated Lock, not the current NIOLock? NIOLock doesn't have such an issue.

@kkebo I've wondered the same thing. The obvious possible reason I've thought of is that the new lock is named `NIOLock`, and the repos where `Lock` is copied don't have anything to do with NIO, so it would be weird for them to have a thing called `NIOLock` in them.

Honestly, I've wondered why we don't have set up an official `Lock` and `RecursiveLock` in `Synchronization`. This business of copy-pasting a `Lock` implementation between several repos to avoid inter-dependency isn't the best solution compared to mainstreaming the implementation either in `Synchronization` or a standalone package.

Anyways, for now, we at least need to keep the implementations in sync. Once this PR merges, I'm planning to synchronize the changes to `swift-metrics` and `swift-distributed-tracing` as well (both have copy-pastes of this as well.

czechboy0 (approved):

PR #418, opened 2026-02-18
SLG-0004: Metadata privacy levels implementation

kukushechkin:
This implements the SLG-0004 proposal introducing an **attributed metadata system** that allows attaching attributes to metadata values. Privacy level is the first concrete attribute, enabling developers to mark metadata values as `.private` or `.public` so `LogHandler` implementations can redact sensitive values before logging.

### Motivation:

SwiftLog lacks a mechanism to attach attributes to metadata values as sensitive to allow Log Handlers to control logging behavior.

### Modifications:

- Attributed metadata added.
- Privacy levels can be attached to Attributed Metadata via string interpolation.
- `Logger` and `LogHandler` have new API surface for using Attributed Metadata.

### Result:

Sensitive metadata values can be marked as `.private` to signal compatible Log Handlers to mask the value when logging.


kukushechkin:
Closing in favor of a revised version #439

PR #419, opened 2026-02-25
Update proposal SLG-0003 after public review

samuelmurray:
Motivation:

The public review of the original proposal
(https://forums.swift.org/t/proposal-slg-0003-standardized-error-metadata-via-logger-convenience/84518) is done, and these changes adjust for what was decided on.

Modifications:

The main difference is to pass Error instances as-is to LogHandlers, rather than serializing it to metadata in Logger.


Sources/Logging/Docs.docc/Proposals/SLG-0003.md
```diff
@@ -1,6 +1,6 @@
-# SLG-0003: Standardized Error Metadata via Logger Convenience
+# SLG-0003: Support for Attaching Errors to Log Entries
```
samuelmurray, line 1: I was unsure about updating the name, but also felt that the original name didn’t really fit anymore.

ktoso: Sounds good to me :)

PR #439, opened 2026-04-09
[SLG-0004]: Revised Metadata value attributes + sensitivity attribute implementation

kukushechkin:
> [!IMPORTANT]
> This is a revision of SLG-0004 previously discussed [here](https://forums.swift.org/t/proposal-slg-0004-metadata-values-privacy-attribute/85249/15). Note this is a draft, including more than just the proposal implementation.

Add an extensible per-value attribute mechanism for metadata, with sensitivity as the first attribute.

### Motivation:

Metadata values in `swift-log` are opaque strings by the time the `LogHandler` receives them. The call site often
knows things about a value that the handler cannot infer — for example, whether the value should be redacted in
different environments.

### Modifications:

- SLG-0004 proposal is updated based on the feedback. 
- Added generic `AttributedMetadataValue`-based mechanism to `Logging`.
- Added `Sensitivity` attribute to the new `LoggingSensitivity` target.
- Added `OSLogHandler` as an example of a concrete log handler taking advantage of the `Sensitivity` attribute.
- Added benchmarks and benchmark thresholds for the new attributed metadata.

### Result:

- `swift-log` provides an optional extensible mechanism for attaching attributes to metadata values.
- `swift-log` provides a common `Sensitivity` attribute and examples how to use it.


Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
-# SLG-0004: Metadata values privacy attribute
+# SLG-0004: Metadata value attributes
```
weissi: nit: The file name still says "privacy"

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-1. **`AttributedMetadataValue`**: Wraps a standard `MetadataValue` with associated attributes.
-2. **`MetadataValueAttributes`**: Container for attributes.
-3. **`AttributedMetadata`**: Dictionary type mapping `String` keys to attributed values.
-4. **Parallel logging API**: New log methods accepting `attributedMetadata` parameter.
+Attributes are **classifications, not enforcement**. A sensitivity attribute does not guarantee redaction — the handler
+may not support it. Applications needing guaranteed PII protection must enforce it at the application level.
 
-This design separates the transport mechanism (attributed metadata) from specific attribute semantics (privacy labels), allowing future extensions without breaking existing attributed metadata code.
+The attribute mechanism is extensible: handler packages and middleware can define their own attribute types using the
+same `MetadataAttributeKey` protocol. A `LogHandler` middleware can read the attributes it cares about, act on them,
```
weissi: SwiftLog doesn't have a middleware logger concept. You could implement one but I think that's besides the point.

I reckon this is meant to be an example but I think it's very confusing. Anything relating to middlewares is squarely a "middle end" concern and not a concern of the the deliberately as-policy-free-as-possible SwiftLog. I would advise to remove this, there are no middlewares, let's not invent them in an unrelated proposal.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-        /// String representation redacts private values to the redaction marker.
-        public var description: String {
-            attributes.privacy == .public ? value.description : Self.redactionMarker
-        }
-
-        public init(_ value: MetadataValue, attributes: MetadataValueAttributes)
-        public init(_ value: MetadataValue, privacy: PrivacyLevel)
-    }
+    public protocol MetadataAttributeKey: Sendable,
+        RawRepresentable where RawValue == Int {}
```
weissi: This is one of the very few instances where I would think that a fixed-width integer type (e.g. `Int64`) would maybe make sense? What if somebody adds `case special = 9223372036854775807` (64 `1` bits), then this would only work in 64 bit environments.

This is a fairly minor concern but I do think this is meant to potentially be used with bit masks, so this could be a concern I think.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-    /// `AttributedMetadataValue` wraps a standard `MetadataValue` with privacy attributes,
-    /// allowing you to mark metadata as private or public for privacy-aware logging.
+    /// This protocol is designed for **small, fixed-vocabulary attributes** represented as
+    /// enums. Each attribute value is stored as an `Int` raw value alongside an `ObjectIdentifier`
+    /// for the key type, so attributes occupy minimal space. Attributes that need to carry
+    /// associated data, strings, or richer payloads are outside the scope of this protocol.
     ///
-    /// ## Creating Attributed Metadata
+    /// The conforming type must be an enum with `Int` raw values. Each attribute occupies just
+    /// `(ObjectIdentifier, Int)` in the inline slot, avoiding heap allocation for the
```
weissi: This unnecessarily leaks the guts. I don't think we should reveal the implementation here. The int-bit is fine but the `ObjectIndentifier` should be opaque IMHO

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-            public init(privacy: PrivacyLevel = .public)
-        }
-
-        /// The underlying metadata value without privacy attributes.
-        public var value: MetadataValue
-
-        /// The privacy attributes associated with this metadata value.
-        public var attributes: MetadataValueAttributes
+Each conforming type serves as both the key and the value for an attribute. The metatype (`Key.Type`) identifies
+the attribute at runtime via `ObjectIdentifier` — no coordination between packages is needed.
```
weissi: no need to leak how we're storing the types I believe. The only thing the user knows is that it's identified by `Key.Type` -- and crucially not how we're storing it.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
+}
+```
 
-    /// Metadata dictionary with privacy attributes.
-    ///
-    /// A dictionary mapping string keys to ``AttributedMetadataValue`` instances, used with
-    /// the `attributedMetadata` parameter of logging methods.
-    ///
-    /// ## Example
+The common case (0–1 attributes) avoids heap allocation. `Equatable` compares entries as an unordered set,
```
weissi: good callout. I think this is fine.

weissi: But honestly, can't you store them sorted anyway? We shouldn't reveal this but you can store by the pointer value of `Key.Type`. Yes, the ordering won't be stable across executions but within one execution it's stable. That way it'd even be O(n)

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-        /// Interpolation with explicit privacy parameter.
-        public mutating func appendInterpolation<T>(_ value: T, privacy: Logger.PrivacyLevel = .public)
-            where T: CustomStringConvertible
-    }
+Attribute packages add ergonomic overloads. The `LoggingAttributes` target adds:
+
+```swift
+/// Interpolation with explicit sensitivity parameter.
+public mutating func appendInterpolation<T: CustomStringConvertible & Sendable>(
+    _ value: T, sensitivity: Logger.Sensitivity
```
weissi: It's a mistake to bundle two things together:
- A general attribution system
- A random application for it

I think it's extremely important not to bundle them. This proposal should be the introduction of a generalised attribution system. And I like the idea and +1.

I'm definitely -1 to introduce the sensitivity itself into `swift-log`. This should come as a separate package such as `swift-log-sensitivity` or similar. Otherwise we're setting a bad precedent that everything should live in `swift-log`. IMHO it really should not.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
+)
+```
 
-    /// Creates an attributed metadata value from a string literal (defaults to .public).
-    public init(stringLiteral value: String)
+When a single `AttributedMetadataValue` contains multiple interpolated segments with different attributes, the merging
+behavior is defined by each attribute's `appendInterpolation` implementation. For sensitivity, the strictest level
+wins — if any segment is `.sensitive`, the entire value is `.sensitive`.
```
weissi: This is a prime example of what SwiftLog must _not_ do: Policy.

I'm not saying the suggested policy is wrong or bad. But SwiftLog is the _mechanism_ and we worked very hard to separate the _policy_ as much as possible. See https://en.wikipedia.org/wiki/Separation_of_mechanism_and_policy

weissi: In this case the fix is straightforward, take out the privileged position of sensitivity, put it into another package and solve it there.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-    /// Creates an attributed metadata value from string interpolation.
-    public init(stringInterpolation: StringInterpolation)
+#### `Logger` API additions
+
+New attributed metadata properties and subscripts on `Logger`:
+
+```swift
+extension Logger {
+    /// Get or set an attributed metadata value by key.
+    public subscript(attributedMetadataKey attributedMetadataKey: String)
```
weissi: This is quite ugly. To the user we really shouldn't separate between metadata keys that have attributes and others that have not. The user won't know.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-                     file: String = #fileID, function: String = #function, line: UInt = #line)
-    public func critical(_ message: @autoclosure () -> Message,
-                        attributedMetadata: @autoclosure () -> AttributedMetadata?,
-                        source: @autoclosure () -> String? = nil,
-                        file: String = #fileID, function: String = #function, line: UInt = #line)
+    /// Log a message with attributed metadata at the specified level.
+    public func log(
+        level: Logger.Level,
+        _ message: @autoclosure () -> Logger.Message,
+        attributedMetadata: @autoclosure () -> Logger.AttributedMetadata?,
```
weissi: This is also quite ugly. This should go into the regular metadata

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
-    ///     "user.id": "\(userId, privacy: .private)",
-    ///     "request.id": "\(requestId, privacy: .public)",
-    ///     "action": "purchase"  // String literal defaults to .public
-    /// ]
-    /// logger.info("User action", attributedMetadata: metadata)
-    /// ```
+    /// `AttributedMetadataValue` wraps a standard `MetadataValue` with custom attributes,
+    /// allowing you to associate application-defined or handler-defined attributes
+    /// with metadata values.
+    public struct AttributedMetadataValue: Sendable, Equatable, CustomStringConvertible,
```
weissi: Pretty strong opinion that we really ought not to bifurcate the core concept of metadata here. It might make evolutional sense and also facilitates the implementation if we just add another, separate layer on top.

But if we look at it with a fresh pair of eyes, it really doesn't make sense to have metadata and attributed metadata.

This should be added to `MetadataValue` itself. We shouldn't bifurcate the whole API for this.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
+
+`LogEvent` carries both plain and attributed metadata through a single internal storage enum. Handlers access whichever
+representation they need via computed properties:
 
 ```swift
-protocol LogHandler {
-    func log(level: Logger.Level, message: Logger.Message,
-            attributedMetadata: Logger.AttributedMetadata?, source: String,
-            file: String, function: String, line: UInt)
+public struct LogEvent: Sendable {
```
weissi: We also have a log event: `LogMessage`. Let's not split the whole API into two core types that are the same

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
 ```
 
-### API stability
-
-- All existing APIs unchanged.
-- Plain and attributed metadata APIs coexist.
-- No deprecation planned.
-- Adoption is optional and incremental on both application and Log Handler sides.
-- Default implementation ensures existing handlers work; logging private metadata is an application concern requiring a compatible `LogHandler`.
+The existing `get()` method continues to return plain `Metadata`. When called on an attributed provider, it strips
```
weissi: Right, we should not do this. There shouldn't be "plain" and "attributed" metadata. We have to find a way that can do both. The existing metadata should gain attributes

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
+- **Existing `LogHandler` implementations.** Attributed metadata flows through `LogEvent`, which handlers already
+  receive via `log(event:)`. Handlers that only read `event.metadata` continue to work — plain values are returned
+  with attributes stripped. Handlers that want to interpret attributes read `event.attributedMetadata` instead.
+  Existing handlers work without changes.
 
-The current design prioritizes **explicitness and data-centric privacy** over **configuration-based simplicity**. Privacy decisions are made where data is logged, making them visible during code review and keeping privacy attributes coupled to the data they protect.
+### Future directions
 
-2. **Convenience methods (`.private()` and `.public()`):** Add extension methods to `String` and `MetadataValue` for creating attributed metadata:
+- **Metrics extraction middleware.** A `LogHandler` middleware that reads a metric attribute and dual-writes to
```
weissi: Middleware doesn't exist in SwiftLog, I think this should be removed

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
+  swift-metrics. The call site annotates a value as a counter or histogram, and the middleware updates the metric then
+  forwards the attributed metadata to the next handler.
+- **Cardinality hints.** An attribute indicating whether a field is safe to index as a label/dimension in observability
+  backends. Only the call site knows a field's cardinality ahead of time.
+- **Typed metadata values.** Adding typed variants to `MetadataValue` (`.int64`, `.double`, `.bool`) would reduce the
```
weissi: This isn't a future direction for SwiftLog, it's an idea for a new package that works _with_ SwiftLog, maybe `swift-log-type-annotations`

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
+  need for attributes that compensate for stringly-typed metadata.
+- **More stored slots.** If real-world usage shows growth in the number of attributes per value, it would make sense
```
weissi: I would see this as an adjustment. Let's just not document how many are guaranteed to be allocation free. You can leave a comment saying that and how we will monitor what seems like a good amount of inline storage.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
```diff
@@ -1,467 +1,432 @@
...
+#### Bitmask storage
 
-```swift
-logger.info("User action", attributedMetadata: [
-    "user.id": "\(userId, privacy: .private)",  // Explicit private (redundant with default)
-    "action": "\(action, privacy: .public)"  // Must explicitly mark as public
-])
-```
+Pack attribute values into an inline `UInt64` using declared bit offsets. Smaller struct and O(1) access, but requires
+authors to coordinate bit layout and risks collisions between independent packages.
```
weissi: FWIW, the collisions could be avoided if SwiftLog had API to provide "your bit" to you. But yes, it's more difficult to design and might not carry its weight.

I'm happy with the "N (1 to start with) attributes are free because we store them inline" proposal

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -199,6 +199,224 @@ public struct PrintLogHandler: LogHandler {
 }
 ```
 
+#### Adopting attributed metadata in LogHandlers
```
weissi, line 183: I think the adoption should not be this difficult. What we're proposing in disguise in this proposal is to essentially create a whole second API for SwiftLog. I do not think that this is a good direction.

weissi (changes requested):
The general idea and the implementation suggestion of using a fixed-sized inline slot is good.

But overall -1 because of these concerns:

- Major concern: Complete bifurcation of SwiftLog's core APIs. There's a new `LogMessage` called `LogEvent`. There's new, duplicated metadata. This is essentially shipping a whole new version disguised as a feature addition.
- Major concern: Bundles with sensitivity, privileges sensitivity & introduces unduly policy decisions regarding sensitivity into the core mechanism package. This one's easy to fix, remove anything relating to sensitivity from this proposal and create a new package `swift-log-sensitivity-attributes` that has no special privileges
- Minor: Reveals too much about how the storage is implemented without reason

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Yeah, fixed.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Agree, middleware is a bad term here. Rephrased.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Removed.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Removed sensitivity attribute from the proposal. It is still in the PR for the reference (along with all the other examples).

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Let's use `Int64`.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Removed.

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Removed sensitivity attribute from the proposal. It is still in the PR for the reference (along with all the other examples)

Sources/Logging/Docs.docc/Proposals/SLG-0004-metadata-values-privacy-attributes.md
kukushechkin: Ok, I've spend some time exploring this. `MetadataValue` enum is unfortunate, the ideal approach would've been to add another `.attributed` case there. However, we can try to leverage `.stringConvertible` and build convenience API around it:

```
internal final class AttributedStringCarrier: CustomStringConvertible {
...
}

extension Logger.MetadataValue {
    public var attributes: Logger.MetadataValueAttributes {
        if case .stringConvertible(let box) = self,
            let carrier = box as? Logger.AttributedStringCarrier
        {
            return carrier.metadataAttributes
        }
        return .init()
    }
}
```

The API surface stays as is, API is still clean. The downside is, according to benchmarks for a [PoC](https://github.com/kukushechkin/swift-log/tree/attributed-metadata-as-stringConvertible) it is less performant (~10%) and an extra allocation. Given the fact an actual log handler (benchmarks are using a NoOp one) would do 10-100 times more, this performance difference is not important. Another gap is abusing `.stringConvertible` would make it impossible to use attributes with another type passed through `.stringConvertible`. What do you think?

kukushechkin:
Closed in favor of https://github.com/apple/swift-log/pull/453

PR #442, opened 2026-04-13
`Logger.MetadataProvider.multiplex` to not return optional

LePips:
### Motivation:

Just ran into this at work. Unless there is some reasoning I can't find on the repo or forums, this is a basic type discrepancy from [SLG-0001](https://github.com/apple/swift-log/blob/6f3fbace6629c8fcabcd56d211501e92e1919a94/Sources/Logging/Docs.docc/Proposals/SLG-0001-metadata-providers.md?plain=1#L161) and an instance is always provided anyways.


kukushechkin:
@LePips thank you for opening the PR! Unfortunately, we cannot change `Logger.MetadataProvider.multiplex(_:)` at this moment, as it is [part of the public API and would require a major semantic version bump](https://github.com/apple/swift-log/actions/runs/24359616467/job/71143298088?pr=442) to fix this oversight. Could you please instead open an issue and we will fix it in a hypothetical next major `swift-log` version?

PR #444, opened 2026-04-15
Improve readability by extracting separate types to their own files

### LINKED ISSUE #441: Improve code readability by splitting large source files
Some source files in this repo are getting quite long, especially `Logging.swift`. How about moving some code to separate files? 
For example, we could extract the included `LogHandler` implementations to separate files (similar to [providers in swift-configuration](https://github.com/apple/swift-configuration/tree/main/Sources/Configuration/Providers)), and maybe separate all the log methods to `Logging+methods.swift`.

samuelmurray:
Improve readability be extracting separate types to their own files

### Motivation:

`Logger.swift` is >2000 LoC, which makes it hard to navigate both for contributors and users of `swift-log`. Widely different concepts, such as concrete `LogHandler` implementations, are mixed with core types such as `Logger.Level`. Extracting separate types should make it easier to find what you are looking for.
Closes #441

### Modifications:

* Extract all (not `Logger`-namespaced) types from `Logging.swift` to separate files. 
* Rename `Logging.swift` to `Logger.swift`, as this is the only remaining top-level type.
* Change `LoggingSystem.factory` from `fileprivate` to `internal` to be accessible from `Logger.swift`.

### Result:

Each (non-private) top-level type lives in a separate file, named after the type.


samuelmurray:
We still have ~1000 LoC in `Logger.swift` of various log methods. To me, it would make sense to move all except the base `Logger.log(...)` and the internal `Logger._log()` to a separate file.

kukushechkin:
Thanks for tackling this, @samuelmurray!

The license headers check is failing because `StdioOutputStream.swift` uses an Xcode-generated file header instead of the project's standard license header.

But another thing that bothers me for some reason is `StdioOutputStream.swift` being in the package root, but used by the handler in `Handlers` folder. I would suggest moving `struct StdioOutputStream` into the corresponding handler.

samuelmurray:
> But another thing that bothers me for some reason is `StdioOutputStream.swift` being in the package root, but used by the handler in `Handlers` folder. I would suggest moving `struct StdioOutputStream` into the corresponding handler.

Yeah, you're right. I was a bit unsure where to put it, as it was referenced from `LoggingTest.swift`, but actually that just points to the tests could benefit from having the same separation. I'll move tests that clearly belong to a specific handler to corresponding test files.

samuelmurray:
@kukushechkin I had a typo in my commit message ("be" -> "by") - I'm not sure what ends up in the final commit message when you merge, but I didn't want to ammend/rebase during active review. Can you fix it when merging?

kukushechkin (approved):

kukushechkin:
Thank you @samuelmurray for this improvement! I've fixed "be" -> "by" in the merge commit.

PR #445, opened 2026-04-15
Update documentation to reflect recent changes

samuelmurray:
Update documentation to reflect recent changes

### Motivation:

The documentation was not updated in SLG-0003 or SLG-0005. Concretely, the examples given in the documentation used the now deprecated multiparameter methods in `LogHandler`, and the new `LogEvent`-based method were not listed among the others. Likewise, the new logging methods (`Logger.trace` etc.) that take an `Error` as a parameter were not added. Lastly, `LogEvent` had no separate entry. 
While the default rendered docs for these were acceptable, it decreased the discoverability of the new methods, and the examples using deprecated methods were misleading.

### Modifications:

* Update all code examples to use non-deprecated methods.
* List all new methods on `Logger` and `LogHandler` along the existing ones.
* Add a minimal file for `LogEvent` to give it the same look as all other types.
* Update code examples in `.md` files to follow swift format.

### Result:

Documentation reflects the current state and best practices of `swift-log`.

### DIFF
--- a/Sources/Logging/Docs.docc/ImplementingALogHandler.md
+++ b/Sources/Logging/Docs.docc/ImplementingALogHandler.md
@@ -23,11 +23,11 @@ func logHandlerValueSemantics() {
     var logger1 = Logger(label: "first logger")
     logger1.logLevel = .debug
     logger1[metadataKey: "only-on"] = "first"
-    
+
     var logger2 = logger1
     logger2.logLevel = .error                  // Must not affect logger1
     logger2[metadataKey: "only-on"] = "second" // Must not affect logger1
-    
+
     // These expectations must pass
     #expect(logger1.logLevel == .debug)
     #expect(logger2.logLevel == .error)
@@ -51,37 +51,29 @@ public struct PrintLogHandler: LogHandler {
     private let label: String
     public var logLevel: Logger.Level = .info
     public var metadata: Logger.Metadata = [:]
-    
+
     public init(label: String) {
         self.label = label
     }
-    
-    public func log(
-        level: Logger.Level,
-        message: Logger.Message,
-        metadata: Logger.Metadata?,
-        source: String,
-        file: String,
-        function: String,
-        line: UInt
-    ) {
+
+    public func log(event: LogEvent) {
         let timestamp = ISO8601DateFormatter().string(from: Date())
-        let levelString = level.rawValue.uppercased()
-        
+        let levelString = event.level.rawValue.uppercased()
+
         // Merge handler metadata with message metadata
         let combinedMetadata = Self.prepareMetadata(
             base: self.metadata
-            explicit: metadata
+            explicit: event.metadata
         )
-        
+
         // Format metadata
         let metadataString = combinedMetadata.map { "\($0.key)=\($0.value)" }.joined(separator: ",")
-        
+
         // Create log line and print to console
-        let logLine = "\(label) \(timestamp) \(levelString) [\(metadataString)]: \(message)"
+        let logLine = "\(label) \(timestamp) \(levelString) [\(metadataString)]: \(event.message)"
         print(logLine)
     }
-    
+
     public subscript(metadataKey key: String) -> Logger.Metadata.Value? {
         get {
             return self.metadata[key]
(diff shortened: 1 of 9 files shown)


Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -51,37 +51,29 @@ public struct PrintLogHandler: LogHandler {
...
+
+    public func log(event: LogEvent) {
         let timestamp = ISO8601DateFormatter().string(from: Date())
-        let levelString = level.rawValue.uppercased()
-        
+        let levelString = event.level.rawValue.uppercased()
+
         // Merge handler metadata with message metadata
         let combinedMetadata = Self.prepareMetadata(
             base: self.metadata
```
kukushechkin: This is a pre-existing issue, but worth fixing:

```suggestion
            base: self.metadata,
```

kukushechkin (approved):
Thanks for this documentation update, @samuelmurray!

heckj (approved):

heckj:
An additional thank you @samuelmurray - on top of @kukushechkin ! Appreciate you taking the time with this!

Sources/Logging/Docs.docc/ImplementingALogHandler.md
samuelmurray: Good catch! Out of curiosity, are you aware of any way to lint/check format of code blocks in `.md` or docc-comments?

Sources/Logging/Docs.docc/Reference/LogEvent.md
samuelmurray, line 1: Should this type be added to `index.md`? Currently, the type is listed separately under `Structures`.

For example:
```diff
--- a/Sources/Logging/Docs.docc/index.md
+++ b/Sources/Logging/Docs.docc/index.md
@@ -84,6 +84,7 @@ backend.
 ### Log Handlers
 
 - ``LogHandler``
+- ``LogEvent``
 - ``MultiplexLogHandler``
 - ``StreamLogHandler``
 - ``SwiftLogNoOpLogHandler``
```

kukushechkin:
FYI, the CI is in shambles (not only in this repo), there is nothing wrong with the PR.

Sources/Logging/Docs.docc/Reference/LogEvent.md
heckj: Yes, it should - thank you

Sources/Logging/Docs.docc/ImplementingALogHandler.md
heckj: There's a feature in Swift + DocC called "snippets" that's available but not well understood/documented/or used - and in particular is really awkward to use when you're using Xcode (no support in Xcode for it) - (I know, ironic)

Anyway, the idea is that you create a directory Snippets/ in the root of the project, drop executable code in there that the project (when it's using swift-docc-plugin) compiles and verifies that it compiles, then it creates bits of code based on comment marking within it that you can then reference from articles, which gives you the code blocks from code that the compiler verified.

You can see some example of it being used in https://github.com/swiftlang/swift-markdown/, where it started. I'm slowly introducing it into the various libraries like this - but haven't circled back to really set it all up yet, if you want to take a stab, I'd be happy to help answer questions and provide reviews!

heckj: (to be clear - if you want to take a run at that, I'd definitely recommend that as a follow-up PR)

samuelmurray: @heckj Sounds like I found the perfect person to ask - thank you! I can image it's quite nice to use especially for larger code examples. Also in this case, perhaps it would have help spot usage of deprecated methods in the documentation?

I think I get the basic idea from looking a `swift-markdown`, however I also noticed that the documentation differs slightly between [GH pages](https://swiftlang.github.io/swift-markdown/documentation/markdown) (which the `README.md` links to) and [swiftpackageindex.com](https://swiftpackageindex.com/swiftlang/swift-markdown/0.7.3/documentation/markdown). It seems the documentation on GH pages hasn't been updated in some time. 
What I was actually was looking for was to see if Snippets were supported/rendered correctly on swiftpackageindex - and of that I'm still not sure; the Snippets seem to be missing there, both from the left-side bar and from the "Markdown" frontpage, which is curious.

Yeah, it's outside the scope of this PR. I might play around with it some, and I really appreciate your guidance regardless of if it ends up in a PR or not!

samuelmurray:
After looking at the documentation at [swiftpackageindex](https://swiftpackageindex.com/apple/swift-log/main/documentation/logging), I realised that the `InMemoryLogging` target is missing. When building the documentation in Xcode, I get both target, but perhaps swiftpackageindex requires a `.docc` catalog?
Anyway, I could add a minimal documentation for that target as well, either in this PR or in a separate one.

heckj:
@samuelmurray the lack of those other catalogs in SPI is a result of the configuration in .spi.yaml, which currently (as I'm writing this) reads:
```yaml
version: 1
builder:
  configs:
    - documentation_targets: [Logging]
```

Which intentionally only builds the docs for the primary Logging. A catalog for InMemoryLogging would definitely help organize things, so it'd be a lovely follow up add - along with an update to the .spi.yaml to add it into Swift Package Index's set that it builds and presents

PR #449, opened 2026-04-26
Add tests for SwiftLogNoOpLogHandler

crleonard:
Add tests for SwiftLogNoOpLogHandler

### Motivation:

`SwiftLogNoOpLogHandler` is a public type with no dedicated test coverage.

### Modifications:

* Add tests for both initialisers, `logLevel`, `metadata`, the metadata subscript, `log(event:)`, and end-to-end usage via `Logger`.

### Result:

`SwiftLogNoOpLogHandler` has test coverage on par with other handlers.
Coverage from 89% over all tests, to 90%.

### DIFF
--- a/Tests/LoggingTests/SwiftLogNoOpLogHandlerTest.swift
+++ b/Tests/LoggingTests/SwiftLogNoOpLogHandlerTest.swift
@@ -48,11 +48,11 @@ struct SwiftLogNoOpLogHandlerTest {
         #expect(handler[metadataKey: "key"] == nil)
     }
 
-    @Test func logEventIsDiscarded() {
+    @Test func logEventDoesNotCrash() {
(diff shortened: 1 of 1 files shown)


kukushechkin:
@crleonard thank you for the contribution! I understand the motivation to fill the coverage gap, but I think the main value in testing `NoOpLogHandler` would be to verify it didn't do anything in `logEventIsDiscarded` and `loggerUsingNoOpHandlerSilentlyDiscards`. Right now these tests have no assertions and test this log handler does not crash. I don't see any reasonable way of doing, so recommend to change the misleading test names — the implementation does not verify anything is discarded.

crleonard:
@kukushechkin thanks for the quick review, updated the test names.

kukushechkin (approved):

Issue #450, opened 2026-04-27
CONTRIBUTORS.txt references deleted contributor-generation script


crleonard:
`CONTRIBUTORS.txt` currently says:

`Please do not edit this file manually. It is generated using `./scripts/generate_contributors_list.sh`. If a name is misspelled or appearing multiple times: add an entry in `./.mailmap`

However, `./scripts/generate_contributors_list.sh` no longer exists on main. It looks like the script was removed as part of [#328, [CI] Migrate to GHA](https://github.com/apple/swift-log/pull/328), in commit ee1bfd3.

Could you confirm the intended maintenance path for CONTRIBUTORS.txt?

If the file is no longer meant to be generated, I can raise a PR to remove or update the stale instructions, and potentially delete the generated-file wording.

If the file is still meant to be generated, should the script be restored and hooked back into the current workflow instead?

PR #452, opened 2026-04-28
Remove stale CONTRIBUTORS.txt header references

crleonard:
Remove stale `CONTRIBUTORS.txt` references from file headers.

### Motivation:

Follow-up to #451 and related to #450.

`CONTRIBUTORS.txt` was removed as part of the repository cleanup, but several source, test, and benchmark file headers still referenced it. Since the file no longer exists, those references are now stale.

### Modifications:

Removed the `See CONTRIBUTORS.txt...` header line from the remaining files across `Sources`, `Tests`, and `Benchmarks`.

### Result:

File headers no longer reference the deleted `CONTRIBUTORS.txt` file.

### DIFF
--- a/Benchmarks/MaxLogLevelWarning/Benchmarks/MaxLogLevelWarningBenchmarks/MaxLogLevelWarning.swift
+++ b/Benchmarks/MaxLogLevelWarning/Benchmarks/MaxLogLevelWarningBenchmarks/MaxLogLevelWarning.swift
@@ -6,7 +6,6 @@
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
-// See CONTRIBUTORS.txt for the list of Swift Logging API project authors
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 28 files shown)


crleonard:
@kukushechkin thanks again for picking up #450 and landing #451.

I noticed these stale refs post merge.

kukushechkin:
@crleonard thank you for pointing this out. Unfortunately we need to change the [license header check first](https://github.com/apple/swift-log/blob/main/.github/workflows/pull_request.yml#L13), it is expecting the `CONTRIBUTORS.txt` line in the license header and some repos still have it.

crleonard:
@kukushechkin  thanks, that makes sense.

Should I leave this PR open as a follow-up to #451, or would you prefer that I close it for now until the shared license header check supports headers without the `CONTRIBUTORS.txt` line?

Also, is `swiftlang/github-workflows` the right place to raise the license header issue?

kukushechkin:
@crleonard I would prefer to close it for now. When the shared license header check supports headers without the `CONTRIBUTORS.txt` line, we will update the repos.

Yes, `swiftlang/github-workflows` is the correct place to ask for the checks improvements.

PR #453, opened 2026-04-29
[SLG-0004]: metadata value attributes implementation

### LINKED ISSUE #204: Adding support for privacy level in string interpolation
On macOS, the (Apple’s) Logger object uses a custom `OSLogMessage` object which allows it to have a custom string interpolation with support for format arguments privacy levels.

Would it be possible to add support for said string interpolation so we could finally have an efficient and fully feature-compatible OSLog `Logger` for `swift-log`?

Thanks!

kukushechkin:
Proposal discussion thread: https://forums.swift.org/t/proposal-slg-0004-metadata-value-attributes-revision-2/86488

Add an extensible per-value attribute mechanism for metadata.

### Motivation:

Metadata values in `swift-log` are opaque strings by the time the `LogHandler` receives them. The call site often
knows things about a value that the handler cannot infer — for example, whether the value should be redacted in
different environments.

### Modifications:

- SLG-0004 proposal is updated based on the feedback. 
- Added `AttributedStringCarrier: CustomStringConvertible`-based mechanism to `Logging`.
- Added benchmarks and benchmark thresholds for the new attributed metadata.

### Result:

`swift-log` provides an optional extensible mechanism for attaching attributes to metadata values.


Sources/Logging/Logger.swift
```diff
@@ -222,34 +221,39 @@ extension Logger {
     ///    - line: The line this log message originates from. There's usually no need to pass it explicitly, as it
     ///            defaults to `#line`.
     @inlinable
-    public func log(
+    package func _log(
```
weissi: did the order this and the next function change? I think we should keep the old order to minimise the diff

kukushechkin: Fixed

kukushechkin:
Sensitivity attribute and `OSLogHandler` examples moved to https://github.com/kukushechkin/swift-log/tree/attributed-metadata-as-stringConvertible-with-examples

Sources/Logging/Logger+Attributes.swift
```diff
@@ -0,0 +1,441 @@
...
+//===----------------------------------------------------------------------===//
+
+// MARK: - MetadataValueAttributes storage
+
+extension Logger {
+    /// Attributes that can be associated with metadata values.
+    ///
+    /// `MetadataValueAttributes` stores one attribute inline without heap allocation. When more than one attribute
+    /// is needed, additional attributes spill over to a heap-allocated array.
+    /// Use the generic subscript to get/set attributes by their ``Attribute`` type.
```
FranzBusch: This doesn't belong in public docs since it is an implementation detail that can change at any point.

Sources/Logging/Logger+Attributes.swift
```diff
@@ -0,0 +1,441 @@
...
+    public struct MetadataValueAttributes: Sendable {
+        /// A protocol for defining custom metadata attributes.
+        ///
+        /// Conform to this protocol to define a custom attribute that can be stored in
+        /// ``MetadataValueAttributes``. Each conforming type acts as both the key (identified
+        /// by its metatype) and the value.
+        ///
+        /// This protocol is designed for **small, fixed-vocabulary attributes** represented as
+        /// enums. Each attribute value is stored as an `Int64` raw value, so attributes occupy minimal
```
FranzBusch: Nothing forces an attribute to be implemented with an enum. Anything that conforms to `RawRepresentable` can become an attribute. We should update the doc here.

Sources/Logging/Logger+Attributes.swift
```diff
@@ -0,0 +1,441 @@
...
+        /// space (one inline slot without heap allocation for the common single-attribute case).
```
FranzBusch, line 28: Same here. This is implementation detail

Sources/Logging/Logger+Attributes.swift
```diff
@@ -0,0 +1,441 @@
...
+        /// Create metadata value attributes using a builder closure.
+        ///
+        /// ```swift
+        /// let attrs = Logger.MetadataValueAttributes {
+        ///     $0[Sensitivity.self] = .sensitive
+        ///     $0[Priority.self] = .high
+        /// }
+        /// ```
+        @inlinable
+        public init(_ build: (inout Self) -> Void) {
```
FranzBusch: I don't think we should expose this as a public method

Sources/Logging/Logger+Attributes.swift
```diff
@@ -0,0 +1,441 @@
...
+extension Logger {
+    /// Internal carrier that wraps a string with metadata attributes.
+    ///
+    /// Produced by `MetadataValue.StringInterpolation` when any interpolation
+    /// segment specifies attributes, or by ``MetadataValue/attributed(_:attributes:)``.
+    /// This is the only type that carries attributes; the `.attributes` property on
+    /// `MetadataValue` checks for it with a concrete type comparison rather than a
+    /// protocol conformance lookup.
+    @usableFromInline
+    internal final class AttributedStringCarrier: CustomStringConvertible, Sendable {
```
FranzBusch: Does this need to be a class?

kukushechkin: Does not need to. My assumption was it should be faster for final class rather than struct-on-heap, but now I've crafted benchmarks and it seems structured will be more appropriate here — in case we miss `as?` (which will be most of the cases) class is much slower:

```
  │      Benchmark      │ Instructions │ CPU (ns) │ Mallocs  │
  ├─────────────────────┼──────────────┼──────────┼──────────┤
  │ class_as_cast_hit   │ 457          │ 22       │ 0        │
  ├─────────────────────┼──────────────┼──────────┼──────────┤
  │ struct_as_cast_hit  │ 551          │ 24       │ 0        │
  ├─────────────────────┼──────────────┼──────────┼──────────┤
  │ class_as_cast_miss  │ 1662         │ 70       │ 1 / 1000 │
  ├─────────────────────┼──────────────┼──────────┼──────────┤
  │ struct_as_cast_miss │ 620          │ 27       │ 0        │
```

Sources/Logging/Logger+Attributes.swift
kukushechkin: Removed completely, some remnants of API experimentations, array literal is much better.

Sources/Logging/Logger+Attributes.swift
kukushechkin: Removed.

Sources/Logging/Logger+Attributes.swift
kukushechkin: Added example with a `struct`.

Sources/Logging/Logger+Attributes.swift
kukushechkin: Removed.

Sources/Logging/Docs.docc/DisableLogLevelsDuringCompilation.md
```diff
@@ -73,3 +73,12 @@ logger.error("This still works")
 logger.critical("This still works")
 logger.log(level: .error, "This still works")
 ```
+
+When `MaxLogLevelNone` is enabled no log statement can ever fire, so
```
FranzBusch, line 77: This doesn't belong here right?

Sources/Logging/Logger+Attributes.swift
```diff
@@ -0,0 +1,433 @@
...
+        /// public struct Priority: Sendable, Hashable, Logger.MetadataValueAttributes.Attribute {
+        ///     public let rawValue: Int64
+        ///     public init(rawValue: Int64) {
+        ///         self.rawValue = rawValue
+        ///     }
+        ///     public static let low = Priority(rawValue: 1)
+        ///     public static let high = Priority(rawValue: 2)
+        /// }
+        /// ```
+
```
FranzBusch:
```suggestion
```

Tests/LoggingTests/MetadataAttributesTest.swift
```diff
@@ -0,0 +1,444 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Logging API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Testing
+
+@testable import Logging
```
FranzBusch: Do we need the `@testable` import?

FranzBusch (approved):
LGTM

Sources/Logging/Docs.docc/DisableLogLevelsDuringCompilation.md
kukushechkin: Merged from main.

Tests/LoggingTests/MetadataAttributesTest.swift
kukushechkin: Not really, removed.

PR #455, opened 2026-05-08
Fix Swift 6.2 build error with BUILD_LIBRARY_FOR_DISTRIBUTION=YES

JHeisecke:
Fix Swift 6.2 build error: remove @inlinable from Logger.Storage.init                                    

### Motivation:

Swift 6.2 enforces stricter definite initialization rules for `@inlinable` initializers under library       evolution mode. `Logger.Storage.init(label:handler:)` is marked `@inlinable` but assigns stored properties without using `self = .init(...)` syntax (the compiler now rejects this with a definite initialization error when building as a library with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES.`) 

This prevents downstream consumers from building swift-log as an XCFramework or any library evolution–enabled target.  

### Modifications:

Replaced `@inlinable` with `@usableFromInline` on `Logger.Storage.init(label:handler:)` in                  Sources/Logging/Logger.swift. The initializer remains accessible to inlinable callers (Logger.init is `@inlinable` and calls it), but is no longer itself required to be inlined across module boundaries.  

### Result:

`swift-log` compiles cleanly under Swift 6.2 with library evolution enabled. No behavior change, the stored properties are still initialized the same way; only the inlining guarantee for this one initializer is removed.


kukushechkin (commented):
Thank you @JHeisecke for the contribution. Could you clarify the use-case where you need to compile `swift-log` with the library evolution? Do you vendor `swift-log` to your framework users?

kukushechkin (approved):

PR #459, opened 2026-05-18
[SLG-0006]: task-local logger implementation

kukushechkin:
This proposal adds task-local logger storage so that metadata can accumulate across async
call stacks without threading a `Logger` through every function signature.

### Motivation:

Metadata propagation requires threading a logger through every layer. Every function on
the path from request ingress to the bottom of the call stack has to accept, mutate, and
forward a `Logger`. Libraries must choose between polluting their public API with a `logger:` parameter and
losing all the caller's context. There is no third option today.

### Modifications:

- New API surface for reading and modifying the `@TaskLocal` `Logger` instance added.
- Best practices on how to correctly propagate logger into libraries updated with `Logger.current` API.

### Result:

Users can now automatically propagate accumulated metadata into libraries without explicitly passing loggers through the API boundaries.


Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
```diff
@@ -0,0 +1,433 @@
...
+Metadata accumulates through nesting and task-local values propagate through structured
+concurrency (`async let`, `withTaskGroup`, child `Task {}`). `Task.detached` does not
+inherit context — capture the logger explicitly if needed.
+
+### Detailed design
+
+#### `Logger.current`
+
+Returns the logger bound by the nearest enclosing `withLogger` scope. If none has been set
+up, returns a fallback logger: the globally bootstrapped handler if
```
weissi: Hmm, we globally bootstrap _a factory_ to make a handler, not actually a handler.

And that's a problem for this. I don't think we should invoke the factory for every access even if you didn't `withLogger { ... }` anything up. Why isn't there a single, global fallback `Logger` that gets created lazily by invoking the factory once on first use?

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
```diff
@@ -0,0 +1,433 @@
...
+`LoggingSystem.bootstrap` has been called, otherwise a silent `SwiftLogNoOpLogHandler`.
```
weissi: I like making the default silent but in the original SwiftLog design we decided that that would be too unexpected for users. I think we should keep the original default: If you bootstrap you get that, otherwise you get the default no frills stdio logger.

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
```diff
@@ -0,0 +1,433 @@
...
+The fallback logger uses the empty label `""` — an empty-label line in production output
+is the deliberate diagnostic signal that no `withLogger` scope was set up before that
+emission, rather than a named-but-misleading placeholder.
+
+The no-op branch returns a cached logger and emits a **one-time warning on stderr** the
+first time it is taken so a user who to wrap their entry point in `withLogger`
+(or forgot to call `LoggingSystem.bootstrap`) doesn't see logs silently disappear with no diagnostic.
+
+The bootstrapped branch invokes `LoggingSystem.factory` on every access. `Logger.current` is
```
weissi: Why not a single, lazy global fallback `Logger` instead of calling the factory many times?

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
```diff
@@ -0,0 +1,433 @@
...
+    /// Task-local values propagate through structured concurrency (`async let`,
+    /// `withTaskGroup`, child `Task {}`) but are **not** inherited by `Task.detached`.
+    /// Capture the logger explicitly across detached boundaries.
+    ///
+    /// For many log calls in a tight scope, prefer extracting the logger once —
+    /// either `let logger = Logger.current` or the closure's `logger` parameter
+    /// from `withLogger { logger in ... }` — instead of re-reading the task-local
+    /// on every call.
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    public static var current: Logger { get }
```
weissi, line 123: I'm _somewhat_ sceptical of calling it just `current`. It's definitely convenient and reads nicely but it feels pretty magic. `Logger.taskLocal.debug("hello")` or something similar may be better because it explains nicely where it comes from.

In some systems, there will be multiple loggers to choose from and I think adding more clarity here would help.

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
```diff
@@ -0,0 +1,433 @@
...
+///     await handleRequests()  // reads Logger.current
+/// }
+///
+/// - Parameters:
+///   - logger: The logger to bind for the duration of `operation`.
+///   - operation: The closure to run with `logger` bound. Receives `logger` as a
+///     parameter so the body does not need to re-read ``Logger/current``.
+/// - Returns: The value returned by `operation`.
+@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public func withLogger<Result>(
```
weissi: Why not `Logger.withLogger(...) { ... }`? reads nicer than the free `withLogger`

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
kukushechkin: It suppose to rhyme with `Task` using `current`, something like "all my code is structure concurrency tasks, so I know which logger to us in my current task — the current logger".

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
kukushechkin: The idea was to match `withSpan`. Metrics now also have free `withMetricsFactory` following the same logic. Also would make one logger less over `Logger.withLogger(logger) { logger in ... }`.

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
kukushechkin: Hm, I think we can use the `Logger.init()` logic directly. And also push root task-local logger on the first `.current` usage — I've pushed the update, this simplifies the whole logic.

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
weissi: @kukushechkin I see you fixed this, right? I think we can resolve.

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
weissi: @kukushechkin I see you fixed this, right? I think we can resolve.

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
weissi: Yeah, I do get it. I like specificity but Swift APIs sometimes prefer looking nicer of being specific. Your call

Sources/Logging/Docs.docc/Proposals/SLG-0006-task-local-logger.md
weissi: Fair. I don't like it but you're right about the precedent, let's keep it or change it in all places at some other point in time

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,357 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Logging API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// MARK: - withLogger() free functions for task-local logger
+
+// Note: `rethrows` (not `throws(Failure)`) and no `Sendable` constraint on `Result` mirror
```
ktoso: This should already do typed throws; the reasons stdlib could not are tricky, but here you can just wrap the `throws(E)`  closure and cast it back on throwing `catch { throw error as! E }`

Sources/Logging/Logger.swift
```diff
@@ -1454,6 +1454,71 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
...
+    /// That means whichever backend is configured at first access becomes the unbound
+    /// default for the rest of the process — bootstrap before any logging for predictable
+    /// behavior.
+    @TaskLocal
+    static var taskLocalLogger: Logger = Logger(label: "")
+
+    /// Thin wrapper over `TaskLocal.withValue` so call sites don't need to reach into
+    /// the task-local storage directly.
+    #if compiler(>=6.2)
+    nonisolated(nonsending)
```
ktoso: very weird formatting?

Sources/Logging/Logger.swift
```diff
@@ -1454,6 +1454,71 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
...
+        static func withTaskLocalLogger<Return, Failure: Error>(
+            _ value: Logger,
+            operation: nonisolated(nonsending) () async throws(Failure) -> Return
+        ) async rethrows -> Return
```
ktoso:
```suggestion
        ) async throws(Failure) -> Return
```

No reason not to adopt typed throws here right away; do an `} catch { throws error as! Failure }` :)

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,357 @@
...
+///   - metadata: When non-nil, replaces the handler's base metadata dictionary for the
+///     scope. Pass `[:]` to erase inherited metadata; pass a fresh dictionary to start
+///     a scope from a known state. When `nil`, the inherited base metadata is preserved.
+///   - operation: The async closure to run with the modified logger bound. Receives the
+///     modified logger as a parameter.
+/// - Returns: The value returned by `operation`.
+#if compiler(>=6.2)
+@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+nonisolated(nonsending)
+    public func withLogger<Result>(
```
ktoso: Nitpick: While it's not a hard rule, since we're bad about documenting these......., in general please prefer `public nonisolated(nonsending)` order of params (Swift stdlib style, and we should update guidance about this in Swift style guide...)

kukushechkin: Changed the formatting

Sources/Logging/Logger.swift
kukushechkin: Responded on the forums. Updated to the typed throws.

Sources/Logging/Logger.swift
kukushechkin: Yeah, changed as well.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -31,70 +41,141 @@ struct RequestProcessor {
...
-        // Further propagate the logger to other services.
         let dbResult = try await databaseService.query(data.query, logger: logger)
-        
         logger.debug("Business logic completed")
         return HTTPResponse(data: dbResult)
     }
 }
 ```
 
-#### Alternative: Accept logger through initializer when appropriate
```
FranzBusch, line 61: Why did this become recommended?

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -31,70 +41,141 @@ struct RequestProcessor {
...
     private let logger: Logger
-    
+
     init(logger: Logger) {
         self.logger = logger
     }
-    
+
     func run() async {
-        // Execute some long running work
```
FranzBusch, line 73: Why did we remove the the below comment?

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -31,70 +41,141 @@ struct RequestProcessor {
...
         logger.debug("Update about long running work")
-        // Execute some more long running work
     }
 }
 ```
 
+#### Recommended: Read ``Logger/current`` from the task-local
+
+When a `logger:` parameter would clutter an otherwise logger-unrelated API, read the
```
FranzBusch: I would recommend not saying “clutter” but instead say that if an API doesn’t want to make Logging part of its public API I.e. to hide the dependency

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -31,70 +41,141 @@ struct RequestProcessor {
...
+    public func track(_ event: String) {
+        Logger.current.info("event", metadata: ["event.name": "\(event)"])
+    }
+}
+
+// Application binds at @main and scopes per-request metadata.
+@main
+struct MyServer {
+    static func main() async throws {
+        let logger = Logger(label: "my-server")
```
FranzBusch: This example is missing the bootstrap. I am wondering if we want a `bootstrap` method that also binds the take local

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -31,70 +41,141 @@ struct RequestProcessor {
...
+
+func handleRequest(_ req: HTTPRequest) async throws {
+    try await withLogger(mergingMetadata: ["request.id": "\(req.id)"]) { _ in
+        AnalyticsClient().track("request.received")    // sees request.id automatically
+    }
+}
+```
+
+For per-statement metadata, pass it via the `metadata:` parameter on the log call —
+it does not propagate and does not leak into other code:
```
FranzBusch: Remove the part about leak

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,383 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Logging API open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift Logging API project authors
```
FranzBusch: 2026

Sources/Logging/Logger.swift
```diff
@@ -1454,6 +1454,81 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
...
+    /// That means whichever backend is configured at first access becomes the unbound
+    /// default for the rest of the process — bootstrap before any logging for predictable
+    /// behavior.
+    @TaskLocal
+    static var taskLocalLogger: Logger = Logger(label: "")
+
+    /// Thin wrapper over `TaskLocal.withValue` so call sites don't need to reach into
+    /// the task-local storage directly.
+    #if compiler(>=6.2)
+    nonisolated(nonsending) static func withTaskLocalLogger<Return, Failure: Error>(
```
FranzBusch, line 1477: All of those methods should be inlinable

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,383 @@
...
+/// }
+/// ```
+///
+/// - Parameters:
+///   - logger: The logger to bind for the duration of `operation`.
+///   - operation: The closure to run with `logger` bound. Receives `logger` as a
+///     parameter so the body does not need to re-read ``Logger/current``.
+/// - Returns: The value returned by `operation`.
+@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+public func withLogger<Result, Failure: Error>(
```
FranzBusch, line 51: All methods in here should be inlinable

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: I think it will be the same technically, but confusing semantically — `LoggingSystem.bootstrap(logger) { }` vs `withLogger(logger) {}`. Can you call `bootstrap` only once? What is the difference with just `withLogger`?

Sources/Logging/Logger.swift
kukushechkin: Change to @inlinable.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Reworded as just whenever there is no logger in the API for whatever reasons.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Removed.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Shouldn't be recommended. The thinking was if you want a specific logger for the lifetime of an object, caching the logger might be a preferred way — the task-local one can be different over time.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: No reasons, restored.

Sources/Logging/Logger+With.swift
kukushechkin: 2026

Sources/Logging/Logger+With.swift
kukushechkin: Change to @inlinable.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
FranzBusch: I wouldn’t recommend that because caching the logger for the lifetime of an object means the metadata is not inherited down the call stack. I would consider storing the logger actually a bad practice now with task local propagation

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
FranzBusch: I am mostly thinking of how we recommend application developers to structure their application Main method. Now we recommend calling bootstrap and then withLogger right afterwards maybe providing a 2-1 solution will make it easier

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Yeah, let's keep it just "Alternative", not a recommended approach.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Right. Realistically, an application get completely get rid of `bootstrap` if it itself uses task-local or explicit loggers AND no dependencies create loggers (== do not rely on the global factory bootstrap). This is needed if application has to deal with loggers construction. [Changed](https://github.com/apple/swift-log/pull/459/changes/ebbc5253fe80812e5a5aa412f689ac90998be4c3) the paragraph in the corresponding section.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -1,20 +1,30 @@
...
+``Logger/current`` — never by constructing your own logger.
 
 ## Overview
 
-Libraries should accept logger instances through method parameters rather than
-storing them as instance variables. This practice ensures metadata (such as
-correlation IDs) is properly propagated down the call stack, while giving
-applications control over logging configuration.
+Libraries should obtain a `Logger` in one of two ways: accept one through a method or
+initializer parameter, or read ``Logger/current`` from the task-local. Both approaches
```
FranzBusch: We probably want to discourage the `initializer parameter` here since it loses the call stack metadata information

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
FranzBusch: We probably want to show at least that the explicit logger parameter defaults to `Logger.current`

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -31,44 +41,40 @@ struct RequestProcessor {
...
         let dbResult = try await databaseService.query(data.query, logger: logger)
-        
         logger.debug("Business logic completed")
         return HTTPResponse(data: dbResult)
     }
 }
 ```
 
-#### Alternative: Accept logger through initializer when appropriate
+#### Alternative: Accept logger through initializer for long-lived components
```
FranzBusch, line 116: We should move this below the other recommended section and change the checkmark to a warning sign

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -77,24 +83,103 @@ final class BackgroundJobProcessor {
...
+
+```swift
+// ✅ Good: capture before the boundary, rebind on the other side.
+try await withLogger(mergingMetadata: ["request.id": "r1"]) { _ in
+    let captured = Logger.current
+    URLSession.shared.dataTask(with: req) { data, _, _ in
+        withLogger(captured) { logger in
+            logger.info("response")    // request.id preserved
+        }
+    }.resume()
```
FranzBusch, line 183: These examples read weird. The parameter `mergingMetadata` is never used. Furthermore, I don't see why you would use `withLogger` in the good example

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,392 @@
...
+    _ logger: Logger,
+    _ operation: (Logger) throws(Failure) -> Result
+) throws(Failure) -> Result {
+    do {
+        return try Logger.withTaskLocalLogger(logger) {
+            try operation(logger)
+        }
+    } catch {
+        throw error as! Failure
```
FranzBusch, line 60: Let's create an issue once this is merged to track adopting the typed throws task local method

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,392 @@
...
+            try await operation(logger)
+        }
+    } catch {
+        throw error as! Failure
+    }
+}
+#else
+@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+@inlinable
+public func withLogger<Result, Failure: Error>(
```
FranzBusch, line 114: All these 6.2 below methods should have an `isolation: isolated (any Actor?) = #isolation` parameter which is equivalent to the `nonisolated(nonsending)` spelling in 6.2+

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,392 @@
...
+/// Unlike ``withLogger(mergingMetadata:_:)-(_,(Logger)(Failure)->Result)``, which layers
+/// metadata on top of the
+/// inherited base, this overload **replaces** the base metadata when `metadata` is
+/// non-nil. Pass `metadata: [:]` to wipe the inherited metadata entirely for the scope.
+///
+/// With no arguments, this overload re-binds ``Logger/current`` unchanged — a convenient
+/// way to extract it into a local variable for repeated use inside the closure.
+///
+/// ```swift
+/// withLogger(mergingMetadata: ["request.id": "\(request.id)"]) { _ in
```
FranzBusch, line 251: This example is wrong for this method

Sources/Logging/Logger+With.swift
```diff
@@ -0,0 +1,392 @@
...
```
FranzBusch, line 315: Same here this is using the wrong method in the example

Sources/Logging/Logger.swift
```diff
@@ -1454,6 +1454,84 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
...
+    /// That means whichever backend is configured at first access becomes the unbound
+    /// default for the rest of the process — bootstrap before any logging for predictable
+    /// behavior.
+    @TaskLocal
+    static var taskLocalLogger: Logger = Logger(label: "")
+
+    /// Thin wrapper over `TaskLocal.withValue` so call sites don't need to reach into
+    /// the task-local storage directly.
+    #if compiler(>=6.2)
+    @usableFromInline
```
FranzBusch: Can this be made `@inlinable` and the `taskLocalLogger` made `@usableFromInline`?

Sources/Logging/Logger.swift
```diff
@@ -1454,6 +1454,84 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
...
+        operation: nonisolated(nonsending) () async throws(Failure) -> Return
+    ) async throws(Failure) -> Return {
+        do {
+            return try await Self.$taskLocalLogger.withValue(value, operation: operation)
+        } catch {
+            throw error as! Failure
+        }
+    }
+    #else
+    @usableFromInline
```
FranzBusch: Same here `@inlinable`

Sources/Logging/Logger.swift
```diff
@@ -1454,6 +1454,84 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
...
+    ) async throws(Failure) -> Return {
+        do {
+            return try await Self.$taskLocalLogger.withValue(value, operation: operation)
+        } catch {
+            throw error as! Failure
+        }
+    }
+    #endif
+
+    @usableFromInline
```
FranzBusch: and here

Sources/Logging/Logger.swift
```diff
@@ -1454,6 +1454,84 @@ extension Logger.MetadataValue: ExpressibleByArrayLiteral {
...
+    ///
+    /// Because the unbound default is captured at first access, ``LoggingSystem/bootstrap(_:)``
+    /// must be called before any task-local logger API is exercised; otherwise the
+    /// unbound default uses the swift-log default ``StreamLogHandler`` and a later
+    /// bootstrap is not visible to ``current`` outside of a `withLogger` scope.
+    ///
+    /// Task-local values propagate through structured concurrency (`async let`,
+    /// `withTaskGroup`, child `Task {}`) but are **not** inherited by `Task.detached`.
+    /// Capture the logger explicitly across detached boundaries.
+    public static var current: Logger {
```
FranzBusch, line 1532: This should also be `@inlinable`

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: It does not necessary loose it, the decision is up to the caller to pass `Logger.current`. Will work just fine for a per-operation logger. I expect some users to prefer the explicit parameter, especially if it is already there.

Sources/Logging/Logger+With.swift
kukushechkin: The actual example is in the nested `withLogger(metadata:)` showing wiping metadata if it is not desired to be passed down the stack.

Sources/Logging/Logger+With.swift
kukushechkin: Same — the example is in the nested `withLogger`.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
FranzBusch: I think there is a difference here the explicit per method parameter is fine but an initializer parameter is in almost all use-cases wrong. It won't properly propagate the metadata unless they themselves extract the metadata from `Logger.current` again which at that point they should just use that logger. So in particular I just want to remove the `or initializer` from this sentence

Sources/Logging/Logger.swift
kukushechkin: Requires some juggling. `@TaskLocal` cannot be `@usableFromInline` (@ktoso is aware), so either use a `TaskLocal<>` or have a wrapper.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Default `Logger.current` value is a great idea. I've also reframed this as a "good alternative". Updated.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Updated.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: `mergingMetadata` param is not used in the first example — this is the anti-pattern highlighted. `withLogger` binds the task-local scope inside the callback so anything downstream can use `Logger.current`. I've updated a bit the examples.

Sources/Logging/Logger+With.swift
kukushechkin: Updated.

Sources/Logging/Logger.swift
kukushechkin: Do you think we shouldn't worry about evolution of .current and ABI?

Sources/Logging/Logger.swift
FranzBusch: Can we use `TaskLocal<>` for now to get around this? It is only internal anyways. Also is there a GH issue for the macro problem?

Sources/Logging/Logger.swift
FranzBusch: `swift-log` is not ABI stable so we shouldn't worry about it

Sources/Logging/Logger.swift
kukushechkin: Updated to `TaskLocal<>`.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
kukushechkin: Removed.

Sources/Logging/Docs.docc/BestPractices/003-AcceptingLoggers.md
```diff
@@ -1,74 +1,129 @@
...
+the application has set up, rather than constructing their own from scratch. This
+ensures correlation IDs and other contextual metadata flow through the entire call
+stack, and gives the application a single place to redirect or filter all log output.
+
+Two propagation mechanisms are available, and they coexist. Choose the explicit
+parameter when the library's API already accepts a `Logger` (or it's natural to add
+one) — the call site stays declarative about what gets logged where. Choose the
+task-local when adding a `logger:` parameter would pollute an API that otherwise has
+no logging concern in its signature. Application code drives the task-local binding;
+library code reads it.
```
FranzBusch:
```suggestion
no logging concern in its signature. Application code binds the initial task-local;
library code reads it and updates the metadata.
```

FranzBusch (approved):
Only a small doc nits otherwise LGTM

PR #464, opened 2026-05-19
Print swift toolchains in the benchmarks

kukushechkin:
Print swift toolchains in the benchmarks.


kukushechkin:
I don't see any anomalies with the Xcode versions.

PR #466, opened 2026-05-20
Adopt multi-package macOS benchmarks workflow

rnro:
### Motivation

* The two-invocation pattern for `macos-benchmarks` and `macos-benchmarks-MaxLogLevelWarningBenchmarks-trait` allocates one self-hosted macOS runner per (Swift version x package), incurring duplicate runner-recycling cost.
* The recent change on main adds a `benchmark_package_paths` input letting a single invocation run all packages serially on each runner.

### Modifications

* Replace the two macOS benchmark jobs in `main.yml` and `pull_request.yml` with a single consolidated job that passes both packages via `benchmark_package_paths: '["Benchmarks/NoTraits", "Benchmarks/MaxLogLevelWarning"]'`.
* Drop Swift 6.2 from the matrix; only Swift 6.3 is shared between the two packages (`MaxLogLevelWarning` has no `Thresholds/Xcode swift 6.2/`).

### Result

* One macOS runner allocation per Swift version instead of one per (version x package).
* `NoTraits` loses Swift 6.2 coverage

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,27 +31,10 @@ jobs:
       windows_nightly_main_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
 
   macos-benchmarks:
-    name: macOS benchmarks (NoTraits)
+    name: macOS benchmarks
     uses: apple/swift-nio/.github/workflows/macos_benchmarks.yml@main
     with:
-      benchmark_package_path: "Benchmarks/NoTraits"
-      macos_swift_6_2_enabled: true
-      macos_swift_6_3_enabled: true
-
-  macos-benchmarks-MaxLogLevelWarningBenchmarks-trait:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -31,27 +31,10 @@ jobs:
       windows_nightly_main_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
 
   macos-benchmarks:
-    name: macOS benchmarks (NoTraits)
+    name: macOS benchmarks
     uses: apple/swift-nio/.github/workflows/macos_benchmarks.yml@main
     with:
-      benchmark_package_path: "Benchmarks/NoTraits"
-      macos_swift_6_2_enabled: true
```
kukushechkin, line 38: `NoTraits/Thresholds/Xcode swift 6.2` thresholds can be removed now.

kukushechkin (approved):

PR #431, opened 2026-03-24
[WIP] Test Darwin tests

czechboy0:
### Motivation:

Iterating on PR CI here until we get it stabilized.

### Modifications:

N/A

### Result:

N/A

PR #473, opened 2026-06-24
Enable Xcode tests

kukushechkin:
Simulators seem to have stabilized in CI, we can enabled the tests.

### Motivation:

`swift-log` is expected to be used on various platforms, we should run tests against them.

### Modifications:

- Run tests on various tests in simulators in PRs and nightly.
- Fix tests to actually compile there.

### Result:

Tests are now running in simulators. Supersedes #431.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,12 +52,19 @@ jobs:
     with:
       runner_pool: nightly
       build_scheme: swift-log-Package
-      debug_output_enabled: true
-      macos_xcode_test_enabled: false
-      ios_xcode_test_enabled: false
-      watchos_xcode_test_enabled: false
-      tvos_xcode_test_enabled: false
-      visionos_xcode_test_enabled: false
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -52,12 +52,19 @@ jobs:
     with:
       runner_pool: nightly
       build_scheme: swift-log-Package
-      debug_output_enabled: true
-      macos_xcode_test_enabled: false
-      ios_xcode_test_enabled: false
-      watchos_xcode_test_enabled: false
-      tvos_xcode_test_enabled: false
-      visionos_xcode_test_enabled: false
+      xcode_16_3_enabled: false
```
FranzBusch: I think the new inputs are not called Xcode anymore but rather swift based names

kukushechkin: Yes, right. Removed.

FranzBusch (approved):

PR #474, opened 2026-06-24
Add missing API references

kukushechkin:
Add missing API refs for metadata attributes and task-local logger API.

### Motivation:

API got a few new features, and they are missing from the API references.

### Modifications:

- Task-local logger API added to the references.
- Metadata attributes API added to the references.

### Result:

Docs have all the recently added APIs.

PR #475, opened 2026-06-24
Add testing logs best practice

kukushechkin:
Add a Best Practice guide on how to test logging.

### Motivation:

There are 2 recommended ways of propagating logs in components, both are testable. But there is no any kind of documentation on how to do it.

### Modifications:

Add a "Testing code that logs" Best Practice guide.

### Result:

Logs testing is now documented.

### DIFF
--- a/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
+++ b/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
@@ -19,12 +19,19 @@ Add the product to your test target:
 
 When the code under test accepts a `Logger` (see <doc:003-AcceptingLoggers>), back it with
 an `InMemoryLogHandler` and validate the results on `entries`.
+
 The following code illustrates how to set up and validate a handler using `InMemoryLogging` to check the expected results:
 
 ```swift
 import InMemoryLogging
 import Logging
 import Testing
+
+struct RequestProcessor {
+    func handle(id: String, logger: Logger) {
+        logger.info("request received", metadata: ["request.id": "\(id)"])
+    }
+}
 
 @Test
 func stampsRequestID() {
@@ -46,12 +53,20 @@ capture lower levels.
 ### Capture logs from code that reads `Logger.current`
 
 When the code you want to test reads the task-local ``Logger/current`` instead of taking a parameter, bind an `InMemoryLogHandler` for the scope of that by using ``withLogger(logLevel:handler:metadata:_:)``.
-The handler shares its storage by reference, so the copy you hold sees what was logged:
+The handler shares its storage by reference, so the copy you hold sees what was logged.
+
+Bind the handler for the scope of the call and validate the captured entry:
 
 ```swift
 import InMemoryLogging
 import Logging
 import Testing
+
+struct AnalyticsClient {
+    func track(_ event: String) {
+        Logger.current.info("event", metadata: ["event.name": "\(event)"])
+    }
+}
 
 @Test
 func tracksEvent() {
--- a/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
+++ b/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
@@ -42,7 +42,6 @@ func stampsRequestID() {
 Each `Entry` is equatable and exposes `level`, `message`, `metadata`, and `error`.
 The handler log level defaults to ``Logger/Level/info``; set `handler.logLevel = .trace` before you use it to
 capture lower levels.
-Don't use `withLogger(logLevel:)` to adjust the log level, as the log level is overwritten when you also pass `handler`).
 
 ### Capture logs from code that reads `Logger.current`
 
--- a/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
+++ b/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
@@ -52,7 +52,7 @@ capture lower levels.
 
 ### Capture logs from code that reads `Logger.current`
 
-When the code you want to test reads the task-local ``Logger/current`` instead of taking a parameter, bind an `InMemoryLogHandler` for the scope of that by using ``withLogger(logLevel:handler:metadata:_:)``.
+When the code you want to test reads the task-local ``Logger/current`` instead of taking a parameter, bind an `InMemoryLogHandler` for the scope of that by using ``withLogger(logLevel:handler:metadata:_:)-2pd9p``.
 The handler shares its storage by reference, so the copy you hold sees what was logged.
 
 Bind the handler for the scope of the call and validate the captured entry:
--- a/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
+++ b/Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
@@ -1,0 +1,69 @@
+# 004: Testing code that logs
+
+Assert on what your code logged by capturing entries in memory, instead of bootstrapping
+a process-wide backend.
+
+## Overview
+
+The `InMemoryLogging` product ships an `InMemoryLogHandler` that collects every emitted
+record into an array you can inspect with `#expect`. Keep it local to each test: a handler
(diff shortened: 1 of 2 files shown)


kukushechkin:
Depends on #474

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,69 @@
+# 004: Testing code that logs
+
+Assert on what your code logged by capturing entries in memory, instead of bootstrapping
+a process-wide backend.
```
heckj:
```suggestion
Validate what you code logs by capturing entries in memory rather than bootstrapping a process-wide backend.
```

Says the same thing, but I think it's a little more direct and easier to parse. WDYT?

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Overview
+
+The `InMemoryLogging` product ships an `InMemoryLogHandler` that collects every emitted
+record into an array you can inspect with `#expect`. Keep it local to each test: a handler
+created in the test has no shared state, so tests stay isolated and parallel-safe — unlike
+``LoggingSystem/bootstrap(_:)``, which sets one process-wide backend and traps if called
+twice.
```
heckj:
```suggestion
Import and use `InMemoryLogging`, which provides an `InMemoryLogHandler` that collects every emitted record into an array you can inspect.
Keep a handler you create local to each test.
Each handler that you create in a test has no shared state, so the test stays isolated is safe to run on parallel with other tests.
Don't use ``LoggingSystem/bootstrap(_:)`` in a test, it sets one process-wide backend and traps if called twice.
```

Rewording suggestions to make it more direct, and break the up sentences for a bit more clarity.

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,69 @@
...
+Add the product to your test target:
+
+```swift
+.product(name: "InMemoryLogging", package: "swift-log")
+```
+
+### Capture logs from a logger you pass in
+
+When the code under test accepts a `Logger` (see <doc:003-AcceptingLoggers>), back it with
+an `InMemoryLogHandler` and assert on `entries`:
```
heckj:
```suggestion
an `InMemoryLogHandler` and validate the results on `entries`.
The following code illustrates how to set up and validate a handler using `InMemoryLogging` to check the expected results:
```

I think I'm avoiding "assert" here because I of overlap with the assert() and precondition() language options, which this isn't using, so I'm leaning into "validate" as the English-generic version of using #expect from swift-testing.

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,69 @@
...
+    #expect(handler.entries.count == 1)
+    #expect(handler.entries[0].message == "request received")
+    #expect(handler.entries[0].metadata["request.id"] == "42")
+}
+```
+
+Each `Entry` exposes `level`, `message`, `metadata`, and `error`, and is `Equatable`. The
+handler defaults to ``Logger/Level/info``; set `handler.logLevel = .trace` before use to
+capture lower levels (not via `withLogger(logLevel:)`, which is overwritten when you also
+pass `handler`).
```
heckj:
```suggestion
Each `Entry` is equatable and exposes `level`, `message`, `metadata`, and `error`.
The handler log level defaults to ``Logger/Level/info``; set `handler.logLevel = .trace` before you use it to
capture lower levels.
Don't use `withLogger(logLevel:)` to adjust the log level, as the log level is overwritten when you also pass `handler`).
```

The earlier phrasing made it sounds like all of Entry, level, message, metadata, and error were equatable. I *think* you mean that Entry is explicitly equatable, so I moved it forward in the sentence to avoid making it sound like `error` might be the thing you're asserting is equatable.

I tried to break out the `withLogger(logLevel:)`, but I'm not sure I caught what you were saying correctly. Is that correct?

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Capture logs from code that reads `Logger.current`
+
+When the code reads the task-local ``Logger/current`` instead of taking a parameter, bind
+an `InMemoryLogHandler` for the scope with ``withLogger(logLevel:handler:metadata:_:)``. The
+handler shares its storage by reference, so the copy you hold sees what was logged:
```
heckj:
```suggestion
When the code you want to test reads the task-local ``Logger/current`` instead of taking a parameter, bind an `InMemoryLogHandler` for the scope of that by using ``withLogger(logLevel:handler:metadata:_:)``.
The handler shares its storage by reference, so the copy you hold sees what was logged:
```

Added just a smidge of additional wording to make it clearer advice, assuming you don't want to change the upstream code you're testing.

heckj (commented):
tweaks to language to try and break up the sentences a little bit into more easily consumable and direct pieces. See what you think, and please triple-check that I didn't infer incorrectly and say something incorrect in the rephrasing suggestions.

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
kukushechkin: I think this `withLogger(logLevel:)` part is an artifact, I can't recollect what was the idea here. I would remove it completely.

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,67 @@
...
+    #expect(handler.entries.count == 1)
+    #expect(handler.entries[0].message == "request received")
+    #expect(handler.entries[0].metadata["request.id"] == "42")
+}
+```
+
+Each `Entry` is equatable and exposes `level`, `message`, `metadata`, and `error`.
+The handler log level defaults to ``Logger/Level/info``; set `handler.logLevel = .trace` before you use it to
+capture lower levels.
+Don't use `withLogger(logLevel:)` to adjust the log level, as the log level is overwritten when you also pass `handler`).
```
kukushechkin:
```suggestion
```

heckj (approved):

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,66 @@
...
+import InMemoryLogging
+import Logging
+import Testing
+
+@Test
+func stampsRequestID() {
+    let handler = InMemoryLogHandler()
+    let logger = Logger(label: "test", factory: { _ in handler })
+
+    RequestProcessor().handle(id: "42", logger: logger)
```
FranzBusch, line 41: Could we define this method above to show the users exactly what it does?

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
```diff
@@ -0,0 +1,66 @@
...
+```swift
+import InMemoryLogging
+import Logging
+import Testing
+
+@Test
+func tracksEvent() {
+    let handler = InMemoryLogHandler()
+    withLogger(handler: handler) { _ in
+        AnalyticsClient().track("signup")  // calls Logger.current.info(...)
```
FranzBusch, line 75: Same here

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
kukushechkin: Yeah, makes sense, added.

Sources/Logging/Docs.docc/BestPractices/004-TestingLogging.md
kukushechkin: Added.

FranzBusch (approved):

PR #477, opened 2026-06-24
Make `StreamLogHandler` initializers public

### LINKED ISSUE #273: Make StreamLogHandler initializers public
### Feature Requests
I'd like to reuse the logic for formatting the log message and prettify the metadata from StreamLogHandler, but using my own `TextOutputStream` implementation.

We'd need to make the [StreamLogHandler initializers](https://github.com/apple/swift-log/blob/main/Sources/Logging/Logging.swift#L1318-L1328) public to pass a different stream.

subhramit:
Make `StreamLogHandler` initializers public.

### Motivation:

Closes https://github.com/apple/swift-log/issues/273

### Modifications:

Promote the two stream-accepting initializers from `internal` to `public`. Also as a result, their stream parameter is spelled out as `any TextOutputStream & Sendable` rather than the internal `_SendableTextOutputStream` typealias, which cannot appear in a public signature. ~The private stored property and the internal typealias are otherwise unchanged.~ The internal typealias has been completely removed as it was only used to store the stream, and the private stored property is updated to use the spelled-out type.

Also updates the docc documentation accordingly.

### Result:

Users would be able to construct a `StreamLogHandler` with their own `TextOutputStream` implementation, reusing the existing logic to format log messages and prettify metadata instead of reimplementing them.

### DIFF
--- a/Sources/Logging/Handlers/StreamLogHandler.swift
+++ b/Sources/Logging/Handlers/StreamLogHandler.swift
@@ -38,8 +38,6 @@ import WASILibc
 /// 2. The handler's ``metadataProvider`` is invoked, overriding any existing keys.
 /// 3. The per-log-statement metadata is merged, overriding any previously set keys.
 public struct StreamLogHandler: LogHandler {
-    internal typealias _SendableTextOutputStream = TextOutputStream & Sendable
-
     /// Creates a stream log handler that directs its output to STDOUT.
     public static func standardOutput(label: String) -> StreamLogHandler {
         StreamLogHandler(
@@ -68,7 +66,7 @@ public struct StreamLogHandler: LogHandler {
         StreamLogHandler(label: label, stream: StdioOutputStream.stderr, metadataProvider: metadataProvider)
     }
 
-    private let stream: any _SendableTextOutputStream
+    private let stream: any TextOutputStream & Sendable
     private let label: String
 
     /// Get the log level configured for this `Logger`.
--- a/Sources/Logging/Handlers/StreamLogHandler.swift
+++ b/Sources/Logging/Handlers/StreamLogHandler.swift
@@ -102,13 +102,24 @@ public struct StreamLogHandler: LogHandler {
         }
     }
 
-    // internal for testing only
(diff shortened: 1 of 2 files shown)


subhramit:
The failing CI checks seem to be a toolchain/package-config mismatch and unrelated to my changes.

Sources/Logging/Handlers/StreamLogHandler.swift
```diff
@@ -102,13 +102,24 @@ public struct StreamLogHandler: LogHandler {
...
 
-    // internal for testing only
-    internal init(label: String, stream: any _SendableTextOutputStream) {
+    /// Create a ``StreamLogHandler`` that directs its output to the stream you provide,
+    /// using the global metadata provider from ``LoggingSystem``.
+    ///
+    /// - parameters:
+    ///   - label: The label for this log handler.
+    ///   - stream: The stream to write log output to.
+    public init(label: String, stream: any TextOutputStream & Sendable) {
```
kukushechkin, line 109: As this is a new public API, could you also update the docs https://github.com/apple/swift-log/blob/main/Sources/Logging/Docs.docc/Reference/StreamLogHandler.md?

subhramit: Sure, done.

Sources/Logging/Handlers/StreamLogHandler.swift
```diff
@@ -102,13 +102,24 @@ public struct StreamLogHandler: LogHandler {
         }
     }
 
-    // internal for testing only
-    internal init(label: String, stream: any _SendableTextOutputStream) {
```
kukushechkin, line 106: I think we can remove `typealias _SendableTextOutputStream`, it is only used to store the stream.

kukushechkin (changes requested):

Sources/Logging/Handlers/StreamLogHandler.swift
subhramit: Makes sense, removed.

kukushechkin (approved):
@subhramit Thank you!

PR #483, opened 2026-07-03
Add millisecond precision to StreamLogHandler timestamp

### LINKED ISSUE #212: a hardcoded timestamp does not contain mili- or nanoseconds in any format
### Expected behavior
Timestamp should have nanoseconds by default or it should be possible to customise timestamp.

### Actual behavior
 There are only seconds on a time value. For example there is a log line:
```
2021-11-06T18:47:14+0300 info
```

### Steps to reproduce
1. Just use `StreamLogHandler.standardOutput` at `LoggingSystem.bootstrap`

### There is minimal reproducer code
```
        LoggingSystem.bootstrap{ label in
            var logHandler = StreamLogHandler.standardOutput(label: " org.example.myproject." + label)
            logHandler.logLevel = Logger.Level.debug
            return logHandler
        }
```

### SwiftLog version (tag)

_1.4.2_
This issue also affects all versions before it.

### Swift & OS version
Swift:
```
vital@Vitals-MacBook-Air ~ % swift --version && uname -a
swift-driver version: 1.26.9 Apple Swift version 5.5.1 (swiftlang-1300.0.31.4 clang-1300.0.29.6)
Target: arm64-apple-macosx11.0
Darwin Vitals-MacBook-Air.local 20.6.0 Darwin Kernel Version 20.6.0: Tue Oct 12 18:33:38 PDT 2021; root:xnu-7195.141.8~1/RELEASE_ARM64_T8101 arm64
vital@Vitals-MacBook-Air ~ %
```
Docker container (Linux Ubuntu)
```
root@1765fec3b27c:/usr/src# swift --version && uname -a
Swift version 5.3 (swift-5.3-RELEASE)
Target: x86_64-unknown-linux-gnu
Linux 1765fec3b27c 5.10.47-linuxkit #1 SMP PREEMPT Sat Jul 3 21:50:16 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
root@1765fec3b27c:/usr/src
```

### Code line contains a bug
There is a place where timestamp is hardcoded:
[source](https://github.com/apple/swift-log/blob/1.4.2/Sources/Logging/Logging.swift#L933)

krishnapermi:
The `timestamp()` function in `StreamLogHandler` uses `time(nil)` with `strftime("%Y-%m-%dT%H:%M:%S%z")`, which only provides second-level precision. When multiple log events occur within the same second they share an identical timestamp, making it impossible to determine their ordering from the log output.

**Fix:** On non-Windows platforms, replace `time(nil)` with `clock_gettime(CLOCK_REALTIME, &ts)` to obtain nanosecond-resolution wall-clock time, then inject milliseconds between the seconds and timezone fields. Example output:

```
2021-11-06T18:47:14.123+0300 info myapp: message
```

I noticed this while looking at swift-log's `StreamLogHandler` implementation - the hardcoded format string never included sub-second digits, and on busy systems many consecutive log lines ended up with the same timestamp.

The Windows path is unchanged; adding sub-second support there needs different APIs and is a separate concern.

Fixes #212


kukushechkin (changes requested):
Thank you @krishnapermi for looking into this! My concern is — not making the change across all the supported platforms is breaking user expectations for a cross-platform package. Could you please look into the windows implementation? In addition, this functionality is testable, so I would recommend adding tests.

krishnapermi:
Good catches - thanks for the thorough review.

I've pushed two more commits to address both points:

**Windows support** (`0c8dc77`): The Windows branch now uses `_ftime64_s` with `__timeb64` to get millisecond-precision wall-clock time, the same way the existing code already uses `_localtime64_s` for local time conversion. I refactored the formatting code (buffer-to-string conversion, zero-padding, the final return) to sit after the `#endif` so it's shared between platforms rather than duplicated.

**Tests** (`24686d1`): Added `StreamLogHandlerTimestampTests.swift` in `Tests/LoggingTests/`. The tests create a `StreamLogHandler` backed by a simple in-memory `CapturingStream`, emit log lines, and verify that the timestamp field contains a 3-digit millisecond component in the right position and within `[0, 999]`. There's also a structural check confirming the overall ISO 8601 shape.

Sources/Logging/Handlers/StreamLogHandler.swift
```diff
@@ -38,401 +38,383 @@ import WASILibc
...
-        flush: @escaping (F) -> Int32
-    ) {
-        self.flushMode = flushMode
-        self._writeBytes = { bytes in
-            lock?(file)
-            defer { unlock?(file) }
-            if let base = bytes.baseAddress, bytes.count > 0 {
-                _ = write(base, 1, bytes.count, file)
-            }
+        internal let flushMode: FlushMode
```
kukushechkin: I think something went terribly wrong with the formatting, please, fix it. You can run `swift-format -rip .` locally to make sure formatting is correct.

kukushechkin (changes requested):

Sources/Logging/Handlers/StreamLogHandler.swift
krishnapermi: Good catch - you're right, the GitHub web editor corrupted the indentation across the whole file when I saved. I've pushed two commits to fix it: 4e52d78 restores correct 4-space indentation in `StreamLogHandler.swift` and 112c9ca does the same for the test file.

kukushechkin: It seems not all format issues have been resolved: https://github.com/apple/swift-log/actions/runs/28844226495/job/85594476307?pr=483

kukushechkin (changes requested):
It seems Wasm tests are failing for a legit reason, `CLOCK_REALTIME` is not available. There are 2 jobs:
- https://github.com/apple/swift-log/actions/runs/29158273220/job/86911285818?pr=483
- https://github.com/apple/swift-log/actions/runs/29158273220/job/86911285870?pr=483

Could you please follow the same pattern as in other places, leveraging ifdef to provide implementation compatible with Wasm?

krishnapermi:
Fixed in the latest commit by adding a `#elseif canImport(WASILibc)` branch before `#else`, following the same pattern used elsewhere in the codebase (e.g. the `StdioOutputStream` stdin/stdout/stderr setup).

WASI defines `CLOCK_REALTIME` as a pointer-to-struct macro, which Swift cannot import. The Wasm branch falls back to `time(nil)`, giving second-level precision (same as the original implementation had before this PR). The `ms` field is set to `0` so the timestamp format stays consistent across all platforms.

Sources/Logging/Handlers/StreamLogHandler.swift
```diff
@@ -63,7 +63,7 @@ public struct StreamLogHandler: LogHandler {
 
     /// Creates a stream log handler that directs its output to STDERR using the metadata provider you provide.
     public static func standardError(label: String, metadataProvider: Logger.MetadataProvider?) -> StreamLogHandler {
-        StreamLogHandler(label: label, stream: StdioOutputStream.stderr, metadataProvider: metadataProvider)
+        StreamLogHandler(label: label, stream: StdioOutputStream.st$err, metadataProvider: metadataProvider)
```
kukushechkin:
```suggestion
        StreamLogHandler(label: label, stream: StdioOutputStream.stderr, metadataProvider: metadataProvider)
```

kukushechkin (changes requested):

FranzBusch:
@weissi WDYT? Do you consider this a semantically breaking change?

PR #485, opened 2026-07-12
Add CMake support for Swift targets and test configurations

### LINKED ISSUE #484: Support building with CMake
Support CMake build compatible with [SwiftPM](https://github.com/swiftlang/swift-package-manager/blob/main/CMakeLists.txt#L43) similar to [swift-certificates](https://github.com/apple/swift-certificates/blob/main/CMakeLists.txt), [swift-argument-parser](https://github.com/apple/swift-argument-parser/blob/main/CMakeLists.txt), or other SPM dependencies of SwiftPM.

mitras-git:
Add CMake build support to the project to enable toolchain integration and bootstrapping compatibility.

### Motivation:

Currently, swift-log does not support building with CMake. This addition resolves issue #484, allowing swift-log to be resolved and compiled within CMake-based environments (like the repos swift-certificates and swift-argument-parser which aer mentioned in the issue).

### Modifications:

1. **Project Root Config:** Added [CMakeLists.txt](file:///Users/adityamitra/Developer/swift-log/CMakeLists.txt) configuring the CMake project with Swift 6 and C, alongside setting up compiler flags for MSVC runtime library on Windows.
2. **Helper Modules:** Added the `cmake/` folder:
   - `SwiftSupport.cmake`: dylib/module installations and triple detection helper.
   - `SwiftLogConfig.cmake.in`: package config export templates.
3. **Target CMake Configurations:** Created [Sources/Logging/CMakeLists.txt](file:///Users/adityamitra/Developer/swift-log/Sources/Logging/CMakeLists.txt) & [Sources/InMemoryLogging/CMakeLists.txt](file:///Users/adityamitra/Developer/swift-log/Sources/InMemoryLogging/CMakeLists.txt) to define the library targets. Included generator expressions pointing `INTERFACE_INCLUDE_DIRECTORIES` to target output directories for Xcode generator support.
4. **SwiftPM Exclusions:** Excluded the new CMake files from target targets inside [Package.swift](file:///Users/adityamitra/Developer/swift-log/Package.swift) and [Package@swift-6.1.swift](file:///Users/adityamitra/Developer/swift-log/Package@swift-6.1.swift) to keep SwiftPM targets warning-free.
5. **Compile Options Variable:** Consolidates compile options under a single global definition `SWIFT_LOG_COMMON_COMPILE_OPTIONS` in the root configuration, enabling Swift 6 compiler features (`ExistentialAny`, `MemberImportVisibility`, `InternalImportsByDefault`, `NonisolatedNonsendingByDefault`, and explicit Sendable checks) across all targets.
6. **CI Integration & Workflow:** Integrated CMake CI validation in `.github/workflows/pull_request.yml` using the shared workflow from `apple/swift-nio/.github/workflows/cmake_tests.yml@main` running in a `swift:6.1-jammy` container. This job compiles the core libraries and verifies the validity of the CMake files.
7. **License Header Compliance:** Cleaned up license headers across all CMake files

### Result:

* **SwiftPM:** All 164 unit tests compile and pass successfully under `swift test` with no warnings.
* **CMake:** Successfully compiles (`** BUILD SUCCEEDED **`) under Xcode generator (`cmake -B build -G Xcode && cmake --build build`).

### Issue Tracking:
Fixes #484


CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+include(CTest)
+include(GNUInstallDirs)
+include(SwiftSupport)
+
+if(NOT APPLE)
+  find_package(dispatch QUIET)
+  find_package(Foundation QUIET)
+  if(BUILD_TESTING)
+    find_package(XCTest CONFIG QUIET)
```
kukushechkin: `swift-log` is using Swift Testing, is `XCTest` package necessary?

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,112 @@
+# Returns the architecture name in a variable
+#
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# Sets ${result_var_name} with the converted architecture name derived from
+# CMAKE_SYSTEM_PROCESSOR.
+function(get_swift_host_arch result_var_name)
```
kukushechkin: Where is `get_swift_host_arch` being used?

Sources/Logging/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  LoggingSystem.swift
+  MetadataProvider.swift
+  Handlers/MultiplexLogHandler.swift
+  Handlers/StreamLogHandler.swift
+  Handlers/SwiftLogNoOpLogHandler.swift)
+
+set_target_properties(Logging PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES "$<BUILD_INTERFACE:$<TARGET_PROPERTY:Swift_MODULE_DIRECTORY>>;$<BUILD_INTERFACE:$<TARGET_FILE_DIR:Logging>>")
+
+target_compile_options(Logging PRIVATE
```
kukushechkin, line 32: Can there be a reusable list of build options? Right now they are copied from target to target.

kukushechkin (changes requested):
Thank you @mitras-git for taking this! This looks very promising. I've left a few comments. I would also ask you to add CMake-specific CI, [`swift-argument-parser`](https://github.com/apple/swift-argument-parser/blob/main/.github/workflows/pull_request.yml#L38) has an example.

CMakeLists.txt
mitras-git: I see, the XCTest package is not necessary. It was a leftover assumption from standard testing boilerplate. I've removed the XCTest find statement.

Sources/Logging/CMakeLists.txt
mitras-git: Good idea! These compile options are quite verbose, I'll create a new definition for it in the root CMakeLists file in the root directory

mitras-git: There should now be a single definition called `SWIFT_LOG_COMPILE_OPTIONS` in the root CMakeLists file. This definition is now reused across all targets

cmake/modules/SwiftSupport.cmake
mitras-git: Thanks for spotting this, `get_swift_host_arch()` is indeed unused. I added it from a template helper module to find the host architecture for path handling, but didn't end up using it and left it there. I have removed it

mitras-git:
I have addressed all the comments and committed the changes requested! I have also gone ahead and enforced a minimum Swift compiler version of 6.1 in commit 1787824, which I missed on doing. The requested update to the CI workflow has been implemented too after taking inspiration from the mentioned file in the latest commit c49e266. Thanks!

.github/workflows/pull_request.yml
```diff
@@ -87,3 +87,27 @@ jobs:
     uses: apple/swift-nio/.github/workflows/wasm_swift_sdk.yml@main
     with:
       nightly_next_enabled: false
+
+  cmake-build:
```
kukushechkin: Oh, after another look at this, it seems [`swift-configuration` has a much better way of doing this](https://github.com/apple/swift-certificates/blob/main/.github/workflows/pull_request.yml#L42), using a shared workflow. `swift-log` should adapt that one instead.

kukushechkin (changes requested):

Sources/Logging/CMakeLists.txt
kukushechkin: I think it is `SWIFT_LOG_COMMON_COMPILE_OPTIONS` actually, but yeah, there is.

.github/workflows/pull_request.yml
mitras-git: To clarify, should I reference the `swift-nio` shared workflow directly using `uses: apple/swift-nio/...` as shown in the link attached, or do you want me to copy its actual implementation details into a local workflow file in this repository?

kukushechkin: Referencing `swift-nio` workflows is the right way, similar to how it is done in `swift-configuration`.

mitras-git: Done! I also made sure that the image the job uses the Swift 6.1 container.

kukushechkin:
@mitras-git there are CI failures that need addressing, could you take a look?

- CMake build fails https://github.com/apple/swift-log/actions/runs/29422281941/job/87580457000?pr=485
- License headers cleanup https://github.com/apple/swift-log/actions/runs/29422281941/job/87580457048?pr=485

mitras-git:
Thanks, I've made the updates to the license headers. I also updated capitalization on the CMake workflow. 

The CMake CI tests failed because `BUILD_TESTING` was defaulted to ON. So Cmake tried to configure the test targets `LoggingTesting` and the in memory one. The other reason was I hardcoded the Xcode `plugin-path` to macOS only, which was an oversight on my end. 

These issues should now be fixed!

mitras-git:
Updated the license headers again, I missed out on adding the double hash comment. They should pass this time, I ran the license header check locally and passed the test.

cmake/modules/CMakeLists.txt
kukushechkin, line 1: I think this file should also have the license header.

mitras-git: Done!

Tests/InMemoryLoggingTests/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+#
+# See LICENSE.txt for license information
+# See CONTRIBUTORS.txt for the list of Swift Logging API project authors
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+#===----------------------------------------------------------------------===#
+
+if(APPLE)
+  add_library(InMemoryLoggingTests MODULE
```
kukushechkin: How are you running tests? I don't think creating xctest bundles help here, as we use Swift Testing. Neither we seem to create executables (on macOS at least), so no tests run with, e.g. `ninja run`. There is a discussion on how to run tests with CMake (https://forums.swift.org/t/using-swift-testing-with-cmake/79651/13). Also we do not run tests in the CI, only build.

mitras-git: Thanks for for catching that. I was compiling the test targets into `.xctest` bundle modules on macOS, I was working with an older macOS template, that I now realize was for XCTest frameworks. I'll get rid of these bundle constraints.

Reagarding your question about running tests, currently, I too only used CMake purely to verify compilation rather than running tests with ninja. I still use the native `swift test`.

I'll remove the macOS bundle logic from the test files and replace them to build as standard library targets using `add_library`.

kukushechkin: `add_library` won't help either, there is nothing to run tests with. [Swift Testing guide](https://github.com/swiftlang/swift-testing/blob/main/Documentation/CMake.md) recommends creating special entry point, but this makes the whole construct too complicated. I think at this stage it would make more sense to not to add tests in the CMake build, there is no CI to run them atm. Could you please remove CMake from tests?

mitras-git: Got it! I've removed the test targets

Sources/InMemoryLogging/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+add_library(InMemoryLogging
+  "InMemoryLogHandler.swift"
+)
+
+set_target_properties(InMemoryLogging PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES "$<BUILD_INTERFACE:$<TARGET_PROPERTY:Swift_MODULE_DIRECTORY>>;$<BUILD_INTERFACE:$<TARGET_FILE_DIR:InMemoryLogging>>")
+
+target_compile_options(InMemoryLogging PRIVATE
+  ${SWIFT_LOG_COMMON_COMPILE_OPTIONS}
+  "$<$<BOOL:${BUILD_TESTING}>:SHELL:-enable-testing>")
```
kukushechkin: `BUILD_TESTING` and everything related to it can be removed as well now

Sources/Logging/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  "LoggingSystem.swift"
+  "MetadataProvider.swift"
+)
+
+set_target_properties(Logging PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES "$<BUILD_INTERFACE:$<TARGET_PROPERTY:Swift_MODULE_DIRECTORY>>;$<BUILD_INTERFACE:$<TARGET_FILE_DIR:Logging>>")
+
+target_compile_options(Logging PRIVATE
+  ${SWIFT_LOG_COMMON_COMPILE_OPTIONS}
+  "$<$<BOOL:${BUILD_TESTING}>:SHELL:-enable-testing>")
```
kukushechkin: `BUILD_TESTING` and everything related to it can be removed as well now

CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+
+set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+
+if(MSVC)
```
kukushechkin, line 56: @mitras-git did you have a chance to verify the MSCV branch?

Sources/InMemoryLogging/CMakeLists.txt
mitras-git: Done!

Sources/Logging/CMakeLists.txt
mitras-git: Done!

CMakeLists.txt
mitras-git: Unfortunately, I did not. I added this block from a standard config template for Swift compilation with CMake on Windows.

mitras-git: I added this block because in my experience, sometimes compiler mismatches occur if the dependencies use different runtime libraries than the client's own binary.

kukushechkin:
@mitras-git thank you for keeping working on this, I really appreciate it. We're getting closer. Could you update the PR description to match all the recent changes, like removal of tests?

CMakeLists.txt
kukushechkin: Ok, got it. I still would like to verify this before approving.

mitras-git: Let me get a Windows VM set up on my Mac. I will try and compile it and come back to you with the results

mitras-git:
Oh absolutely, I'm glad I could be of help. I'll update the description, thanks for pointing it out, should've done it before

mitras-git:
PR description has been updated!

CMakeLists.txt
mitras-git: Update: I successfully compiled and ran the code on a Windows 11 VM running on my Mac through VMWare Fusion. I'm actually glad you asked me to check because I hit a edge case. 

Since my Windows VM was running on `ARM64` architecture because my Mac's chip is the M5 Pro; unlike the usual x64 architecture that most Windows users run on, I was hitting loads of issues trying to get the project to build. After some debugging I figured, without explicit instructions, standard `swiftc` on Windows ARM will default to targeting `aarch64` even if you build it using the `x64 Native Tools Command Prompt`.

When building on Windows on `ARM64` (such as inside a Windows 11 VM on Apple Silicon or native ARM chips like the new Snapdragon or NVIDIA RTX Spark devices), this causes severeal target architecture mismatches. For example, if someone runs inside Visual Studio's default `x64-Debug profile` (which is the default for Visual Studio) or the `x64 Native Tools Command Prompt`, `swiftc` still outputs ARM64 objects while MSVC links against x64 CRT libraries (`msvcrt.lib`), causing linker failures (`LNK4272` / `LNK2001`). 

Thanks for asking me to check. I have made the necessary fixes in commit d64433b to support any developer workflow.

mitras-git:
Here's what I changed in commit d64433b to support different Windows architectures

1. **Two-Stage Language Initialization:** Enables the C language first so CMake can inspect the active MSVC compiler environment using `CMAKE_C_COMPILER_ARCHITECTURE_ID`.
2. **Automatic Target Alignment:** Maps `CMAKE_Swift_COMPILER_TARGET` to either `aarch64-unknown-windows-msvc` or `x86_64-unknown-windows-msvc` to match whatever architecture the MSVC environment is currently configured to build for, before enabling Swift.
3. **C Runtime Consistency:** Configures `CMAKE_MSVC_RUNTIME_LIBRARY` and `CMAKE_Swift_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY` so Swift and MSVC share the same C Runtime DLL (/MD or /MDd), resolving LNK4098 warnings.

### Testing:
* Verified across multiple environments on Windows 11 Pro Build 25H2 ARM64 (Apple Silicon VM via VMware Fusion):
* Visual Studio Pro IDE (x64-Debug and arm64-Debug presets)
* x64 Native Tools Command Prompt
* ARM64 Native Tools Command Prompt

kukushechkin (approved):

FranzBusch (approved):

PR #487, opened 2026-07-22
Handle Windows path separators in Logger.currentModule(filePath:)

ZayanKhan-12:
### Motivation

`Logger.currentModule(filePath:)` derives a log's `source` from `#file` by taking the second-to-last path component — but it only recognizes `/` as a path separator. On Windows, `#file` produces backslash-separated paths (`C:\...\Sources\MyModule\File.swift`), so the lookup finds no separator and falls through to `"n/a"`. Every log emitted through the deprecated file-based `LogHandler` compatibility path (`LogHandler.swift`'s `log(level:...file:...)` shim, which calls `currentModule(filePath: file)`) therefore reports `source: n/a` on Windows instead of the module name, breaking source-based filtering.

The `#fileID`-based overload is unaffected (`#fileID` always uses `/`).

### Modifications

- `currentModule(filePath:)` now treats both `/` and `\` as path separators when extracting the second-to-last component.
- Added `currentModuleFromFilePath` tests covering POSIX-style paths, Windows-style paths, and paths with too few components. The Windows cases fail against the previous implementation and pass with this change.

### Result

Logs emitted on Windows via the deprecated file-based API report the correct module as their `source`. No behavior change on Darwin/Linux (all 165 existing tests pass unchanged).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/Logging/Logger.swift
+++ b/Sources/Logging/Logger.swift
@@ -1374,9 +1374,12 @@ extension Logger {
 extension Logger {
     @inlinable
     internal static func currentModule(filePath: String = #file) -> String {
+        // `#file` paths use the platform's path separator, so on Windows the
+        // components are separated by backslashes rather than slashes.
+        let isPathSeparator: (UInt8) -> Bool = { $0 == UInt8(ascii: "/") || $0 == UInt8(ascii: "\\") }
         let utf8All = filePath.utf8
-        return filePath.utf8.lastIndex(of: UInt8(ascii: "/")).flatMap { lastSlash -> Substring? in
-            utf8All[..<lastSlash].lastIndex(of: UInt8(ascii: "/")).map { secondLastSlash -> Substring in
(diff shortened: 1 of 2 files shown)


samuelmurray:
I wonder if it's actually correct to call the `currentModule(filePath: String = #file)` function in the default implementations of the deprecated `LogHandler` functions. Since all log methods in `Logger` defaults to `#fileID`, the only way we could get a full path is if the user explicitly passes it (or wraps `Logger` with their own API that uses `#file`). I guess it might be considered breaking for such use-cases (but it seems like an unlikely combination, and uses deprecated APIs), but wouldn't it be better to remove the `currentModule(filePath: String = #file)` altogether and call `currentModule(fileID: String = #fileID)` from the default implementations as well?

EDIT: Something like this: https://github.com/apple/swift-log/pull/495

kukushechkin:
@ZayanKhan-12 thank you for addressing this, but I would prefer we have a discussion before making any changes. Could you please open an Issue instead?

PR #489, opened 2026-07-22
constrain the scheduled builds so that they only invoke on the apple/swift-log repo

heckj:
Constrain the scheduled builds so that they only invoke on the apple/swift-log repo

### Motivation:

I noticed that my fork was running nightly scheduled jobs - which didn't add any value, and chewed up my GH Actions resources. 

### Modifications:

adds a check-repository semaphore and has all the nightly scheduled jobs depend on checking it so that forks won't automatically build and run this jobs.

### Result:

Nightly jobs for forks are only on request, not scheduled

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,8 +10,16 @@ on:
     - cron: "0 8,20 * * *"
 
 jobs:
+  check-repository:
+    name: Check repository
+    runs-on: ubuntu-latest
+    if: github.event_name != 'schedule' || github.repository == 'apple/swift-log'
+    steps:
+      - run: exit 0
+
   unit-tests:
     name: Unit tests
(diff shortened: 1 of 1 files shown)


czechboy0:
You can also disable actions on your fork in your fork's repo settings, under Actions, which won't require updating the upstream workflows.

heckj:
yeah, I did that first - but thought this would be sufficiently valuable for anyone else forking the repo down the road that it was worthwhile to submit for consideration.

FranzBusch:
I'm not sure yet. We would have to do this across all our repos for it to become truly effective.

FranzBusch:
I'm going to close this since this is something users can control on their own. A new fork will normally start with the right settings so actions are not running.

PR #490, opened 2026-07-23
Document how to construct log handlers that need more than a label

### LINKED ISSUE #80: Limited usage of this as a general purpose log handler
I was looking into creating a `FileLogHandler` but I am unsure of a safe and convenient way to go about doing that due to issues inherent with logging to files vs something like stdout with the current limitations of this framework.

For example, there is no available throwable factory overload and nowhere is it documented that a logger _requires_ an initializer that takes a single String argument. Looking through the code it becomes much easier to see that you need some kind of an initializer that takes a single string argument. Why is there no official protocol requirement for the initializer if this is clearly a requirement?

I want to make a FileLogHandler, but with the current standard functionality, everyone would just be accustomed to passing in a label. I could just log in a common location like /var/log on linux, but people should be able to specify a custom file location if they so choose.

The Logger initializer that takes a LogHandler argument is internal so I cannot use that and the restrictions on the factory signature are too restrictive to allow this. I could make the String parameter the path to the log file, but since it is not throwable or nullable I cannot do any sort of validation that the path exists or is writable by the current process. Do I just silently fail? Should I just `print` a message?

Why are there no throwable/nullable overloads for the factory? or why is there no ability to use a custom factory signature? The fileprivate restrictions make it more difficult to provide my own handy extensions to achieve my desired functionality.

Is there something I'm missing here or was there just massive oversight when designing this? From what I can tell it would be extremely difficult to set up a logger that sends logs to an external service requiring connection parameters and there's no way you could validate those parameters during initialization of the handler.

kukushechkin:
Document best practice for creating complex `LogHandlers`. 

### Motivation:

`LogHandler`s often need an expensive resource to write or export log events. This pattern should be documented.

### Modifications:

Add "Constructing handlers that need more than a label" section to `ImplementingALogHandler.md` guide.

### Result:

Closes #80.

### DIFF
--- a/Sources/Logging/Docs.docc/ImplementingALogHandler.md
+++ b/Sources/Logging/Docs.docc/ImplementingALogHandler.md
@@ -104,21 +104,20 @@ public struct PrintLogHandler: LogHandler {
 
 ### Constructing handlers that need more than a label
 
-Many backends need more than a `label` to build: a file path, a remote address, credentials, or
+Many log handler backends need more than a `label` in order to construct them: a file path, a remote address, credentials, or
 a buffer. A handler is an ordinary value that you construct, so nothing constrains its
 initializer — it can take any parameters it needs, and it can `throw`. There is no
 protocol-required `init(label:)`. The single `label` argument on factory methods such as
 `StreamLogHandler.standardOutput(label:)` is a convention, not a requirement.
 
-Open expensive or fallible resources — a file, a socket — **up front**, in your own `throw`ing
-initializer, so failures surface at setup time. ``LogHandler/log(event:)`` cannot throw, so an
-error deferred to logging time has nowhere to go. Keep the handler `struct` cheap and copyable,
-and place the shared resource behind a thread-safe reference type: value semantics apply to a
-handler's *configuration* — its level and metadata — not its *destination*, so copies writing to
+Open expensive or fallible resources, such as a file, a socket, early and up front in your own initializer that can `throw`, so that failures surface at setup time.
+``LogHandler/log(event:)`` cannot throw, so an error deferred to logging time has nowhere to go.
+Keep the handler `struct` cheap and copyable by placing any shared resource behind a thread-safe reference type.
+Value semantics apply to a handler's *configuration* — its level and metadata — not its *destination*, so copies writing to
 the same file are correct and expected.
 
-The `FileLogHandler` below opens its file once, in a `throw`ing initializer on a shared
-`Destination`. The handler `struct` itself stays a cheap, copyable value:
+The `FileLogHandler` below opens its file once, in an initializer that throws and exception when you create it with an overlapping `Destination`.
+The handler `struct` itself stays a cheap, copyable value:
 
 ```swift
 import Foundation
@@ -127,8 +126,9 @@ import Synchronization
 
 /// A log handler that appends formatted log lines to a file on disk.
 public struct FileLogHandler: LogHandler {
-    /// The shared, thread-safe destination. Open it once, then share it across handlers: being a
-    /// reference type, all those value-semantic handlers write through one locked file handle.
+    /// The shared, thread-safe destination. Open it once, then share it across handlers: because 
+    /// the destination is a reference type, all those value-semantic handlers write through a single
+    /// locked file handle.
     public final class Destination: @unchecked Sendable {
         private let fileHandle: Mutex<FileHandle>
 
@@ -202,10 +202,10 @@ public struct FileLogHandler: LogHandler {
 }
 ```
 
-#### Installing it
+#### Using your log handler
 
-Construct the destination where you can `try`, build a logger backed by the handler, and bind it
-as ``Logger/current`` for the application's lifetime with `withLogger`:
+Construct the destination where you can use `try` to build a logger backed by your handler, then bind it
+as ``Logger/current`` for the application's lifetime using `withLogger`:
 
 ```swift
 // Open the file once, where setup can throw.
--- a/Sources/Logging/Docs.docc/ImplementingALogHandler.md
+++ b/Sources/Logging/Docs.docc/ImplementingALogHandler.md
@@ -115,7 +115,7 @@ initializer, so failures surface at setup time. ``LogHandler/log(event:)`` canno
 error deferred to logging time has nowhere to go. Keep the handler `struct` cheap and copyable,
 and place the shared resource behind a thread-safe reference type: value semantics apply to a
 handler's *configuration* — its level and metadata — not its *destination*, so copies writing to
-the same file is correct and expected.
+the same file are correct and expected.
 
 The `FileLogHandler` below opens its file once, in a `throw`ing initializer on a shared
 `Destination`. The handler `struct` itself stays a cheap, copyable value:
--- a/Sources/Logging/Docs.docc/ImplementingALogHandler.md
+++ b/Sources/Logging/Docs.docc/ImplementingALogHandler.md
@@ -123,14 +123,14 @@ The `FileLogHandler` below opens its file once, in a `throw`ing initializer on a
 ```swift
 import Foundation
 import Logging
+import Synchronization
 
 /// A log handler that appends formatted log lines to a file on disk.
 public struct FileLogHandler: LogHandler {
     /// The shared, thread-safe destination. Open it once, then share it across handlers: being a
     /// reference type, all those value-semantic handlers write through one locked file handle.
     public final class Destination: @unchecked Sendable {
-        private let lock = NSLock()
-        private let fileHandle: FileHandle
+        private let fileHandle: Mutex<FileHandle>
 
         /// Opens `url` for appending. All validation happens here, at setup time.
         public init(writingTo url: URL) throws {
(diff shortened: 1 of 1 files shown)


Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+initializer — it can take any parameters it needs, and it can `throw`. There is no
+protocol-required `init(label:)`. The single `label` argument on factory methods such as
+`StreamLogHandler.standardOutput(label:)` is a convention, not a requirement.
+
+Open expensive or fallible resources — a file, a socket — **up front**, in your own `throw`ing
+initializer, so failures surface at setup time. ``LogHandler/log(event:)`` cannot throw, so an
+error deferred to logging time has nowhere to go. Keep the handler `struct` cheap and copyable,
+and place the shared resource behind a thread-safe reference type: value semantics apply to a
+handler's *configuration* — its level and metadata — not its *destination*, so copies writing to
+the same file is correct and expected.
```
heckj:
```suggestion
the same file are correct and expected.
```

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+`Destination`. The handler `struct` itself stays a cheap, copyable value:
+
+```swift
+import Foundation
+import Logging
+
+/// A log handler that appends formatted log lines to a file on disk.
+public struct FileLogHandler: LogHandler {
+    /// The shared, thread-safe destination. Open it once, then share it across handlers: being a
+    /// reference type, all those value-semantic handlers write through one locked file handle.
```
heckj:
```suggestion
    /// The shared, thread-safe destination. Open it once, then share it across handlers: because 
    /// the destination is a reference type, all those value-semantic handlers write through a single
    /// locked file handle.
```

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
 
 ```
 
+### Constructing handlers that need more than a label
+
+Many backends need more than a `label` to build: a file path, a remote address, credentials, or
```
heckj:
```suggestion
Many log handler backends need more than a `label` in order to construct them: a file path, a remote address, credentials, or
```

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+a buffer. A handler is an ordinary value that you construct, so nothing constrains its
+initializer — it can take any parameters it needs, and it can `throw`. There is no
+protocol-required `init(label:)`. The single `label` argument on factory methods such as
+`StreamLogHandler.standardOutput(label:)` is a convention, not a requirement.
+
+Open expensive or fallible resources — a file, a socket — **up front**, in your own `throw`ing
+initializer, so failures surface at setup time. ``LogHandler/log(event:)`` cannot throw, so an
+error deferred to logging time has nowhere to go. Keep the handler `struct` cheap and copyable,
+and place the shared resource behind a thread-safe reference type: value semantics apply to a
+handler's *configuration* — its level and metadata — not its *destination*, so copies writing to
```
heckj:
```suggestion
Open expensive or fallible resources, such as a file, a socket, early and up front in your own initializer that can `throw`, so that failures surface at setup time.
``LogHandler/log(event:)`` cannot throw, so an error deferred to logging time has nowhere to go.
Keep the handler `struct` cheap and copyable by placing any shared resource behind a thread-safe reference type.
Value semantics apply to a handler's *configuration* — its level and metadata — not its *destination*, so copies writing to
```

I broke up the sentences a bit as a suggestion, and down graded the bolding "up-front" since that's a stand out against the Apple Style Guide. I _think_ the rest of this narrative description nails down the do it early, not later - but we can add the bolding back in if you think it pushes the detail where it needs to be. If it's super relevant, we may even want a full call-out aside, such as `> Important` to highlight that detail for the sake of later performance impacts.

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+the same file is correct and expected.
+
+The `FileLogHandler` below opens its file once, in a `throw`ing initializer on a shared
+`Destination`. The handler `struct` itself stays a cheap, copyable value:
```
heckj:
```suggestion
The `FileLogHandler` below opens its file once, in an initializer that throws and exception when you create it with an overlapping `Destination`.
The handler `struct` itself stays a cheap, copyable value:
```

re-worked the sentence so not use a symbol (throws) mixed with narrative setups - ing, plural, etc.

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+```swift
+import Foundation
+import Logging
+
+/// A log handler that appends formatted log lines to a file on disk.
+public struct FileLogHandler: LogHandler {
+    /// The shared, thread-safe destination. Open it once, then share it across handlers: being a
+    /// reference type, all those value-semantic handlers write through one locked file handle.
+    public final class Destination: @unchecked Sendable {
+        private let lock = NSLock()
```
heckj: I'll defer to you entirely - but is NSLock() the locking mechanism that we're encouraging folks to replicate? Other docs I've seen lean into a more direct Mutex around a class. Conceptually I'm 100% good with this, just want to make sure that it matches what we think are current best practices here - and I've never been 100% clear on when to use one form of locking over another.

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+    }
+
+    public subscript(metadataKey key: String) -> Logger.Metadata.Value? {
+        get { self.metadata[key] }
+        set { self.metadata[key] = newValue }
+    }
+}
+```
+
+#### Installing it
```
heckj:
```suggestion
#### Using your log handler
```

What do you think of this as an alternate section name?

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+
+Construct the destination where you can `try`, build a logger backed by the handler, and bind it
+as ``Logger/current`` for the application's lifetime with `withLogger`:
```
heckj:
```suggestion
Construct the destination where you can use `try` to build a logger backed by your handler, then bind it
as ``Logger/current`` for the application's lifetime using `withLogger`:
```

Sources/Logging/Docs.docc/ImplementingALogHandler.md
```diff
@@ -102,6 +102,140 @@ public struct PrintLogHandler: LogHandler {
...
+
+The `try` on `Destination` is where setup fails if the path is unwritable: validation happens
+before the logger is constructed, so you never need a throwing factory overload. The same pattern
+extends to handlers that reach a remote service — validate and connect up front, keep
+``LogHandler/log(event:)`` non-blocking by buffering onto a background task, and provide a
+`close()` or `shutdown()` (or a
+[Swift Service Lifecycle](https://github.com/swift-server/swift-service-lifecycle) service) to
+flush on shutdown. For runtime failures the handler cannot avoid — a full disk, a broken pipe —
+degrade gracefully inside `log(event:)` by dropping, retrying, or reporting to `stderr`. Never
+call `fatalError` on the logging path.
```
heckj:
```suggestion
The `try` on `Destination` is where setup fails if the path is unwritable: validation happens
before the logger is constructed, so you never need a throwing factory overload. 
Use the same pattern for handlers that connect to a remote service:
  - validate and connect up front
  - keep ``LogHandler/log(event:)`` non-blocking by buffering onto a background task
  - provide a `close()` or `shutdown()` (or use [Swift Service Lifecycle](https://github.com/swift-server/swift-service-lifecycle) service) to flush on shutdown

For runtime failures the handler can't avoid, such as a full disk or a broken pipe, degrade gracefully inside `log(event:)` by dropping the log message, retrying, or reporting to `stderr`.
Never call `fatalError` on the logging path.
```

The middle bit about the pattern for a remote logger was super dense, so I thought to try and break it out as a short bulleted list. What do you think?

heckj (commented):
I reworked some of the language to try and make it a bit more accessible and lean into narrative English vs. mixing symbols and English phrasing together. One nit for a verb tense agreement, but mostly it's just trying to open this up a bit of make it easier to scan and read.

Double check on the importance note of the detail of opening precious resources early and keeping them protected - see if what I suggested aligns, if we should add a little more emphasis back in.

Sources/Logging/Docs.docc/ImplementingALogHandler.md
kukushechkin: I think it is fine without bolding.

Sources/Logging/Docs.docc/ImplementingALogHandler.md
kukushechkin: I agree, let's promote `Mutex`, updated.

Sources/Logging/Docs.docc/ImplementingALogHandler.md
kukushechkin: Yes, this is more clear.

heckj (approved):

PR #504, opened 2026-09-02
Document that setting attributes on a collection value is a programmer error

lenamonj:
### Motivation:

`Logger.MetadataValue.attributes` documents the setter as a no-op for `.dictionary` and `.array`, but it calls `assertionFailure` there. Debug builds trap, release builds do the documented no-op.

### Modifications:

Doc comment only. The assertion stays; the comment now says setting attributes on a collection value is a programmer error, that debug builds assert and release builds leave the value unchanged, and to set attributes on the leaf values instead.

### Result:

The documentation matches the code. No behaviour change.

Found by an automated audit loop (Claude); patch and description reviewed by me.


kukushechkin:
Thanks for this. I think setting attributes for an array or a dictionary should be a visible programmer error. We cannot express this as an API limitation unfortunately.

What is the expected behavior for that intermediate log handler, would it rather propagate attributes inside containers?

lenamonj:
Fair enough, a programmer error works for me. My real complaint is that debug traps while release silently drops it, and the doc says “no-op”. If you want the assertion, I’ll rework this to keep it and fix the doc to say so.
For the intermediate handler, I’d only set attributes on leaf values, recursing into containers, rather than on the container itself. Happy to go either way.

lenamonj:
Reworked as doc-only: the assertion stays, the comment now calls it a programmer error and points at setting attributes on the leaf values instead.

kukushechkin (approved):

PR #516, opened 2026-09-21
Fix nested metadata values handling

### LINKED ISSUE #515: Nested `Logger.MetadataValue.description` re-escapes on every level
`Logger.MetadataValue.description`, used for example in `StreamLogHandler`, renders a composite child node by first turning it into a plain `String`, then handing that string back to `Array`/`Dictionary`'s own `.description`, which quotes and escapes it as if it were raw text. Each level of nesting therefore re-escapes the entire string produced by the level below it, so a 5-level-deep single-element array chain renders as `["[\"[\\\"[\\\\\\\"[\\\\\\\\\\\\\\\"leaf\\\\\\\\\\\\\\\"]\\\\\\\"]\\\"]\"]"]` instead of the more expected `[[[[["leaf"]]]]]`. Because of this, serialization time and output size grow exponentially with depth rather than linearly: depth 30 (not very common, but still) already takes significant time to serialize.

kukushechkin:
Only quote leaf metadata values.

### Motivation:

There are several issues with how nested `MetadataValue`s are handled:
1. For highly nested (10k+) metadata, `.description` and `==` operator hit stack overflow. This is probably not very realistic depth, but still.
2. For not so highly nested metadata (starting at about 20) `.description` and `==` operator get very computationally intensive because of re-escaping of the already escaped children. This makes the above stack overflow crash not reachable in practice.
3. This also results in simple structured metadata to be rendered something like `["[\"[\\\"[\\\\\\\"[\\\\\\\\\\\\\\\"leaf\\\\\\\\\\\\\\\"]\\\\\\\"]\\\"]\"]"]`.
4. Finally, even after solving the re-quoting and stack over flow on rendering, there is still an issue of [recursion in `.deinit`](https://github.com/swiftlang/swift/issues/58080), resulting in a _different_ stack overflow crash.

This PR provides a fix for 2 and 3, leaving 1 and 4 out because there is no practical way of fixing it in this package, while the case is borderline unrealistic.

### Modifications:

- Only quote left `String` and `StringRepresentable` values in `Logger.MetadataValue.description`.
- Add benchmarks for metadata values description rendering.

### Result:

100 nested metadata values render cleanly. 

Resolves #515 

### Alternatives

This is discussable if this fix is really necessary. Iterative approach, because of allocations, is slower than recursion.
- Is it worth fixing the recursion, or a metadata depth check is enough?
- Is it worth fixing the nested metadata formatting?


czechboy0:
Do you need to turn the recursion into a loop to fix #515, or is this PR just fixing multiple issues in one?

Because I'm not sure the recursion one is actually a problem for which we should complicate the code. What was the original motivation?

FranzBusch:
I'm overall not sure if this is something that needs fixing in the first place. Nobody should be rendering the metadata value description as-is. @weissi @ktoso what do you think about this?

kukushechkin:
@czechboy0 no, recursion unrolling is not necessary for fixing the escaping, it solves that another issue with the unbound recursion. However, as `.deinit` recursion we cannot fix, there is indeed no value in this. I've reverted back to recursion with just the escaping fix.

czechboy0:
The equality check still seems to be done through iteration rather than recursion.

Sources/Logging/Logger.swift
```diff
@@ -1325,18 +1325,30 @@ extension Logger.MetadataValue: Equatable {
     ///   - rhs: The second metadata value.
     /// - Returns: Returns `true` if the metadata values are equivalent; otherwise `false`.
     public static func == (lhs: Logger.Metadata.Value, rhs: Logger.Metadata.Value) -> Bool {
-        switch (lhs, rhs) {
```
czechboy0, line 1328: Should this be reverted as well?

kukushechkin: Yes, right, updated now.

czechboy0 (approved):

kukushechkin:
I've fixed the benchmark thresholds file names and updated the numbers.

